PR #23, opened 2021-07-13
[T4PUB-360] Bugfix/chmod w follow symlink false

obi-t4:
# what is this PR
_copy_complete のsrcがsymlinkの場合&&dstのsymlink-toが存在しない場合, _copy_completeが失敗する問題の修正

# issue
https://tier4.atlassian.net/browse/T4PUB-360

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -29,6 +29,7 @@ class OtaError(Exception):
     """
     OTA error
     """
+
     pass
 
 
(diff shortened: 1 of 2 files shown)


Bodong-Yang:
If I understand right, the **_copy_complete** function should copy the symlink itself as it without copying the symlink target?
So **follow_symlinks** should always be false.

I have check the reference of **_copy_complete**, it seems that no callers are using this function with parameter **follow_symlinks=True** though, so I think the change to this function is OK.

obi-t4:
@Bodong-Yang 
> If I understand right, the **_copy_complete** function should copy the symlink itself as it without copying the symlink target right?
> So **follow_symlinks** should always be false.

right.

No caller are using option `follow_symlinks=True`, so I removed this optional argument.

Bodong-Yang (approved):

PR #25, opened 2021-07-13
[T4PUB-349] Paralleling regular files processing

Bodong-Yang:
(Function of logging to file has been temporarily removed due to paralleling)

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -1010,16 +1010,16 @@ class OtaClient:
         main entry for paralleling processing regular files
         """
         try:
-            prev_inf = ""
+            prev_inf = None
             # hardlinked file
             if int(rfile_inf.links) >= 2:
                 prev_inf = global_var_dict["tmp-dict-hardlink_reg"].setdefault(
                     rfile_inf.sha256hash, rfile_inf
                 )
                 # if the upcoming rfile entry is the first copy of hardlinked file
-                # then the prev_inf should be ""
+                # then the prev_inf should be None
                 if prev_inf == rfile_inf:
-                    prev_inf = ""
+                    prev_inf = None
 
             if rfile_inf.path.find("/boot/") == 0:
                 # /boot directory file
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -566,9 +566,6 @@ class OtaClient:
             logger.error(f"hash missmatch: {dest_file}")
             logger.error(f"  dl hash: {digest}")
             logger.error(f"  hash: {target_hash}")
-            fl = global_var_dict["tmp-queue-log_queue"]
-            if fl != "":
-                fl.put("hash missmatch: " + dest_file + "\n")
             return False
         return True
 
@@ -939,42 +936,17 @@ class OtaClient:
                 elif vname[1] == "list":
                     setattr(self, vname[-1], list(v))
 
-    def _process_regular_files_log2file_listener(self, flog_path, log_queue, event):
-        """
-        logger worker for _process_regular_files
-        """
-        with open(flog_path, "w") as f:
-            while not event.is_set() or not log_queue.empty():
-                msg = log_queue.get()
-                f.write(msg + "\n")
-
     def _process_regular_files(self, rootfs_dir, rfiles_list, target_dir):
-        with Manager() as manager, tempfile.NamedTemporaryFile(delete=False) as flog:
-            log_file_path, log_q, ev = flog.name, manager.Queue(), manager.Event()
+        with Manager() as manager:
             ecb_queue = manager.Queue()
             # variables passed to child processes
             # variable naming pattern: <prefix>-<type>-<var_name>
             # prefix tmp: for temporary use
             # prefix staging: used to update corresponding class attribute
             gvar_dict = {
-                "tmp-queue-log_queue": log_q,
                 "tmp-dict-hardlink_reg": manager.dict(),
                 "staging-dict-_rollback_dict": manager.dict(),
             }
-
-            # launch log2file logger
-            #   passing file descriptor to sub-process is not possible,
-            #   so we pass the log file path to the logger, and let the logger
-            #   open the log file by itself
-            p_logger = Process(
-                target=self._process_regular_files_log2file_listener,
-                args=(
-                    log_file_path,
-                    log_q,
-                    ev,
-                ),
-            )
-            p_logger.start()
 
             # default to one worker per CPU core
             with Pool(
@@ -1012,14 +984,6 @@ class OtaClient:
                 # wait for all tasks to complete
                 # not apply timeout currently
                 pool.join()
-
-                # notify the logger to terminate
-                ev.set()
-                # avoid logger hangging
-                log_q.put("All tasks finished!")
-
-            # wait for logger
-            p_logger.join()
 
             # check if any exception is triggered
             if not ecb_queue.empty():
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -967,17 +967,25 @@ class OtaClient:
                         not in gvar_dict["tmp-dict-hardlink_reg"]
                     ):
                         # block the flow until the first copy of hardlinked file is ready
-                        pool.apply(
-                            self._process_regular_file,
-                            (rootfs_dir, target_dir, rfile_inf),
-                            error_callback=ecb,
-                        )
+                        try:
+                            pool.apply(
+                                self._process_regular_file,
+                                (rootfs_dir, target_dir, rfile_inf),
+                            )
+                        except Exception as e:
+                            ecb(e)
+                            break
                     else:
-                        pool.apply_async(
-                            self._process_regular_file,
-                            (rootfs_dir, target_dir, rfile_inf),
-                            error_callback=ecb,
-                        )
+                        try:
+                            pool.apply_async(
+                                self._process_regular_file,
+                                (rootfs_dir, target_dir, rfile_inf),
+                                error_callback=ecb,
+                            )
+                        except:
+                            # the only exception will be catched is ValueError
+                            # caused by calling pool.apply_async when the pool terminated
+                            pass
 
                 # stop accepting new tasks
                 pool.close()
@@ -989,9 +997,8 @@ class OtaClient:
             if not ecb_queue.empty():
                 # if any exception being raised in any child processes,
                 # raise it again in the main process.
-                e = ecb_queue.get()
                 logger.error(
-                    f"process regular files failed: {e}. All sub processess terminated."
+                    f"process regular files failed: {ecb_queue.get()}. All sub processess terminated."
                 )
                 raise OtaError(f"process regular files failed!")
             else:  # everything is ALLRIGHT!
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -20,7 +20,7 @@ from hashlib import sha256
 from ota_status import OtaStatus
 from grub_control import GrubCtl
 from ota_metadata import OtaMetaData
-from multiprocessing import Pool, Manager
+from multiprocessing import Process, Pool, Manager
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
@@ -566,8 +566,9 @@ class OtaClient:
             logger.error(f"hash missmatch: {dest_file}")
             logger.error(f"  dl hash: {digest}")
             logger.error(f"  hash: {target_hash}")
-            # if fl != "":
-            #     fl.write("hash missmatch: " + dest_file + "\n")
+            fl = global_var_dict["tmp-queue-log_queue"]
+            if fl != "":
+                fl.put("hash missmatch: " + dest_file + "\n")
             return False
         return True
 
@@ -804,6 +805,7 @@ class OtaClient:
             logger.debug(f"links: {regular_inf.links}")
             os.link(prev_inf.path, regular_inf.path)
         else:
+            staging_rollback_dict = global_var_dict["staging-dict-_rollback_dict"]
             # no hard link
             if (
                 os.path.isfile(regular_inf.path)
@@ -814,7 +816,8 @@ class OtaClient:
                     self._rollback_dir, os.path.basename(regular_inf.path)
                 )
                 _copy_complete(regular_inf.path, rollback_file)
-                self._rollback_dict[regular_inf.path] = regular_inf.path
+
+                staging_rollback_dict[regular_inf.path] = regular_inf.path
                 logger.debug("file already exist! no copy or download!")
             else:
                 if os.path.isfile(regular_inf.path):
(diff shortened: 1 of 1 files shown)


Bodong-Yang:
Issue that has not been covered:
The **self._rollback_dict** is being writen in function **self._gen_boot_dir_file**, which may cause simultaneously read/write on it.

Bodong-Yang:
WIth commit 295dc86, the following issues are addressed:
1. logging to files
    A logger listener process will be launched before hand to collect logs from sub-processes.
2. update class attributes
    Changes and data generated by sub processes (For example, updates to **rollback_dict** class attribute) will be staged, after all the sub processes complete, the corresponding class attributes will be updated after hand.

app/ota_client.py
```diff
@@ -565,15 +566,16 @@ def _download_raw_file(self, url, dest_file, target_hash="", fl=""):
             logger.error(f"hash missmatch: {dest_file}")
             logger.error(f"  dl hash: {digest}")
             logger.error(f"  hash: {target_hash}")
+            fl = global_var_dict["tmp-queue-log_queue"]
             if fl != "":
-                fl.write("hash missmatch: " + dest_file + "\n")
+                fl.put("hash missmatch: " + dest_file + "\n")
```
obi-t4, line 571: This log file is not necessary. So remove this lines, and also remove log file server process.

app/ota_client.py
```diff
@@ -896,48 +892,138 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
+                    setattr(self, vname[-1], list(v))
+
+    def _process_regular_files_log2file_listener(self, flog_path, log_queue, event):
+        '''
+            logger worker for _process_regular_files
+        '''
+        with open(flog_path, "w") as f:
+            while not event.is_set() or not log_queue.empty():
+                    msg = log_queue.get()
+                    f.write(msg+"\n")
```
obi-t4, line 951: Can be removed this function and p_logger process.

app/ota_client.py
```diff
@@ -896,48 +892,138 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
+                )
+            p_logger.start()
+
+            # default to one worker per CPU core
+            with Pool(
+                initializer=self._process_regular_files_pool_init,
+                initargs=(gvar_dict,)
+            ) as pool:
+                for rfile_inf in rfiles_list:
+                    pool.apply_async(
```
obi-t4: return value of `pool.apply_async` should be checked by get?
like:
```suggestion
                    results = [pool.apply_async(...) for rfile_inf in rfiles_list]
                    for result in results:
                        result.get()
```

app/ota_client.py
```diff
@@ -896,48 +892,147 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
+
+            # default to one worker per CPU core
+            with Pool(
+                initializer=self._process_regular_files_pool_init,
+                initargs=(gvar_dict,)
+            ) as pool:
+                for rfile_inf in rfiles_list:
+                    if int(rfile_inf.links) >=2 and rfile_inf.sha256hash not in gvar_dict["tmp-dict-hardlink_reg"]:
+                        # block the flow until the first copy of hardlinked file is ready
+                        pool.apply(
```
obi-t4, line 985: (again since outdated)
return value of pool.apply_async should be checked by get?
like:
return value of `pool.apply_async` should be checked by get?
like:
```suggestion
                    results = [pool.apply_async(...) for rfile_inf in rfiles_list]
                    for result in results:
                        result.get()
```

app/ota_client.py
Bodong-Yang: In commit a20f92c, I update the logic of how sub processes notifying the main process about the exception,
The worker process needs not to return anything.

If everything is OK, the worker process(**_process_regular_file**) will terminate silently without triggering anything.
If an exception is triggered in a sub process, an error_callback function will be triggered to catch that exception, and then pass it to the main process(**_process_regular_files**). The main process will raise a new exception to notify the upper caller that the regular files processing failed and terminated.

To ensure the whole OTA operation is atomic, all sub processes will also be terminated by any exception being triggered.

app/ota_client.py
Bodong-Yang: Resolve in commit 90f2727.

app/ota_client.py
Bodong-Yang: Resolve in commit 90f2727.

app/ota_client.py
```diff
@@ -896,48 +887,149 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
-                    finally:
-                        os.chdir(cwd)
-        shutil.copy(log_file, "./regulars_dl.log")
-        return res
+                        except Exception as e:
+                            ecb(e)
+                            break
+                    else:
+                        try:
+                            pool.apply_async(
```
obi-t4, line 980: @Bodong-Yang 
One question about apply_async.
apply_async returns AsyncResult object and my understanding is AsyncResult should be handled by caller to wait child process process.
But it seems this code doesn't wait child processes.
My understanding is correct?

Bodong-Yang: In line 994, **pool.join()** will block until all the tasks finished.

The worker process will start immediately after we issue pool.async_apply, and since the worker returns nothing, we don't need to deal with **AsyncResult**.

app/ota_client.py
```diff
@@ -896,48 +887,149 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
+                    if (
+                        int(rfile_inf.links) >= 2
+                        and rfile_inf.sha256hash
+                        not in gvar_dict["tmp-dict-hardlink_reg"]
+                    ):
+                        # block the flow until the first copy of hardlinked file is ready
+                        try:
+                            pool.apply(
+                                self._process_regular_file,
+                                (rootfs_dir, target_dir, rfile_inf),
```
obi-t4, line 973: I think adding `gvar_dict` to the _process_regular_file argument is enough.
Also I think following code is enough:
```suggestion
                       hardlink_dict = manager.dict()
                           for rfile_inf in rfiles_list:
                                ...
                                (rootfs_dir, target_dir, rfile_inf, hardlink_dict),
```

app/ota_client.py
obi-t4: I see, join waits the all processes. Understood.

obi-t4:
@Bodong-Yang 
I have another question. Setting `tmp-dict-hardlink_reg` code is necessary I think, but I can't find it. Which code does update the `gvar_dict["tmp-dict-hardlink_reg"]`?

app/ota_client.py
Bodong-Yang: Actually both ways are common methods to pass a sharing object to sub processes.

But in this case, since we already use **gvar_dict** to distribute sharing objects, to unify the sharing method, I think we should keep using the sharing objects by **gvar_dict**.

Bodong-Yang:
@obi-t4 
> I have another question. Setting tmp-dict-hardlink_reg code is necessary I think, but I can't find it. Which code does update the gvar_dict["tmp-dict-hardlink_reg"]?
Oh, that is a mistake, let me fix it.

app/ota_client.py
```diff
@@ -896,48 +887,149 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
+        main entry for paralleling processing regular files
+        """
+        try:
+            prev_inf = ""
+            # hardlinked file
+            if int(rfile_inf.links) >= 2:
+                # if the upcoming rfile entry is the first copy of hardlinked file
+                # then the prev_inf should be ""
+                prev_inf = global_var_dict["tmp-dict-hardlink_reg"].get(
+                    rfile_inf.sha256hash, ""
```
obi-t4, line 1019: prev_inf should be the RegularInf instance, so `rfile_inf` should be return instead of `rfile_inf.sha256`?

Bodong-Yang: Yes, the dict store <hash>:<RegularInf>.

app/ota_client.py
```diff
@@ -896,48 +887,151 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
+            else:  # everything is ALLRIGHT!
+                # update corresponding class attribute
+                self._process_regular_files_exit(gvar_dict)
+
+    def _process_regular_file(self, rootfs_dir, target_dir, rfile_inf: RegularInf):
+        """
+        main entry for paralleling processing regular files
+        """
+        try:
+            prev_inf = ""
```
obi-t4, line 1013: I know the previous code is `prev_inf = ""`. But I think prev_inf = None is appropriate. How do you think about it?

app/ota_client.py
```diff
@@ -896,48 +887,151 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
+
+    def _process_regular_files_exit(self, gvar_dict):
+        """
+        Used by _process_regular_files
+        Update corresponding class attributes
+
+        DO NOT call this method from other functions
+        except for _process_regular_files
+        """
+        for k, v in gvar_dict.items():
```
obi-t4, line 929: About this block, I think 
setattr(self, "rollback_dict", gvar_dict["staging-dict-_rollback_dict"])
since other function already knows 
- gvar_dict["staging-dict-_rollback_dict"] is dict-type
- having "staging-dict-_rollback_dict" key
I like simple implementation.

obi-t4 (commented):
@Bodong-Yang 
I made a comment and I'd like to hear your opinion. thanks.

obi-t4:
@Bodong-Yang 
In my ota test in VM, following error occurred:
```
ERROR:ota_client:worker[7736]: process regular file failed!
Traceback (most recent call last):
  File "/root/ota-client/app/ota_client.py", line 1031, in _process_regular_file
    self._gen_regular_file(rootfs_dir, target_dir, rfile_inf, prev_inf)
  File "/root/ota-client/app/ota_client.py", line 851, in _gen_regular_file
    os.link(src_path, dest_path)
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/bank/./usr/share/im-config/data/90_missing.rc' -> '/mnt/bank/./usr/share/im-config/data/90_missing.rc'
ERROR:ota_client:process regular files failed: [Errno 2] No such file or directory: '/mnt/bank/./usr/share/im-config/data/90_missing.rc' -> '/mnt/bank/./usr/share/im-config/data/90_missing.rc'. All sub processess terminated.
ERROR:ota_client:gen regular files error!
Traceback (most recent call last):
  File "/root/ota-client/app/ota_client.py", line 901, in _gen_regular_files
    self._process_regular_files(rootfs_dir, rfiles_list, target_dir)
  File "/root/ota-client/app/ota_client.py", line 1003, in _process_regular_files
    raise OtaError(f"process regular files failed!")
ota_client.OtaError: process regular files failed!
WARNING:root:vmlinuz or initrd is not set. This condition will be treated as an error in the future.
ERROR:ota_client:Update error!: Can not construct update bank!
```

app/ota_client.py
```diff
@@ -896,48 +887,151 @@ def _gen_regular_files(self, rootfs_dir, regulars_file, target_dir):
...
+        try:
+            prev_inf = ""
+            # hardlinked file
+            if int(rfile_inf.links) >= 2:
+                prev_inf = global_var_dict["tmp-dict-hardlink_reg"].setdefault(
+                    rfile_inf.sha256hash, rfile_inf
+                )
+                # if the upcoming rfile entry is the first copy of hardlinked file
+                # then the prev_inf should be ""
+                if prev_inf == rfile_inf:
```
obi-t4, line 1021:
```suggestion
                if prev_inf.path == rfile_inf.path:
```
since object id is different from sub-process one.

app/ota_client.py
Bodong-Yang: Yes I agree with you!

app/ota_client.py
Bodong-Yang: Sure, currently directly setattr is enough.

app/ota_client.py
Bodong-Yang: I will have a check here!

obi-t4:
another error: (line number might not be correct, since I'm adding debug line)
```
ERROR:ota_client:worker[2898]: process regular file failed!
Traceback (most recent call last):
  File "/root/ota-client/app/ota_client.py", line 1034, in _process_regular_file
    self._gen_regular_file(rootfs_dir, target_dir, rfile_inf, prev_inf)
  File "/root/ota-client/app/ota_client.py", line 881, in _gen_regular_file
    os.chown(dest_path, int(regular_inf.uid), int(regular_inf.gpid))
FileNotFoundError: [Errno 2] No such file or directory: '/mnt/bank/./home/autoware/autoware.proj/install/rosbridge_server/lib/rosbridge_server/rosbridge_tcp'
ERROR:ota_client:process regular files failed: [Errno 2] No such file or directory: '/mnt/bank/./home/autoware/autoware.proj/install/rosbridge_server/lib/rosbridge_server/rosbridge_tcp'. All sub processess terminated.
ERROR:ota_client:gen regular files error!
Traceback (most recent call last):
  File "/root/ota-client/app/ota_client.py", line 901, in _gen_regular_files
    self._process_regular_files(rootfs_dir, rfiles_list, target_dir)
  File "/root/ota-client/app/ota_client.py", line 1004, in _process_regular_files
    raise OtaError(f"process regular files failed!")
ota_client.OtaError: process regular files failed!
WARNING:root:vmlinuz or initrd is not set. This condition will be treated as an error in the future.
ERROR:ota_client:Update error!: Can not construct update bank!
```

app/ota_client.py
Bodong-Yang: Problem confirms and the cause is as your said!
This regression is fixed in commit 4b09390, and the test on my VM setup passed.
Please confirm the newest commit in your setup, thank you!

app/ota_client.py
```diff
@@ -1002,15 +1002,15 @@ def _process_regular_file(self, rootfs_dir, target_dir, rfile_inf: RegularInf):
...
+            prev_inf: RegularInf = None
             # hardlinked file
             if int(rfile_inf.links) >= 2:
                 prev_inf = global_var_dict["tmp-dict-hardlink_reg"].setdefault(
                     rfile_inf.sha256hash, rfile_inf
                 )
                 # if the upcoming rfile entry is the first copy of hardlinked file
                 # then the prev_inf should be None
-                if prev_inf == rfile_inf:
+                if prev_inf and prev_inf.path == rfile_inf.path:
```
obi-t4, line 1013: I think `prev_inf` is always evaluated as True. I think `prev_inf.path == rfile_inf.path` is enough.

Bodong-Yang: Oh I should change it to `prev_inf is not None` instead.

obi-t4:
> another error: (line number might not be correct, since I'm adding debug line)
This error was caused by another bug. I'll file a ticket, but this PR seems to be OK :+1:

obi-t4 (approved):
LGTM :+1: :tada:

PR #27, opened 2021-07-16
[T4PUB-363] Fix incorrect way of assignment for dict type

Bodong-Yang:
Deepcopy should be used when intend to assign one dict's contents to another variable.


obi-t4 (approved):
LGTM :+1:

PR #30, opened 2021-07-20
Remove unused parameter "reboot" in OtaClient._update

Bodong-Yang:
Reboot after OTA applied is being done via OtaClientService currently,
It is also better to execute reboot by OtaClientService.

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -1254,9 +1254,6 @@ class OtaClient:
         return ecu_found
 
     def update(self, ecu_update_info):
-        return self._update(ecu_update_info)
-
-    def _update(self, ecu_update_info):
         """
         OTA update execution
         """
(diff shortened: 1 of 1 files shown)


app/ota_client.py
```diff
@@ -1254,9 +1254,9 @@ def set_update_ecuinfo(self, update_info):
         return ecu_found
 
     def update(self, ecu_update_info):
-        return self._update(ecu_update_info, reboot=False)
+        return self._update(ecu_update_info)
 
-    def _update(self, ecu_update_info, reboot=False):
```
obi-t4, line 1259: Could you remove `def update()` and rename `def _update()` to `def update()`. _update is already unnecessary.

Bodong-Yang: Sure! It seems that this update -> _update design is just for passing some extra parameters. 
(P.S., This kinds of pattern also can be seen in many other places. )

obi-t4 (approved):
LGTM :+1:

PR #32, opened 2021-07-20
[T4PUB-378] Bugfix:unintended symlink is used to copy from active bank

tsugi1203:
Checking the current bank source file is symlink or not, for regular files  downloading.

https://tier4.atlassian.net/browse/T4PUB-378


obi-t4 (approved):
LGTM :+1:

PR #37, opened 2021-08-02
Do unmount before mount.

tsugi1203:
・do unmount before mount
・unmount bank after update download finished.


obi-t4 (approved):
動作問題ありませんでした LGTM

PR #42, opened 2021-08-09
[BugFix] ota-client.py: fix potential bug in _process_regular_files

Bodong-Yang:
In previous implementation, **_process_regular_files** use **error_callback** to terminate the pool if any subprocess fails.
It may cause the main process hanging in some cases.

This bug fix use a synchronized **manager.event** to notify the main process to terminate the pool.
Along with this change, a **manager.list** object is used to act as await counter, to ensure that all subprocess finished their job, instead of using **pool.join**, since we need to actively polling that **manager.event**.

TODO: a timeout counter can be added to prevent hanging.

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -899,11 +899,11 @@ class OtaClient:
                 "staging-_kernel_files": manager.dict(),
             }
             await_c = manager.list()
-            terminate_event = manager.Event()
+            terminate_event = manager.Event()  # use in ecb and _process_regular_files
 
             # default to one worker per CPU core
             with Pool(
-                initializer=self._process_regular_files_pool_init, 
+                initializer=self._process_regular_files_pool_init,
                 initargs=(gvar_dict, await_c),
             ) as pool:
 
@@ -956,8 +956,9 @@ class OtaClient:
                 # if any exception being raised in any child processes,
                 # raise it again in the main process.
                 logger.error(
-                    f"process regular files failed: {ecb_queue.get()}. All sub processess terminated."
+                    f"process regular files failed. All sub processess terminated."
                 )
+                logger.error(f"last exception: {ecb_queue.get()}")
                 raise OtaError(f"process regular files failed!")
             else:  # everything is ALLRIGHT!
                 # update corresponding class attribute
@@ -990,7 +991,7 @@ class OtaClient:
         except Exception as e:
             logger.exception(f"worker[{os.getpid()}]: process regular file failed!")
             raise e
-        
+
         # if job finished successfully
         await_counter.append(os.getpid())
 
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -948,14 +948,15 @@ class OtaClient:
                     if not ecb_queue.empty():
                         pool.terminate()
 
-                        logger.error(f"process regular files failed. All sub processess terminated.")
+                        logger.error(
+                            f"process regular files failed. All sub processess terminated."
+                        )
                         logger.error(f"last exception: {ecb_queue.get()}")
                         raise OtaError(f"process regular files failed!")
 
             # everything is ALLRIGHT!
             # update corresponding class attribute
             self._process_regular_files_exit(gvar_dict)
-
 
     def _process_regular_file(self, rootfs_dir, target_dir, rfile_inf: RegularInf):
         """
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -993,7 +993,7 @@ class OtaClient:
             raise e
 
         # if job finished successfully
-        await_counter.append(os.getpid())
+        await_counter.append(True)
 
     def _setup_regular_files(self, target_dir):
         """
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -862,7 +862,7 @@ class OtaClient:
 
         return True
 
-    def _process_regular_files_pool_init(self, gvar_dict, awc, te):
+    def _process_regular_files_pool_init(self, gvar_dict, awc):
         """
         Used by _process_regular_files
         Init the worker pool with shared variables
@@ -873,7 +873,6 @@ class OtaClient:
         global global_var_dict, await_counter, terminate_event
         global_var_dict = gvar_dict
         await_counter = awc
-        terminate_event = te
 
     def _process_regular_files_exit(self, gvar_dict):
         """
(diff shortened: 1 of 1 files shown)


app/ota_client.py
```diff
@@ -941,15 +944,21 @@ def ecb(e):
                 pool.close()
                 # wait for all tasks to complete
                 # not apply timeout currently
-                pool.join()
+                while len(await_c) < len(rfiles_list):
+                    # if one of the subprocess raise error,
+                    # terminate the whole pool
+                    if terminate_event.is_set():
+                        pool.terminate()
+                        break
```
obi-t4, line 952: This loop is busy waiting if event is not set. I think sleep is required?

app/ota_client.py
```diff
@@ -897,18 +898,20 @@ def _process_regular_files(self, rootfs_dir, rfiles_list, target_dir):
                 "staging-dict-_rollback_dict": manager.dict(),
                 "staging-_kernel_files": manager.dict(),
             }
+            await_c = manager.list()
```
obi-t4, line 901: `await` reminds me of python3 async/await.
I prefer `pid_list` or some such name.

obi-t4 (commented):
@Bodong-Yang 
Please check commnets.

obi-t4:
> It may cause the main process hanging in some cases.

If sub-process raises, sub-process terminates. So I think current implementation - pool.join() - works properly. I'd like to know the hanging condition.

obi-t4:
this code can reproduce this bug!
```python
from multiprocessing import Process, Pool, Manager
from time import sleep
from os import getpid


def sub(i):
    sleep(1)
    print(f"hello {i} {getpid()}")
    if i == 9:
        raise "hoge"


def main():
    print(f"pid: {getpid()}")

    with Manager() as manager:
        ecb_queue = manager.Queue()

        with Pool() as pool:

            def ecb(e):
                print(f"error callback {getpid()}")
                ecb_queue.put(e)
                pool.terminate()
                print("pool terminate done")

            for i in range(30):
                try:
                    print(f"apply {i}")
                    if i % 7 != 0:
                        pool.apply_async(sub, (i,), error_callback=ecb)
                    else:
                        # pool.apply_async(sub, (i,), error_callback=ecb)
                        pool.apply(sub, (i,))
                except Exception as e:
                    print("exception")
                    ecb(e)
                    break
            pool.close()
            print("before join")
            pool.join()
            print("after join")
            # while True:
            #    if not ecb_queue.empty():
            #        pool.terminate()
            #    sleep(1)

    print("finished")


main()
```

obi-t4:
> It may cause the main process hanging in some cases.

It seems that pool.termiante() before pool.close() hangs the code.

app/ota_client.py
```diff
@@ -897,18 +898,20 @@ def _process_regular_files(self, rootfs_dir, rfiles_list, target_dir):
...
             ) as pool:
 
                 # error_callback for workers
-                #   terminate all sub-processes
-                #   and pass exception to main process
+                #   signal the main process to terminate the pool
                 def ecb(e):
                     ecb_queue.put(e)
-                    pool.terminate()
+                    terminate_event.set()
```
obi-t4, line 914: I think ecb_queue can be used without terminate_event.

app/ota_client.py
Bodong-Yang: According to the document, the **is_set** method in **Event** is non-blocking method, it will instantly return True of False depends on the event status.

app/ota_client.py
Bodong-Yang: But actually this variable works as await though, it waits until all processes finish their work, so I think this naming is meaningful?

Bodong-Yang:
> > It may cause the main process hanging in some cases.
> 
> If sub-process raises, sub-process terminates. So I think current implementation - pool.join() - works properly. I'd like to know the hanging condition.

The situation is that, the previous implementation call **pool.terminate()** in the **error_callback** function, the **pool** object in the main process will not know the pool terminates, so the main process hangs forever.

Bodong-Yang:
> > It may cause the main process hanging in some cases.
> 
> It seems that pool.termiante() before pool.close() hangs the code.

Oh! I didn't thought that! I will test out in my small scale test.

app/ota_client.py
obi-t4: umm, for me, this variable is List. Waiting list asynchronously is code, not variable.
https://github.com/tier4/ota-client/pull/42/files#diff-4e43d11537bf0064fc07412772bff9e85eae2bc9334ee70950cd7217088600f0R947

app/ota_client.py
obi-t4: So if event is not set, this loop is busy waiting. Sleep in the loop is required I think.

Bodong-Yang:
> > > It may cause the main process hanging in some cases.
> > 
> > 
> > It seems that pool.termiante() before pool.close() hangs the code.
> 
> Oh! I didn't thought that! I will test out in my small scale test.



> > It may cause the main process hanging in some cases.
> 
> It seems that pool.termiante() before pool.close() hangs the code.

I also confirm this behavior in my test environment!
But I think we should still adopt the fix because we cannot guarantee that the workload dispatching for loop can finish before any worker raises error. (In my setup, the first worker being dispatched that raises error causes hanging with a size of 2000 workload.)

Besides, we also use **pool.apply** method to prepare the first copy of hardlink file.
https://github.com/Bodong-Yang/ota-client/blob/b43f72242989756fdbf3f8bc54f9bd274fd57f65/app/ota_client.py#L923-L930
The previous method(call **ecb**) in **pool.apply** worker will 100% hang the main process.
Call pool.close and pool.terminate in exception catch will also hang the main process.

app/ota_client.py
obi-t4: @Bodong-Yang 
This loop is just waiting the condition, so adding sleep releases the CPU resource. But I think it is optional.

Bodong-Yang: @obi-t4 We can add an interval between pulling with sleep though, it is not so time critical for this part of code. I think 1s interval is OK?

obi-t4 (approved):
OTA operation has been confirmed on the device.

Bodong-Yang:
Also confirmed working as expected in my setup! 
I will merge the PR now.

PR #43, opened 2021-08-17
fix grpc ecustatus message

obi-t4:
# What is this PR?
fix typo. use EcuStatusRequest instead of EcuVersionRequest in proto/otaclient.proto.

# backward compatibility
I confirmed:
- This PR's ota_client_service and current(=not fixed) grpc_client can communicate.


tsugi1203 (approved):
LGTM

PR #44, opened 2021-08-27
[T4PUB-448] refactoring stage 1

Bodong-Yang:
Main refactoring point:
* get rid of all hard-coded paths in the code base
* use **pathlib.Path** instead of os.path with pure string
* introduce **type hint** into the code base
* re-lay the code base structure, introducing **configs.py**, **constatns.py** and **exceptions.py**
* adjust pytest test files to use **tmp_path**(based on pathlib.Path) instead of **tmpdir**
* rewrite some class and methods, remove unused code

More details:
* decouple **ota_client** with **grubctl** and **ota_status**, ota_client now directly initialize grubctl instance and ota_status without passing any arguments
* (in-progress) adjust overlapping functionalities in ota_client and grubctl
* wrap **bank** inside **grubctl**, now only **grubctl** can directly call control **bank**
* rewrite **ota_boot**, introduce new implementation of OtaBoot class
* adjust and rewrite **pytest test files** to align with the new changes
* ... more minor rewrite and adjustment


app/ota_boot.py
```diff
@@ -1,78 +1,201 @@
...
+
+    # alias
+    require_check, bypass_check = True, False
+    check_passed, check_failed = True, False
+    finalize_succeeded, finalize_failed = True, False
+    bypass_finalization_check_passed, bypass_finalization_check_failed = (
+        "no_f_c_p",
+        "no_f_c_f",
+    )
+    bypass_check = "no_c"
```
obi-t4, line 31: What is those alias? Configuration? There are two `bypass_check`.

app/ota_client.py
```diff
@@ -257,91 +230,96 @@ def _header_str_to_dict(header_str):
...
-            f.flush()
+        ftmp.write(yaml.dump(update_ecu_info))
+
     shutil.move(tmp_file_name, output_file)
-    os.sync()
     return True
 
 
-class DirectoryInf:
+class _baseInf:
```
obi-t4, line 247: class name should be PascalCase

app/ota_client.py
```diff
@@ -257,91 +230,96 @@ def _header_str_to_dict(header_str):
...
 
     def __init__(self, info):
-        line = info.replace("\n", "")
-        info_list, last = _get_separated_strings(line, ",", 3)
-        self.mode = info_list[0]
-        self.uid = info_list[1]
-        self.gpid = info_list[2]
-        self.path = _decapsulate(line[last:])
+        super().__init__(info)
+        self.path = Path(self._left.strip("'"))
```
obi-t4, line 268: Can this code parse the file name which contains `'`?
When the file name contains `'`, `'` is escaped with `'\''`.
https://tier4.atlassian.net/wiki/spaces/WEB/pages/1400443313/OTA+meta+data+file+format+specification

obi-t4 (commented):
Could you confirm the comments?

app/ota_boot.py
Bodong-Yang: Fix in commit cd00535.

app/bank.py
```diff
@@ -126,173 +124,145 @@ def _get_bank_info(ota_config_file):
     return banka, bankb
 
 
-class BankInfo:
-    """
-    OTA Bank device info class
-    """
+class _baseBankInfo:
```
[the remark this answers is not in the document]
Bodong-Yang: Fix in bcb6a1e.

app/ota_client.py
Bodong-Yang: Fix in bcb6a1e.

app/ota_client.py
Bodong-Yang: I am confirming the documents to see what how to do it properly.

Bodong-Yang: Fix in 4d4eca4.
Now I use **.+** instead of **[^\']** to capture all possible characters.

Bodong-Yang: Fix in 17220ec.
Now leading \' and ending \' will only be stripped once, to allow path name like "\'\'\'\'\'abc\'\'\'\'" to be processed properly.
--------------------------
Also in 4d4eca4,
I use **.+** instead of **[^\']** to allow \' within the path.

app/bank.py
```diff
@@ -358,21 +331,20 @@ def get_next_bank_uuid_str(self):
...
-    )
+    root_dev, root_uuid, boot_dev, boot_uuid = _get_current_devfile_by_fstab(args.fstab)
     print("root dev: ", root_dev)
     print("root uuid: ", root_uuid)
     print("boot dev: ", boot_dev)
     print("boot uuid: ", boot_uuid)
 
-    print("bank a: ", bank_info._banka)
-    print("bank a uuid:", bank_info._banka_uuid)
+    cfg.FSTAB_FILE = args.fstab
```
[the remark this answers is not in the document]
Bodong-Yang: Fix in d13967d.
Now BankInfo class __init__  can accept arguments to indicate fstab_file and bank_info_file.

obi-t4 (commented):
I've left comments. Please confirm.

app/ota_status.py
```diff
@@ -1,39 +1,29 @@
...
-
-    def __init__(
-        self,
-        ota_status_file="/boot/ota/ota_status",
-        ota_rollback_file="/boot/ota/ota_rollback_count",
-    ):
-        self._status_file = ota_status_file
-        self._rollback_file = ota_rollback_file
+    _status_file = cfg.OTA_STATUS_FILE
+    _rollback_file = cfg.OTA_ROLLBACK_FILE
```
obi-t4, line 24: It is better to use CAPITAL and underscore if _status_file and _rollback_file are constant.
https://www.python.org/dev/peps/pep-0008/#constants

app/bank.py
```diff
@@ -126,173 +124,150 @@ def _get_bank_info(ota_config_file):
...
+
+
+class BankInfo(_BaseBankInfo):
+    """
+    OTA Bank device info class
+    """
+
+    def __init__(self, *, fstab_file: str = None, bank_info_file: str = None):
+        # init current bank status
+        super().__init__()
```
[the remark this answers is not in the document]
Bodong-Yang: Actually what I think is that, default we use configuration values to configure which fstab_file and bank_info_file we should use, these configuration should not be changed in the production. So I initialize these values in the base class.
Values initialized in _BaseBankInfo should not be changed, and if we want to, we should do it in the child class BankInfo.

app/ota_client.py
```diff
@@ -257,91 +230,99 @@ def _header_str_to_dict(header_str):
...
 
     def __init__(self, info):
-        line = info.replace("\n", "")
-        info_list, last = _get_separated_strings(line, ",", 3)
-        self.mode = info_list[0]
-        self.uid = info_list[1]
-        self.gpid = info_list[2]
-        self.path = _decapsulate(line[last:])
+        super().__init__(info)
+        self.path = Path(self._left[1:-1])
```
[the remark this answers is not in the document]
Bodong-Yang: Actually pathlib.Path can correctly handle escaped `'`, that is not a problem.
For example, if we have path `Path("/tmp/\'\'abc")`, pathlib can correctly identify this path as `/tmp/''abc `in the file system.

app/ota_client.py
```diff
@@ -257,91 +230,99 @@ def _header_str_to_dict(header_str):
...
 
 
-class SymbolicLinkInf:
+class SymbolicLinkInf(_BaseInf):
     """
     Symbolik link information class
     """
 
+    _pattern = re.compile(r"'(?P<link>.+)','(?P<target>.+)'")
```
obi-t4, line 277: It seems that the following path can not be parsed:
```
'/var/lib/ieee-data/iab.csv','/usr/share/ieee-data/'\'','\''iab.csv'
```
original file name contains `','` and escaped by `'\'','\''`

app/bank.py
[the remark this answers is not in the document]
Bodong-Yang: b4d6bd1: fix call parent's init method before configuration change.

What I think is that, base class configures the BankInfo via configs.py. These configurations are not mean to be changed during actual run.
But if we want to manually indicate which fstab_file and bank_info_file to run in special cases(like run bank.py standalone), we should do it in the child class.

app/ota_boot.py
```diff
@@ -82,134 +207,63 @@ def _update_finalize_ecuinfo_file(self):
...
-            checker_result_key = failure_key
+    def boot(self):
+        status: str = self._ota_status.get_ota_status()
+        logger.debug(f"Status: {status}")
+        res: tuple[OtaStatusString, OtaBootStatusString]
 
-        if state_table[status][checker_result_key].get(finalize_key) is not None:
-            # Do finalize update/rollback
-            state_table[status][checker_result_key][finalize_key]()
+        try:
```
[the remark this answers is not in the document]
obi-t4: It seems OK.

app/ota_client.py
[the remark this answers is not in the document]
obi-t4: No, OTA metadata generator escapes `'`, with `'\''`. (=One single quote is escaped with One single quote, back slash, and Two single quotes.)

obi-t4 (commented):
I think it seems OK other than single quote escaping.

app/ota_status.py
Bodong-Yang: Actually _status_file and _rollback_file are private class attribute,  we use constants from configs.py to configure them, so I think it is OK leave it as it.

app/ota_client.py
Bodong-Yang: I will take a look at it.

app/bank.py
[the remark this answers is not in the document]
obi-t4:
```
bankinfo = BankInfo("my_fstab", "my_bank_info")
print(bankinfo.fstab_file, BankInfo.fstab_file)
print(bankinfo.bank_info_file, BankInfo.bank_info_file)
```
`bankinfo.fstab_file` and `BankInfo.fstab_file` are different. Is this intended? This may be difficult to understand.

app/ota_client.py
Bodong-Yang: Fix in 9b66362.
I introduce a method `regulate_escape` to replace all `'\''` to `\'` literally before doing regex.
pathlib.Path can handle `\'` properly itself.

Bodong-Yang: Do you mean that the original file name without any escaping is `/usr/share/ieee-data/','iab.csv`?

app/bank.py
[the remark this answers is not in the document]
Bodong-Yang: > ```
> bankinfo = BankInfo("my_fstab", "my_bank_info")
> print(bankinfo.fstab_file, BankInfo.fstab_file)
> print(bankinfo.bank_info_file, BankInfo.bank_info_file)
> ```
> 
> `bankinfo.fstab_file` and `BankInfo.fstab_file` are different. Is this intended? This may be difficult to understand.

Sorry but where these codes located? It is belonged to line 327~353 in bank.py?

app/ota_client.py
obi-t4: > Do you mean that the original file name without any escaping is /usr/share/ieee-data/','iab.csv?

Yes!

app/bank.py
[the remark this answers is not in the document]
obi-t4: There is no code to access BankInfo.fstab_file. But base holds fstab_file as **class variable** and subclass also holds it as **instance variable**. This is not very clear.
I think we can do the following in the base class.
```python
def __init__(self, fstab_file=None, bank_info_file=None):
    if fstab_file:
        self.fstab_file = fstab_file
    else:
        self.fstab_file = cfg.BANK_INFO_FILE
    ...
```

app/ota_client.py
Bodong-Yang: should be fix in a113294.
I also add this special case into the test suits!

app/bank.py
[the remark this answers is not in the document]
Bodong-Yang: > There is no code to access BankInfo.fstab_file. But base holds fstab_file as **class variable** and subclass also holds it as **instance variable**. This is not very clear.
> I think we can do the following in the base class.
> 
> ```python
> def __init__(self, fstab_file=None, bank_info_file=None):
>     if fstab_file:
>         self.fstab_file = fstab_file
>     else:
>         self.fstab_file = cfg.BANK_INFO_FILE
>     ...
> ```

The idea is that, **fstab_file** and **bank_info_file** as **BaseBankInfo** class's variables are the default configuration(and the actual run configuration) for all the instances of **BankInfo**. 
In normal case, all the instances of **BankInfo** should not change these variables.
But in special case that we need to use custom **fstab_file** and **bank_info_file**, we should do that via that specific instance of BankInfo.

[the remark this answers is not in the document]
Bodong-Yang: OK I will come up with a new way to distribute the configurations to each classes, but I would like to implement it with a new PR.

app/ota_boot.py
```diff
@@ -1,78 +1,203 @@
...
+    # alias
+    # stage 1: check
+    check_passed, check_failed = True, False
+    # stage 2: finalize
+    finalize_succeeded, finalize_failed = True, False
+    # return value
+    bypass_finalization_check_passed, bypass_finalization_check_failed, bypass_check = (
+        "no_f_c_p",
+        "no_f_c_f",
+        "no_c",
```
obi-t4, line 32: I understand what is this now.
"no_f_c_p" should be "bypass_finalization_check_passed",
"no_f_c_f" should be "bypass_finalization_check_failed", and
"no_c" should be "bypass_check", to understand easily.

app/ota_status.py
obi-t4: https://www.python.org/dev/peps/pep-0008/#constants
is a general rule.
When we read a code, and if the value is CAPITAL, we can find that it is read-only.
We can find that it is private since the value name is started with an underscore, but cannot find that it is read-only.
But I think the new PR is OK about this type of fix.

app/bank.py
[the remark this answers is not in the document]
obi-t4: A new PR is OK!

app/ota_boot.py
Bodong-Yang: Yes, because we don't use the value of these variables, so I just use the abbreviation of these long sentences.

app/ota_status.py
Bodong-Yang: This will be fixed along with the PR for new way to distribute configurations.

app/ota_boot.py
Bodong-Yang: Will be updated with an separate enum class.

obi-t4 (approved):
this RP works properly!

app/ota_boot.py
Bodong-Yang: Replace by enum in commit 68b869a.
Now these variables are used as pure enums without caring about the concrete values.

obi-t4 (approved):
LGTM!

PR #47, opened 2021-09-06
Fix/add tailing slash

obi-t4:
first commit 50f7113 is revert's revert.
second commit is adding tailing slash for URL.
example for urljoin:
```python
>>> from urllib.parse import urljoin
>>> urljoin('/media/path/', 'js/foo.js')
'/media/path/js/foo.js'
```
```python
>>> urljoin('/media/path', 'js/foo.js')
'/media/js/foo.js'
>>> urljoin('/media/path', '/js/foo.js')
'/js/foo.js'
```
https://stackoverflow.com/questions/1793261/how-to-join-components-of-a-path-when-you-are-constructing-a-url-in-python


tsugi1203 (approved):
LGTM

PR #50, opened 2021-09-29
improvement for dockerfiles of e2e tests

Bodong-Yang:
Some improvement and code cleanup for e2e tests dockerfiles and docker-compose file.

* dockerfiles: combine related steps into one RUN to reduce unnecessary intermediate layers
* docker-compose file: 
  previously we mount the local code base as rw into the client container, that is not a good idea as
  the container may effect the host file system when we are running pytest. now we mount the local
  code base as ro, and copy the new code base to the working dir.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,10 +16,11 @@ services:
       context: .
       dockerfile: ./docker/client/Dockerfile
     image: ota-client
-    command: python3 -m pytest tests2 --cov=app2 --log-cli-level=INFO
+    command: |
+      bash -c "cp -av /ota-client-origin/* /ota-client && python3 -m pytest tests2 --cov=app2 --log-cli-level=INFO"
     container_name: ota-client
     volumes:
-      - .:/ota-client
+      - .:/ota-client-origin:ro
     depends_on:
       - server
 
--- a/docker/server/Dockerfile
+++ b/docker/server/Dockerfile
@@ -3,23 +3,31 @@ SHELL ["/bin/bash", "-c"]
 ENV DEBIAN_FRONTEND=noninteractive
 ARG KERNEL_VERSION="5.8.0-53-generic"
 
-RUN apt-get update
-RUN apt-get install -y linux-image-${KERNEL_VERSION}
-RUN apt-get install -y wget git python3 python3-pip
+RUN \
+    apt-get update && \
+    apt-get install -y linux-image-${KERNEL_VERSION} && \
+    apt-get install -y wget git python3 python3-pip
 
-RUN mkdir /ota-server
 WORKDIR /ota-server
 
+# prepare environment
+RUN \
+    wget http://cdimage.ubuntu.com/ubuntu-base/releases/20.04/release/ubuntu-base-20.04.1-base-amd64.tar.gz && \
+    mkdir -p data && \
+    tar zxf ubuntu-base-20.04.1-base-amd64.tar.gz -C data && \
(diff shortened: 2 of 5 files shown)


docker-compose.yml
```diff
@@ -16,10 +16,11 @@ services:
       context: .
       dockerfile: ./docker/client/Dockerfile
     image: ota-client
-    command: python3 -m pytest tests2 --cov=app2 --log-cli-level=INFO
+    command: |
+      bash -c "cp -av /ota-client-origin/* /ota-client && python3 -m pytest tests2 --cov=app2 --log-cli-level=INFO"
```
obi-t4, line 20: :+1:

docker/server/Dockerfile
```diff
@@ -3,23 +3,31 @@ SHELL ["/bin/bash", "-c"]
 ENV DEBIAN_FRONTEND=noninteractive
 ARG KERNEL_VERSION="5.8.0-53-generic"
 
-RUN apt-get update
-RUN apt-get install -y linux-image-${KERNEL_VERSION}
-RUN apt-get install -y wget git python3 python3-pip
+RUN \
+    apt-get update && \
+    apt-get install -y linux-image-${KERNEL_VERSION} && \
+    apt-get install -y wget git python3 python3-pip
```
obi-t4, line 9: I think this modification is OK, but ...

docker/server/Dockerfile
```diff
@@ -3,23 +3,31 @@ SHELL ["/bin/bash", "-c"]
...
-RUN cp /boot/vmlinuz-${KERNEL_VERSION} /boot/initrd.img-${KERNEL_VERSION} /boot/config-${KERNEL_VERSION} /boot/System.map-${KERNEL_VERSION} data/boot
-RUN git clone https://github.com/tier4/ota-metadata
-RUN python3 -m pip install -r ota-metadata/metadata/ota_metadata/requirements.txt
-RUN python3 ota-metadata/metadata/ota_metadata/metadata_gen.py --target-dir data --ignore-file ota-metadata/metadata/ignore.txt 
-RUN ./ota-metadata/metadata/key-gen.sh
-RUN python3 ota-metadata/metadata/ota_metadata/metadata_sign.py --sign-key privatekey.pem --cert-file certificate.pem --persistent-file ota-metadata/metadata/persistents.txt --rootfs-directory data
-RUN cp ota-metadata/metadata/persistents.txt .
+# create ota-baseimage
+RUN \
+    git clone https://github.com/tier4/ota-metadata && \
```
obi-t4, line 22: By this modification, docker cache cannot be used when `RUN` is failed.
I'd like to know this merit.

obi-t4: >  to reduce unnecessary intermediate layers

I think for the Dockerfile developer, the layers are necessary to develop. When I try to add a new `RUN` line, I think I want to use layer cache to debug faster.

obi-t4: > apt update/upgrade

is OK for apt cache problem.

> break up the create ota-image step into more parts

Yes. I'd like to prioritize development productivity compare to generating intermediate layers.

Bodong-Yang: in commit 40843f2, the second step breaks up into prepare metadata build environment, and creating ota-baseimage

obi-t4: I think `RUN` for each command is enough. It means the original Dockerfile is enough (other than apt-get update&&apt-get update).
What do you think?

Bodong-Yang: OK I think one line per RUN is OK start from generating the metadata, since the previous steps are independent from our code.

Please check the commit bdf4d85.

obi-t4 (approved):
LGTM

PR #51, opened 2021-09-29
ota_client.py: improve _create_regular_files

Bodong-Yang:
In the previous implementation, when we handle the first copy of hardlink file, the main process will be blocked.
In the new implementation, handling first copy of hardlink file is done in the _create_regular_file method, so only a subprocess will be blocked.

More details as follow:
* hardlink_dict is only used in the main process now, previous_reginf is passed instead of the whole dict,
* hardlink_dict is sha256hash[tuple[reginf, event]] now, a **multiprocessing.Event** object is used to ensure the first_copy.
* process pool dispatching logics is placed back to _create_regular_files method.

### DIFF
--- a/app2/ota_client.py
+++ b/app2/ota_client.py
@@ -28,8 +28,10 @@ def file_sha256(filename: Path) -> str:
     with open(filename, "rb") as f:
         return sha256(f.read()).hexdigest()
 
+
 def verify_file(filename: Path, filehash: str) -> bool:
     return file_sha256(filename) == filehash
+
 
 class _BaseInf:
     _base_pattern = re.compile(
@@ -376,7 +378,7 @@ class OtaClient:
     def _create_regular_files(self, url_base: str, cookies, list_file, standby_path):
         lines = open(list_file).read().splitlines()
         reginf_list = [RegularInf(l) for l in lines]
-        self._total_regular_files = len(reginf_list) # via setter
+        self._total_regular_files = len(reginf_list)  # via setter
 
         with Manager() as manager:
             error_queue = manager.Queue()
@@ -388,26 +390,27 @@ class OtaClient:
 
             boot_standby_path = self._ota_status.get_standby_boot_partition_path()
             _create_regfile_func = partial(
-                self._create_regular_file, 
-                url_base=url_base, 
-                cookies=cookies, 
-                standby_path=standby_path, 
+                self._create_regular_file,
+                url_base=url_base,
+                cookies=cookies,
+                standby_path=standby_path,
                 processed_list=processed_list,
-                boot_standby_path=boot_standby_path
-                )
+                boot_standby_path=boot_standby_path,
+            )
 
             with Pool() as pool:
-                hardlink_dict = dict() # sha256hash[tuple[reginf, event]]
+                hardlink_dict = dict()  # sha256hash[tuple[reginf, event]]
 
                 for reginf in reginf_list:
                     if reginf.nlink >= 2:
                         first_copy = not reginf.sha256hash in hardlink_dict
                         prev_reginf, event = hardlink_dict.setdefault(
-                            reginf.sha256hash, (reginf, manager.Event()))
+                            reginf.sha256hash, (reginf, manager.Event())
+                        )
 
                         pool.apply_async(
                             _create_regfile_func,
-                            (reginf, prev_reginf), 
+                            (reginf, prev_reginf),
                             {"event": event, "first_copy": first_copy},
                             error_callback=error_callback,
                         )
@@ -417,14 +420,14 @@ class OtaClient:
                             (reginf,),
                             error_callback=error_callback,
                         )
-                
+
                 pool.close()
                 while len(processed_list) < len(reginf_list):
                     if not error_queue.empty():
                         error = error_queue.get()
                         pool.terminate()
                         raise error
-                    time.sleep(2) # set pulling interval to 2 seconds
+                    time.sleep(2)  # set pulling interval to 2 seconds
 
     # NOTE:
     # _create_regular_file should be static to be used from pool.apply_async,
@@ -432,7 +435,7 @@ class OtaClient:
     @staticmethod
     def _create_regular_file(
         reginf: RegularInf,
-        prev_reginf: RegularInf=None,
+        prev_reginf: RegularInf = None,
         *,
         # required options
         url_base: str,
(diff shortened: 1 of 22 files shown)


obi-t4:
@Bodong-Yang 
Can you revert auto generated files - app2/otaclient*_pb2*.py - before review?

Bodong-Yang:
> @Bodong-Yang Can you revert auto generated files - app2/otaclient*_pb2*.py - before review?

Which commit should I revert to?

obi-t4:
> Which commit should I revert to?

app2/otaclient*_pb2*.py
Your commit changes 4 files. but 3 are unnecessary.

Bodong-Yang:
> > Which commit should I revert to?
> 
> app2/otaclient*_pb2*.py Your commit changes 4 files. but 3 are unnecessary.

Oh that is cause by the git merge, I should replace these files with upstream version instead of merge them!

obi-t4:
> Oh that is cause by the git merge

Your git or editor setting or something?

Bodong-Yang:
> > Oh that is cause by the git merge
> 
> Your git or editor setting or something?

resolved in the latest commit.

app2/ota_client.py
```diff
@@ -370,117 +375,103 @@ def _create_symbolic_links(self, list_file, standby_path):
...
+                    else:
+                        pool.apply_async(
+                            _create_regfile_func,
+                            (reginf,),
+                            error_callback=error_callback,
+                        )
+
                 pool.close()
                 while len(processed_list) < len(reginf_list):
-                    self._regular_files_processed = len(processed_list)  # via setter
```
obi-t4, line 399: this line is required.

app2/ota_client.py
```diff
@@ -370,117 +375,103 @@ def _create_symbolic_links(self, list_file, standby_path):
...
-                    hardlink_dict,
-                    processed_list,
-                    pool.apply_async,
-                    error_callback,
-                )
+                hardlink_dict = dict()  # sha256hash[tuple[reginf, event]]
+
+                for reginf in reginf_list:
+                    if reginf.nlink >= 2:
+                        first_copy = not reginf.sha256hash in hardlink_dict
```
obi-t4, line 406: first_copy seems redundant. I think we can remove this line, and we can evaluate `first_copy` boolean value in the function
_create_regular_file:
`first_copy = prev_reginf.path == reginf.path`

app2/ota_client.py
```diff
@@ -370,117 +375,103 @@ def _create_symbolic_links(self, list_file, standby_path):
...
+                        prev_reginf, event = hardlink_dict.setdefault(
+                            reginf.sha256hash, (reginf, manager.Event())
+                        )
+
+                        pool.apply_async(
+                            _create_regfile_func,
+                            (reginf, prev_reginf),
```
obi-t4, line 413: Can you please add comments? like:
`# the positional argument in a tuple`
`# the keyword arguments in a dictionary`
(Not all of us are familiar with applying_async usage :smiley: )

obi-t4 (commented):
I left some comments, but basically, this PR is OK! Thanks!

app2/ota_client.py
Bodong-Yang: Oh I miss understood the usage of this line, let me fix it.

obi-t4: > to keep track of how many files are processed?

Yes!

app2/ota_client.py
Bodong-Yang: Yes you are right!

Updated on commit 69cc179.

app2/ota_client.py
Bodong-Yang: Sure!

obi-t4 (approved):
完璧 :+1:

Bodong-Yang:
Oh by the way, I also think that we can use multiprocessing to accelerate the processing of regular_file.txt, since it has thousands of records.

obi-t4:
> Oh by the way, I also think that we can use multiprocessing to accelerate the processing of regular_file.txt, since it has thousands of records.

We have to measure it. If the consumption time to process regular_file.txt is a few percent, I think we don't need to optimize it.

Bodong-Yang:
> > Oh by the way, I also think that we can use multiprocessing to accelerate the processing of regular_file.txt, since it has thousands of records.
> 
> We have to measure it. If the consumption time to process regular_file.txt is a few percent, I think we don't need to optimize it.

Actually this optimization makes the code simpler, 
please check the commit 046e40a.

obi-t4:
> please check the commit 046e40a.

sorry new code should be separated PR. Or, we can't close this PR :pray:

Bodong-Yang:
> > please check the commit [046e40a](https://github.com/tier4/ota-client/commit/046e40a85989517c3081ffd05616a7d5581fa035).
> 
> sorry new code should be separated PR. Or, we can't close this PR pray

I think this change can be within this PR since the change is minor actually.
The code do the same thing as before.

app2/ota_client.py
```diff
@@ -370,117 +375,113 @@ def _create_symbolic_links(self, list_file, standby_path):
...
+                            {"hardlink_event": event},
+                            error_callback=error_callback,
+                        )
+                    else:
+                        pool.apply_async(
+                            _create_regfile_func,
+                            (reginf,),
+                            error_callback=error_callback,
+                        )
+                self._total_regular_files = total_regular_files
```
obi-t4, line 431: `self._total_regular_files is total` should be total number of regular files.

Bodong-Yang: Yes, I use this counter variable `total_regular_files ` to count the number of regular files in the for loop.

Bodong-Yang: > `self._total_regular_files is total` should be total number of regular files. Not number of regular files processed.

Yes I am doing what you said,
```
for reginf in reginf_list:
    total_regular_files += 1
```
is the same as `len(reginf_list)` in previous implementation.

Bodong-Yang: The reason why I count it in the for loop is as follow:
```
                # imap_unordered return a lazy itorator without blocking
                # so we cannot use len() over it
```

obi-t4 (changes requested):
please modify total_regular_files.

app2/ota_client.py
obi-t4: :thinking: 
previous implementation:
```
self._total_regular_files = len(reginf_list) # via setter
````
I think you're saying `self._regular_files_processed`?

Bodong-Yang:
> please modify total_regular_files.

Fix in commit f0680f4.

obi-t4 (approved):
LGTM!

PR #53, opened 2021-10-01
support async update

obi-t4:


Bodong-Yang (approved):
This implementation should work!

PR #54, opened 2021-10-06
fix grpc type and some bug

obi-t4:


app2/ota_client.py
```diff
@@ -138,9 +139,9 @@ def __init__(self):
 
         self._executor = ThreadPoolExecutor(max_workers=1)
 
-    def update(self, version, url_base, cookies, blocking=False):
+    def update(self, version, url_base, cookies: str, blocking=False):
         try:
-            self._update(version, url_base, cookies, blocking)
+            self._update(version, url_base, json.loads(cookies), blocking)
```
Bodong-Yang, line 144: `json.loads` will raise `JSONDecodeError` if decoding failed, I think we should handle this?

Bodong-Yang, line 144:
```suggestion
            try:
                loaded_cookies = json.loads(a)
                self._update(version, url_base, loaded_cookies, blocking)
            exception Exception as e:
                log.debug("failed to load cookies!")
                # do more things to handle the error...
```

app2/ota_client.py
```diff
@@ -138,9 +139,9 @@ def __init__(self):
...
             return self._result_ok()
         except OtaErrorRecoverable as e:
```
Bodong-Yang, line 146: The json.loads may also raise Exception, I think we should also catch it?

obi-t4:
@Bodong-Yang 
43a5037 handles the json.loads exception. thanks!

Bodong-Yang (approved):
LGTM!

PR #55, opened 2021-10-06
[T4PUB-523] test improvement

obi-t4:
- adding procedure to run test individually
- adding ota image generation to run ota on actual environment.

### DIFF
--- a/app2/ota_metadata.py
+++ b/app2/ota_metadata.py
@@ -36,7 +36,7 @@ class OtaMetadata:
     If there is no root or intermediate certificate, certification verification
     is not performed.
     """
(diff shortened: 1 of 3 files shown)


app2/ota_metadata.py
```diff
@@ -36,7 +36,7 @@ class OtaMetadata:
     If there is no root or intermediate certificate, certification verification
     is not performed.
     """
-    CERTS_DIR = Path(__file__).parent / "certs"
+    CERTS_DIR = Path(__file__).parent.parent / "certs"
```
obi-t4, line 39: certs directory is changed.

PR #56, opened 2021-10-07
merge ota client v2 to main

obi-t4:
To merge ota-client v2 to master as is.


Bodong-Yang (approved):

obi-t4:
> I think it is also a good idea to leave a copy of legacy code in the repo,

We can retrieve by commit hash or if we can set tag, we can retrieve old ota-client easier.

PR #81, opened 2021-11-18
[T4PB-13101] support persistent if src file -> dst dir, src dir -> dst file

obi-t4:
# What is this PR?
when /opt/autoware/maps is symlink(or file in this case) and /mnt/standby/opt/autoware/maps is directory, persistent process failed.
This PR fixes this issue.
And also this PR support this scenario: if src is directory and dst is file or symlink, 

# issue
https://tier4.atlassian.net/browse/T4PB-13101

# test
OTA tested on the actual system.


Bodong-Yang (approved):
It should work as expected!

PR #83, opened 2021-11-24
[T4PUB-560] Implement ota logics for subECU with platform awareness

Bodong-Yang:
# Description
Implements subECU ota-client logics for subECU **(currently only support rqx_580(chip_id: 0x19) platform)**.
Interfaces for bootctrl is introduced for splitting the platform specific codes.
Now ota-client is platform aware,  ota-client will choose to run specific version according to platform(grub or cboot/extlinux).

# Implementation details
1. boot_control.py: interface definition for boot switching
2. ota_client.py: BaseOtaClient is introduced to hold platform neutral codes of ota_client. Codes for different platform are integrated by platform specific mixins (GrubControlMixin, CBootControlMixin)
3. extlinux_control.py: implementation of subECU boot switching logics
4. extlinux_control.py: model/chip_id detection, nv_boot_control.conf 's present ensured
5. grub_ota_partition.py: add a wrapper class GrubControlMixin as an adapter for the bootctrl interface
6. configs.py: refactored
7. test units files are updated accordingly
8. other minor fixes and improvements

# Current status
- [x] design ota-client interface and bootctrl interface
- [x] implement interfaces for grub platform by wrapping grub platform codes
- [x] implement subecu bootctrl logics, integrate with general ota-client codes
- [x] test ota-client on grub platform by unit test files
- [x] test ota-client on subECU
- [x] merge latest upstream changes (up to PR#82)
- [ ] test files for subECU cboot_control 

# Issue
https://tier4.atlassian.net/browse/T4PUB-560


app/boot_control.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+
+    _boot_control = None
+    _ota_status = None
+    _mount_point = None
+
+    def initialize_ota_status(self):
+        pass
+
+    def write_standby_ota_status(self, status):
```
obi-t4, line 17: `store_standby_ota_status`, I prefer.
I'd like to use corresponding terms:
write <-> read
load <-> store (retrieve <-> store)
get <-> set
get <-> put (direction)
enter <-> leave
start <-> stop
begin <-> end
etc.

app/configs.py
```diff
@@ -1,45 +1,131 @@
...
+    if platform == "grub":
+        return GrubControlConfig()
+    elif platform == "cboot":
+        return CBootControlConfig()
+
+
+cboot_cfg = CBootControlConfig()
+grub_cfg = GrubControlConfig()
+
+Config = create_config(_detect_platform())
```
obi-t4, line 131: cboot_cfg is an instance, grub_cfg is also an instance, and Config is also an instance.
Only one instance is required I think.

obi-t4: `Config` should be `config` since this `Config` is not Class name, this is instance (=variable).

app/extlinux_control.py
```diff
@@ -0,0 +1,659 @@
...
+def _subprocess_call(cmd: str, *, raise_exception=False):
+    try:
+        logger.debug(f"cmd: {cmd}")
+        subprocess.check_call(shlex.split(cmd), stdout=subprocess.DEVNULL)
+    except subprocess.CalledProcessError as e:
+        logger.warn(
+            msg=f"command failed(exit-code: {e.returncode} stderr: {e.stderr} stdout: {e.stdout}): {cmd}"
+        )
+        if raise_exception:
+            raise e
```
obi-t4, line 41: `e` can be removed to re-raise same exception.

app/extlinux_control.py
```diff
@@ -0,0 +1,659 @@
...
+    @classmethod
+    def get_dev_by_partlabel(cls, partlabel: str) -> str:
+        return cls._findfs("PARTLABEL", partlabel)
+
+    @classmethod
+    def get_dev_by_partuuid(cls, partuuid: str) -> str:
+        return cls._findfs("PARTUUID", partuuid)
+
+
+class nvbootctrl:
```
obi-t4, line 92: class name should be PascalCase.

app/configs.py
```diff
@@ -1,45 +1,131 @@
...
+        boot_dir = Path("/boot")
+        etc_dir = Path("/etc")
+        mount_point = Path("/mnt/standby")
+
+        fstab_file = etc_dir / "fstab"
+        ecu_info_file = boot_dir / "ota" / "ecu_info.yaml"
+        passwd_file = etc_dir / "passwd"
+        group_file = etc_dir / "group"
+
+        ota_partition_folder = Path("ota-partition")
```
obi-t4, line 37: The term `folder` is not used in the ota-client code. `dir` is used.

app/configs.py
```diff
@@ -1,45 +1,131 @@
...
+
+# helper function to detect platform
+def _detect_platform():
+    if platform.machine() == "x86_64" or platform.processor == "x86_64":
+        return "grub"
+    elif platform.machine() == "aarch64" or platform.processor == "aarch64":
+        return "cboot"
+
+
+def create_config(platform):
```
obi-t4, line 121: if `create_config` is private function, use `_create_config`.

obi-t4 (commented):
I left initial comments. but sorry, I'm not finished yet.

app/ota_client.py
```diff
@@ -664,6 +717,52 @@ def _copy_persistent_files(self, list_file, standby_path):
...
+        self.check_rollback_status()
+        self.set_ota_status(OtaStatus.ROLLBACKING)
+        self.write_standby_ota_status(OtaStatus.ROLLBACKING)
+        self.boot_ctrl_pre_rollback()
+
+    def leave_rollbacking(self):
+        self.boot_ctrl_post_rollback()
+
+
+def ota_client_instance():
```
obi-t4, line 746: This function doesn't return instance. returns class. So the function name should be ota_client_class.

obi-t4 (commented):
Can you check the comments? (review is not finished yet!)

app/configs.py
Bodong-Yang: > cboot_cfg is an instance, grub_cfg is also an instance, and Config is also an instance.
> Only one instance is required I think. 
Oh cboot_cfg and grub_cfg are specifically for platform specific codes, like I use grub_cfg for grub_control.py, and cboot_cfg for extlinux.cfg, to avoid using wrong configs for platform specific codes.

app/configs.py
Bodong-Yang: Actually it is intended as a public function, so that we can generate platform specific configs by ourselves if needed, like in pytest test files.

app/extlinux_control.py
```diff
@@ -0,0 +1,659 @@
...
+        pa = re.compile(r"\broot=(?P<rdev>[\w=-]*)\b")
+        ma = pa.search(_subprocess_check_output("cat /proc/cmdline")).group("rdev")
+        uuid = ma.split("=")[-1]
+
+        return Path(helperFuncsWrapper.get_dev_by_partuuid(uuid)).resolve(
+            strict=True
+        ) == Path(dev).resolve(strict=True)
+
+    @classmethod
+    def get_current_slot_dev(cls) -> str:
```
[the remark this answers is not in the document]
Bodong-Yang: Yes we have 3 statuses.
`active` is the slot that we **should** use(the latest updated slot).
`current` is the slot that the ota-client is running on.
`standby` slot is the slot that ota-client is not running on, and the target that we apply ota-update to.

app/configs.py
obi-t4: OK, `create_config` should be return either cboot_cfg intance or grub_cfg instance, not to create new instance.

app/configs.py
obi-t4: Umm, create_config is not called from any outside. When we need to call from outside, we can change to the public.
Currently private is enough.

app/extlinux_control.py
```diff
@@ -0,0 +1,659 @@
...
+    def get_entry(self, name: str) -> dict:
+        """
+        return the copy of specific entry
+        """
+        return self._entry.get(name, default=dict()).copy()
+
+    def get_default_entry(self) -> dict:
+        name = self._heading.get("DEFAULT")
+        if name in self._entry:
+            return self._entry.get(name, default=dict()).copy()
```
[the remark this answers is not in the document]
Bodong-Yang: Oh, that is a bug I should fix!

app/ota_client.py
```diff
@@ -18,11 +18,14 @@
 from requests.exceptions import RequestException
 from retrying import retry
 
-from ota_status import OtaStatusControl, OtaStatus
+from extlinux_control import CBootControl, CBootControlMixin
+from ota_client_interface import OtaClientInterface
+from grub_ota_partition import GrubControlMixin, OtaPartitionFile
```
obi-t4, line 23: `from grub_ota_partition import GrubControlMixin, OtaPartitionFile` can be remove to https://github.com/tier4/ota-client/pull/83/files#diff-4e43d11537bf0064fc07412772bff9e85eae2bc9334ee70950cd7217088600f0R746
not to create unintended config by `config = create_config(_detect_platform())`

app/grub_ota_partition.py
```diff
@@ -446,3 +450,80 @@ def _fsync(self, path):
         finally:
             if fd:
                 os.close(fd)
+
+
+######## bootcontrol adapter for grub_control ########
+class GrubControlMixin(BootControlInterface):
+    def __init__(self):
+        """
+        attributes that needed for this mixin to work
```
obi-t4, line 459: Is this __init__() necessary?
Mixin class uses attributes that are defined by another class.
The mixin class is the class that cannot exist by itself.

app/configs.py
```diff
@@ -1,45 +1,135 @@
...
+        )
+    return
+
+
+def create_config(platform):
+    if platform == "grub":
+        return GrubControlConfig()
+    elif platform == "cboot":
+        return CBootControlConfig()
+
```
obi-t4, line 133: It is better to return None explicitly.
```suggestion
    return None
```

app/nv_boot_control.conf.rqx_580
```diff
@@ -0,0 +1,5 @@
+TNSPEC 2888-400-0004-L.0-1-2-rqx_580-mmcblk0p1
+COMPATIBLE_SPEC 2888-400-0004--1-2-rqx_580-mmcblk0p1
+TEGRA_CHIPID 0x19
+TEGRA_OTA_BOOT_DEVICE /dev/mmcblk0boot0
+TEGRA_OTA_GPT_DEVICE /dev/mmcblk0boot1
```
obi-t4, line 5: add a new line.

app/ota_status.py
```diff
@@ -21,25 +18,17 @@ class OtaStatus(Enum):
...
 
 
-class OtaStatusControl:
+class OtaStatusControlMixin:
     def __init__(self):
-        self._ota_partition = OtaPartitionFile()
-        self._ota_status = self._initialize_ota_status()
-        logger.info(f"ota_status={self._ota_status}")
+        # ota_status will be initialized by ota_client
+        self._ota_status: OtaStatus = None
```
obi-t4, line 24: I think this attribute initialization is not required since this class is mixin.

obi-t4 (commented):
I'm also reviewing now but I'll send some comments.
13/19 are reviewed.

app/grub_control.py
```diff
@@ -1,15 +1,17 @@
...
 from pathlib import Path
 from pprint import pformat
 
-from ota_error import OtaErrorUnrecoverable
-import configs as cfg
 import log_util
+import configs
+from ota_error import OtaErrorUnrecoverable
+
+cfg = configs.create_config("grub")
```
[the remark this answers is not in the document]
Bodong-Yang: Ah the reason for this is that grub_control must use GrubControlConfig, so we explicitly create a GrubControlConfig instance for grub_control module to avoid potential problems caused by initializing grub_control module with mismatching config.

app/grub_ota_partition.py
Bodong-Yang: This __init__ works like a placeholder, indicating that this mixin needs these attributes from the concrete class.
It can remind us to initialize these attributes when initializing the OtaClient, and also hint the autocomplete in IDE.

app/ota_status.py
Bodong-Yang: The same as other mixin, this __init__ method is just a placeholder, we don't need to call it.
_ota_status will be initialized in the OtaClient's __init__ method.

app/ota_client.py
Bodong-Yang: Update in commit cdf6688.
Right now platform specific modules will only be imported when needed.

app/ota_status.py
obi-t4: @Bodong-Yang 
https://stackoverflow.com/questions/36690588/should-mixins-use-parent-attributes
Mixin class should use base class attribute.

obi-t4 (commented):
I'd like to confirm about failure handling.
In case of shutdown (system shutdown or ota-client process shutdown) during updating,
after ota-client is restarted, update failure status should be hold.
Grub part is OK, but how cboot handle it?
cboot retrieve ota_status from current slot. before updating, cboot save ota_status to standby slot. failure status cannot be retrieved I think.

app/ota_status.py
Bodong-Yang: > 
yes I am using base class attribute here.

obi-t4: base class is initialzing this attribute, so this __init__ function should be removed from mixin class I think.

app/ota_client.py
```diff
@@ -664,6 +715,99 @@ def _copy_persistent_files(self, list_file, standby_path):
...
+        self.boot_ctrl_pre_update(version)
+        self.set_ota_status(OtaStatus.UPDATING)
+        self.store_standby_ota_status(OtaStatus.UPDATING)
+        logger.debug("finished pre-update setup")
+
+    def leave_update(self):
+        logger.debug("post-update setup...")
+        self.boot_ctrl_post_update()
+
+    def enter_rollbacking(self):
```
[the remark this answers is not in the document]
Bodong-Yang: sure!

app/grub_ota_partition.py
```diff
@@ -446,3 +450,80 @@ def _fsync(self, path):
...
+        return self._boot_control.load_ota_status()
+
+    def get_standby_boot_partition_path(self) -> Path:
+        return self._boot_control.get_standby_boot_partition_path()
+
+    def get_version(self):
+        return self._boot_control.load_ota_version()
+
+    def boot_ctrl_pre_update(self, version):
+        self.store_standby_ota_status(OtaStatus.UPDATING)
```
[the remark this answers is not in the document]
Bodong-Yang: Ah that is true! I will remove the redundant codes.

app/nv_boot_control.conf.rqx_580
```diff
@@ -0,0 +1,5 @@
+TNSPEC 2888-400-0004-L.0-1-2-rqx_580-mmcblk0p1
```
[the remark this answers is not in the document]
Bodong-Yang: I think that is a necessary safety feature, as the nv_boot_control will not work without this file present, then breaking the ota_client's functionality.
In the worst scenario that this file is not downloaded or not be able to downloaded, we can still deal with this edge condition without relying on external resources.

app/extlinux_control.py
```diff
@@ -0,0 +1,681 @@
...
+        slot = Nvbootctrl.get_current_slot()
+        return Nvbootctrl.is_slot_marked_successful(slot)
+
+    @classmethod
+    def reboot(cls):
+        subprocess.check_call(shlex.split("reboot"))
+
+
+class CBootControlMixin(BootControlInterface):
+    def __init__(self):
```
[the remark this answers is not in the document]
Bodong-Yang: That is a placeholder for hinting the developer and the autocomplete plugin.
We will don't need to call this init method.

Bodong-Yang:
> I'd like to confirm about failure handling. In case of shutdown (system shutdown or ota-client process shutdown) during updating, after ota-client is restarted, update failure status should be hold. Grub part is OK, but how cboot handle it? cboot retrieve ota_status from current slot. before updating, cboot save ota_status to standby slot. failure status cannot be retrieved I think.

That is the reason I introduce **/boot/ota-status/slot_in_use** file in cboot handling.
When ota-client starts to do the ota update, the ota-client will store the updating slot number into the slot_in_use file.
Please check extlinux_control.py:CBootControlMixin:_is_switching_boot method, only when both nvboot status, ota_status, and slot_in_use file are in expected status, we will think the recent ota update is successful, otherwise the recent ota update is failed.

Think of the following condition, the ota-update reboot fails to reboot the system to the standby slot B, the system boot back to slot A(where the ota-update is inited), although the ota-status in slot A is SUCCESS, the **slot_in_use file** indicates slot B, which indicates a failed ota-update reboot.

app/nv_boot_control.conf.rqx_580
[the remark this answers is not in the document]
obi-t4: Umm, this file is required but there is a number of necessary files. e.g. /bin/bash.
If we don't prepare this file in the OTA image, it is a bug.
It is better for ota client to use files that ota list indicates.

obi-t4:
> That is the reason I introduce /boot/ota-status/slot_in_use file in cboot handling.

Sorry, I'm reading the wrong file. I'd like to check the sequence. Thanks for the explanation!

obi-t4 (commented):
All review has been finished.
If all comments have been resolved, I think the merge can be ready!

app/nv_boot_control.conf.rqx_580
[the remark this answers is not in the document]
Bodong-Yang: OK! Should I remove codes related to nv_boot_control.conf?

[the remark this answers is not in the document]
obi-t4: Yes please, we don't need that code.

app/extlinux_control.py
```diff
@@ -0,0 +1,684 @@
...
+from pathlib import Path
+from typing import Tuple
+
+import log_util
+import configs
+from ota_error import OtaErrorUnrecoverable
+from ota_status import OtaStatus
+from ota_client_interface import BootControlInterface
+
+cfg = configs.create_config("cboot")
```
obi-t4, line 14: same here.
```
from configs import config as cfg

assert cfg.PLATFORM == "cboot"
```

Bodong-Yang: updated in commit ce8d794.

app/nv_boot_control.conf.rqx_580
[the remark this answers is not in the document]
Bodong-Yang: cleanup in commit 73031ab.

app/configs.py
```diff
@@ -1,45 +1,135 @@
...
+    PLATFORM = "grub"
+
+    def __init__(self):
+        super().__init__()
+
+        self.grub_dir = self.BOOT_DIR / "grub"
+        self.grub_cfg_file = self.grub_dir / "grub.cfg"
+        self.custom_cfg_file = self.grub_dir / "custom.cfg"
+        self.default_grub_file = self.grub_dir / "default/grub"
```
obi-t4, line 77:
```suggestion
        self.default_grub_file = self.ETC_DIR / "default/grub"
```

Bodong-Yang: configs.py: fix default_grub_file path

obi-t4:
@Bodong-Yang 
Could you extend github actions timeout from 5 to 10 minutes? (CI is failing by timeout)

obi-t4 (approved):
LGTM :tada:

PR #87, opened 2021-12-07
Refactor of OtaClientStatistics collecting logic

Bodong-Yang:
## Description
The previous implementation of OtaClientStatistics collecting logic may cause the statistics not being updated properly or missing updates when grpc *status* API being called with very short interval.

This PR fixes this problem by refactoring the statistics collecting logic, and improves the performance. The format of OtaStatistics is not changed.

Unit tests are passed for this PR.

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -6,15 +6,13 @@ import re
 import os
 import time
 import json
-import operator
 from contextlib import contextmanager
 from hashlib import sha256
 from pathlib import Path
 from json.decoder import JSONDecodeError
 from multiprocessing import Pool, Manager
 from threading import Lock
-from functools import partial, reduce
-from collections import Counter
+from functools import partial
 from enum import Enum, unique
 from requests.exceptions import RequestException
 from retrying import retry
@@ -247,11 +245,9 @@ class OtaClientStatistics(object):
         self._slot = self._init_statistics_storage()
 
     @contextmanager
-    def modify_storage(self):
+    def acquire_staging_storage(self):
         """
-        update the whole storage
-
-        yield a staging storage area for thread-safe modifying the storage
+        acquire a staging storage for updating the slot atomically and thread-safely
         """
         lock_acquired = False
         try:
@@ -567,7 +563,7 @@ class _BaseOtaClient(OtaStatusControlMixin, OtaClientInterface):
             {"errors": int}  # number of errors that occurred when downloading.
         """
         all_processed = len(sts)
-        with self._statistics.modify_storage() as staging_storage:
+        with self._statistics.acquire_staging_storage() as staging_storage:
             # failed to acquire lock for any reason
             if staging_storage is None:
                 return
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -560,17 +560,18 @@ class _BaseOtaClient(OtaStatusControlMixin, OtaClientInterface):
             {"op": str}  # operation. "copy", "link" or "download"
             {"errors": int}  # number of errors that occurred when downloading.
         """
+        all_processed = len(sts)
         with self._statistics.modify_storage() as staging_storage:
             # failed to acquire lock for any reason
             if staging_storage is None:
                 return
 
             already_processed = staging_storage.get("files_processed", 0)
-            if already_processed >= len(sts):
+            if already_processed >= staging_storage.get("total_files",0):
                 return
 
-            staging_storage["files_processed"] += len(sts) - already_processed
-            for st in sts[already_processed:]:
+            staging_storage["files_processed"] += all_processed - already_processed
+            for st in sts[already_processed:all_processed]:
                 _suffix = st.get("op")
                 if _suffix in {"copy", "link", "download"}:
                     staging_storage[f"files_processed_{_suffix}"] += 1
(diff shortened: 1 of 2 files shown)


app/ota_client.py
```diff
@@ -206,25 +205,64 @@ class OtaClientUpdatePhase(Enum):
...
+                lock_acquired = True
+                yield staging_slot
+            else:
+                # failed to acquire lock, so we don't yield a staging_slot
+                yield None
+        finally:
+            if lock_acquired:
+                self._slot = staging_slot.copy()
+                staging_slot.clear()
+                self._lock.release()
```
obi-t4, line 265: self._lock.acquire() always returns True, since timeout is not set. So if condition can be removed.
```suggestion
        try:
            self._lock.acquire()
            staging_slot: dict = self._slot.copy()
            yield staging_slot
        finally:
            self._slot = staging_slot.copy()
            staging_slot.clear()
            self._lock.release()
```

Bodong-Yang: updated in commit e3be438.

app/ota_client.py
```diff
@@ -206,25 +205,58 @@ class OtaClientUpdatePhase(Enum):
...
 
-        self.file_size_processed_copy = 0
-        self.file_size_processed_link = 0
-        self.file_size_processed_download = 0
+    def set(self, attr: str, value):
+        """
+        set a single attr in the slot
+        """
+        with self._lock:
+            self._slot[attr] = value
```
obi-t4, line 239: I didn't know until recently, but this operation doesn't require a lock!
https://docs.python.org/3/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe

app/ota_client.py
```diff
@@ -526,48 +546,38 @@ def _create_symbolic_links(self, list_file, standby_path):
...
+            {"errors": int}  # number of errors that occurred when downloading.
+        """
+        all_processed = len(sts)
+        with self._statistics.acquire_staging_storage() as staging_storage:
+            already_processed = staging_storage.get("files_processed", 0)
+            if already_processed >= staging_storage.get("total_files", 0):
+                return
+
+            staging_storage["files_processed"] += all_processed - already_processed
+            for st in sts[already_processed:all_processed]:
```
obi-t4, line 566: reducing iteration counter, I see.

app/ota_client.py
```diff
@@ -526,48 +546,38 @@ def _create_symbolic_links(self, list_file, standby_path):
...
+            {"op": str}  # operation. "copy", "link" or "download"
+            {"errors": int}  # number of errors that occurred when downloading.
+        """
+        all_processed = len(sts)
+        with self._statistics.acquire_staging_storage() as staging_storage:
+            already_processed = staging_storage.get("files_processed", 0)
+            if already_processed >= staging_storage.get("total_files", 0):
+                return
+
+            staging_storage["files_processed"] += all_processed - already_processed
```
obi-t4, line 565: [L561](https://github.com/tier4/ota-client/pull/87/files?w=1#diff-4e43d11537bf0064fc07412772bff9e85eae2bc9334ee70950cd7217088600f0R561) is using get, but here, get is not used. I think `[]` access is OK since staging_storage always has `files_processed` key.

app/ota_client.py
```diff
@@ -526,48 +546,38 @@ def _create_symbolic_links(self, list_file, standby_path):
...
+                _suffix = st.get("op")
+                if _suffix in {"copy", "link", "download"}:
+                    staging_storage[f"files_processed_{_suffix}"] += 1
+                    staging_storage[f"file_size_processed_{_suffix}"] += st.get(
+                        "size", 0
+                    )
+                    staging_storage[f"elapsed_time_{_suffix}"] += int(
+                        st.get("elapsed", 0) * 1000
+                    )
+                    staging_storage[f"errors_{_suffix}"] = st.get("errors", 0)
```
obi-t4, line 576: I'd like to limit as follows and use increment(+=):
```suggestion
                    if _suffix == "download":
                        staging_storage[f"errors_{_suffix}"] += st.get("errors", 0)
```

obi-t4 (commented):
I left some comments. Please have a check!

app/ota_client.py
Bodong-Yang: OK!

app/ota_client.py
```diff
@@ -573,7 +574,8 @@ def _set_statistics(self, sts):
                     staging_storage[f"elapsed_time_{_suffix}"] += int(
                         st.get("elapsed", 0) * 1000
                     )
-                    staging_storage[f"errors_{_suffix}"] = st.get("errors", 0)
+                    if _suffix == "download":
+                        staging_storage[f"errors_{_suffix}"] = st.get("errors", 0)
```
obi-t4, line 578: increment operator should be used, right?

Bodong-Yang: Oh sorry I miss that!

Bodong-Yang: Oh sorry I miss that!

obi-t4:
@Bodong-Yang RP seems to be fine, I'll try it out!

obi-t4 (approved):
LGTM :+1:

PR #88, opened 2021-12-09
[T4PUB-666] wait reboot sub to support multiple ecu

obi-t4:
# What is this PR?
This PR supports multiple ECU update/status requests.

# issue
~~https://tier4.atlassian.net/browse/T4PB-13539~~
https://tier4.atlassian.net/browse/T4PUB-666

# test
I confirm that update and status request for the two VirtualBox machines works properly.

### DIFF
--- a/app/ota_client_stub.py
+++ b/app/ota_client_stub.py
@@ -2,6 +2,7 @@ import asyncio
 from concurrent.futures import ThreadPoolExecutor
 from threading import Event
 
+import grpc
 import otaclient_v2_pb2 as v2
 from ota_client import OtaClient
 from ota_client_call import OtaClientCall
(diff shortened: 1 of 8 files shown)


app/ota_client_stub.py
```diff
@@ -29,21 +31,50 @@ def __init__(self):
...
                         self._ota_client_call.update(request, secondary["ip_addr"]),
-                        name=secondary,  # register the task name with sub_ecu id
+                        # register the task name with sub_ecu id
+                        name=secondary["ecu_id"],
                     )
                 )
+            else:
+                logger.warning(f"{secondary['ecu_id']} was not found in the {request=}")
+
+        def can_reboot():
```
obi-t4, line 58: If all secondary ecus are SUCCESS, this function returns true.

Bodong-Yang:
Please let me confirm the scenario:
1. mainECU dispatches ota-update request to all subECU, and subECU dispatches ota-update request to subsub...ECU, recursively, in a top-down fashion;
2. Only when all the child ECUs are updated, the parent ECU can update itself, recursively in bottom-up fashion.

Is that right?

Bodong-Yang:
Actually I think the implementation of the logic should be restricted in the grpc-server, not stepped into the ota-client(I can see that a closure function **can_reboot** is passed deep into the ota-client).

More specifically speaking, I think the status pulling should be implemented in the grpc-server side(in ota-client-service), the scenarios may as follow:
(in ota_client_stub)
1. mainECU dispatches the ota-update requests to all subECU,
2. wait for the immediately responses from all subECUs non-blocking, confirm that all subECUs receive the request,
3. mainECU responses to the IoT agent non-blocking,
4. mainECU lauches a background tasks to pulling all subECUs and wait for updating itself.

obi-t4:
> mainECU dispatches ota-update request to all subECU, and subECU dispatches ota-update request to subsub...ECU, recursively, in a top-down fashion;
Only when all the child ECUs are updated, the parent ECU can update itself, recursively in bottom-up fashion.

Right.
the parent ECU start updating itself, and wait reboot until all the child has been finished.

obi-t4:
It is better to start updating for main ecu if update is requested, but the reboot should be waited until all children are finished.

obi-t4:
I'd like to close this PR since #97 was merged.

PR #90, opened 2021-12-17
[T4PUB-663] support ROSCube-X nvme storage

obi-t4:
# What is this PR?
This PR supports ROSCubeX nvme storage.

# issue
https://tier4.atlassian.net/browse/T4PUB-663

# test
OTA update works properly.

### DIFF
--- a/app/extlinux_control.py
+++ b/app/extlinux_control.py
@@ -134,6 +134,8 @@ class HelperFuncs:
 class Nvbootctrl:
     """
     slot num: 0->A, 1->B
+    slot suffix: "", "_b"
+    rootfs default label prefix: APP
     """
 
     CURRENT_STANDBY_FLIP = {"0": "1", "1": "0"}
--- a/app/extlinux_control.py
+++ b/app/extlinux_control.py
@@ -238,13 +238,19 @@ class Nvbootctrl:
         return dev
 
     @classmethod
+    def get_standby_slot_dev_by_partlabel(cls) -> str:
+        slot = cls.get_standby_slot()
+        suffix = cls.get_suffix(slot)
+        dev = HelperFuncs.get_dev_by_partlabel(f"{cls.PREFIX}{suffix}")
+        logger.debug(f"standby slot dev by part label {cls.PREFIX}{suffix}: {dev}")
+        return dev
+
+    @classmethod
     def get_standby_slot_partuuid(cls) -> str:
         """
         NOTE: This partuuid is the UUID of the partition labled APP or APP_b.
         """
-        slot = cls.get_standby_slot()
-        suffix = cls.get_suffix(slot)
-        dev = HelperFuncs.get_dev_by_partlabel(f"{cls.PREFIX}{suffix}")
+        dev = cls.get_standby_slot_dev_by_partlabel()
         return HelperFuncs.get_partuuid_by_dev(dev)
 
 
@@ -395,6 +401,7 @@ class CBootControl:
             self._standby_slot,
             self._standby_dev,
             self._standby_partuuid,
+            self._standby_dev_by_partlabel,
         ) = self._get_slot_info()
 
         # NOTE: only support r580 platform right now!
@@ -421,14 +428,18 @@ class CBootControl:
         _slot = Nvbootctrl.get_standby_slot()
         _dev = Nvbootctrl.get_standby_slot_dev()
         _partuuid = Nvbootctrl.get_standby_slot_partuuid()
+        _dev_by_partlabel = Nvbootctrl.get_standby_slot_dev_by_partlabel()
 
-        return _slot, _dev, _partuuid
+        return _slot, _dev, _partuuid, _dev_by_partlabel
 
     def get_standby_slot(self) -> str:
         return self._standby_slot
 
     def get_standby_dev(self) -> Path:
         return Path(self._standby_dev)
+
+    def get_standby_dev_by_partlabel(self) -> Path:
+        return Path(self._standby_dev_by_partlabel)
 
     def get_standby_partuuid(self) -> str:
         return self._standby_partuuid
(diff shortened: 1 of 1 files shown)


app/extlinux_control.py
```diff
@@ -195,12 +238,12 @@ def get_standby_slot_dev(cls) -> str:
 
     @classmethod
     def get_standby_slot_partuuid(cls) -> str:
```
obi-t4, line 240: This function returns APP or APP_b partition UUID instead of `get_standby_slot_dev`()'s partition.
APP or APP_b partition UUID should be used for extlinux.conf `PARTUUID=` specifier.

app/extlinux_control.py
```diff
@@ -195,12 +238,12 @@ def get_standby_slot_dev(cls) -> str:
...
-        dev = cls.get_standby_slot_dev()
-        return HelperFuncs.get_partuuid_by_dev(dev)
-
-    @classmethod
-    def get_current_slot_partuuid(cls) -> str:
```
obi-t4, line 202: removed since not referenced.

app/extlinux_control.py
```diff
@@ -88,6 +98,42 @@ def get_dev_by_partlabel(cls, partlabel: str) -> str:
...
+        return cls._lsblk(cmd)
+
+    @classmethod
+    def get_family_devs(cls, parent_device: str) -> list:
+        """
+        When `/dev/nvme0n1` is specified as parent_device,
+        ["NAME=/dev/nvme0n1", "NAME=/dev/nvme0n1p1", "NAME=/dev/nvme0n1p2"] is returned.
+        """
+        cmd = f"-Pp -o NAME {parent_device}"
+        return cls._lsblk(cmd).split("\n")
```
Bodong-Yang, line 120:
```suggestion
        return cls._lsblk(cmd).splitlines()
```

app/extlinux_control.py
```diff
@@ -88,6 +98,42 @@ def get_dev_by_partlabel(cls, partlabel: str) -> str:
...
+        """
+        parent = cls.get_parent_dev(device)
+        family = cls.get_family_devs(parent)
+        pa = re.compile(r'NAME="(?P<dev>.*)"')
+        for blk in family:
+            m = pa.match(blk)
+            dev = m.group("dev")
+            if dev != parent and dev != device:
+                return dev
+        raise OtaErrorUnrecoverable(f"lsblk output={family} is illegal")
```
Bodong-Yang, line 135:
```suggestion
    @classmethod
    def get_sibling_dev(cls, device: str) -> str:
        """
        When `/dev/nvme0n1p1` is specified as child_device, /dev/nvme0n1p2 is returned.
        """
        family = cls.get_family_devs(parent)
        partitions = {i.split("=")[-1].strip('"') for i in family[1:]}
        res = partitions - {device}
        if len(res) == 1:
            return *res
        else:
            raise OtaErrorUnrecoverable(f"device is has unexpected partition layout, {family=}")
```

Bodong-Yang, line 135: No need to use regex here.
Also handling unexpected partition layout here, i.e., the device doesn't have exact 2 partitions.

Bodong-Yang:
By reading the nvme-installer scripts in the ota-internal repo, this nvme installing mechanism using switch-root to achieve rootOnStorage.
With this mechanism, it seems that the system will always use kernel from the emmc partition's /boot folder, should we also copy the new kernel from the newly update nvme slot into the corresponding emmc partition's /boot folder to update the kernel?

Bodong-Yang:
By the way, should we keep the backwards compatibility with ECU that doesn't have nvme external root?

obi-t4:
> By the way, should we keep the backwards compatibility with ECU that doesn't have nvme external root?

Nice to have, not required.

Bodong-Yang:
> Nice to have, not required.

Maybe in the future a PR can be made for this.

obi-t4:
> 



> P.S.: I think only updating the `initrd` and `kernel` under the corresponding emmc partition's /boot folder is enough.
> P.S.2: Or we can kexec with kernel from newly updated nvme slot.

I'd like to choose option.1 since we don't want to troubleshoot about kexec. kexec is a popular mechanism?

obi-t4:
Arbitrary extlinux.conf PARTUUID under nvme drive works, since mmc/boot is referenced :smile: 
```
autoware@ubuntu:~$ cat /boot/extlinux/extlinux.conf 
# Auto generated by ota-client, DO NOT EDIT! 
TIMEOUT 30 
DEFAULT primary 
MENU TITLE L4T boot options 
LABEL primary 
	MENU LABEL primary kernel 
	LINUX /boot/Image 
	INITRD /boot/initrd 
	FDT /boot/tegra194-rqx-580.dtb 
	APPEND ${cbootargs} quiet  root=PARTUUID=11111111-1111-1111-1111-111111111111 rw rootwait rootfstype=ext4 console=ttyTCU0,115200n8 console=tty0 fbcon=map:0 net.ifnames=0 
```

Bodong-Yang:
> > 
> 
> > P.S.: I think only updating the `initrd` and `kernel` under the corresponding emmc partition's /boot folder is enough.
> > P.S.2: Or we can kexec with kernel from newly updated nvme slot.
> 
> I'd like to choose option.1 since we don't want to troubleshoot about kexec. kexec is a popular mechanism?

Kexec is for directly kernel booting to new kernel from current kernel, in normal case we don't use it.
But option 1 has a potential problem that it will cause the mismatching of system on emmc and `kernel` from newly updated slot. It maybe OK if both emmc's system and nvme's system have the same major Ubuntu version though, but It is certainly a potential problem.

obi-t4:
For option 1, Is it OK if kernel and initrd in the nvme /boot and corresponding emmc /boot are updated?

Bodong-Yang:
> Arbitrary extlinux.conf PARTUUID under nvme drive works, since mmc/boot is referenced smile
> 
> ```
> autoware@ubuntu:~$ cat /boot/extlinux/extlinux.conf 
> # Auto generated by ota-client, DO NOT EDIT! 
> TIMEOUT 30 
> DEFAULT primary 
> MENU TITLE L4T boot options 
> LABEL primary 
> 	MENU LABEL primary kernel 
> 	LINUX /boot/Image 
> 	INITRD /boot/initrd 
> 	FDT /boot/tegra194-rqx-580.dtb 
> 	APPEND ${cbootargs} quiet  root=PARTUUID=11111111-1111-1111-1111-111111111111 rw rootwait rootfstype=ext4 console=ttyTCU0,115200n8 console=tty0 fbcon=map:0 net.ifnames=0 
> ```

Yes, this mechanism firstly boot into emmc slot, and then use `systemctl switch-boot` to use external drive as rootfs.
So a working system on emmc slot must presented, and cboot will read `extlinux.cfg` and load kernel from emmc slot.

Bodong-Yang:
Oh I just realize we are just updating the extlinux.cfg on the nvme slot, which will not be used by cboot.
Actually with this mechanism, we even don't need to configure the extlinux.cfg, as ota-client won't touch the emmc slots.

Bodong-Yang:
> For option 1, Is it OK if kernel and initrd in the nvme /boot and corresponding emmc /boot are updated?

I don't know, **MAYBE** within the same major OS version it will work.

But if one day the nvme slot updated into a newer major OS version, we may have unexpected problems which won't have any support since combining using mismatching kernel and system is unusual. Worse case we may have a unbootable emmc slot.

obi-t4:
> But if one day the nvme slot updated into a newer major OS version

No. We always update both kernel and initrd on nvme and emmc.

Bodong-Yang:
> > But if one day the nvme slot updated into a newer major OS version
> 
> No. We always update both kernel and initrd on nvme and emmc.

I mean the system on the emmc mismatching with the kernel we update.

Bodong-Yang:
With the above concerns I have, I think kexec is a better and more controllable way to make the system use kernel in the nvme slot, with this method, we don't need to touch emmc slot anymore.

Bodong-Yang:
But we still need some test over it if we want to use it.

PR #92, opened 2021-12-21
[T4PB-12360] support cancel

obi-t4:
# What is this PR?
This PR supports canceling update.

# issue
https://tier4.atlassian.net/browse/T4PB-12360

# test
1. launch HTTP server locally on the host side
2. launch ota-client app/main.py on the target side
3. request update by app/ota_client_service.py. Note IP address should be edit to specify host HTTP server.
4. Ctrl+C to cancel update
5. terminal shows as follows:
```
response=ecu {
  ecu_id: "autoware"
  status {
    status: FAILURE
    failure: RECOVERABLE
    failure_reason: "canceled"
    version: "autoware.1.1"
    progress {
      phase: REGULAR
      total_regular_files: 233942
      regular_files_processed: 41441
      files_processed_download: 41441
      file_size_processed_download: 357209843
      elapsed_time_copy {
      }
      elapsed_time_link {
      }
      elapsed_time_download {
        seconds: 96
        nanos: 533000000
      }
    }
  }
}
```

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -20,7 +20,12 @@ from retrying import retry
 from ota_client_interface import OtaClientInterface
 from ota_metadata import OtaMetadata
 from ota_status import OtaStatus, OtaStatusControlMixin
-from ota_error import OtaErrorUnrecoverable, OtaErrorRecoverable, OtaErrorBusy
+from ota_error import (
+    OtaErrorUnrecoverable,
+    OtaErrorRecoverable,
+    OtaErrorBusy,
+    OtaErrorCancel,
+)
 from copy_tree import CopyTree
 from configs import config as cfg
 import log_util
@@ -291,7 +296,7 @@ class _BaseOtaClient(OtaStatusControlMixin, OtaClientInterface):
             # not setting ota_status
             logger.exception("update busy")
             return OtaClientFailureType.RECOVERABLE
-        except (JSONDecodeError, OtaErrorRecoverable) as e:
+        except (JSONDecodeError, OtaErrorRecoverable, OtaErrorCancel) as e:
             logger.exception(msg="recoverable")
             self.set_ota_status(OtaStatus.FAILURE)
             self.store_standby_ota_status(OtaStatus.FAILURE)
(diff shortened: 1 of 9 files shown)


Bodong-Yang:
I think we should not interrupt an on-going update, if the ota update already starts and changes are made.
The things we can do is to stop the ota update request dispatching.

To achieve the cancelling of pending request sending tasks, **Task.cancel()** is enough I think.
As for the local update, if the ota update doesn't start yet(pending or waiting for executing), we can safely cancel it with **Future.cancel()**.

If the Task or Future can not be cancelled,
```
# According to the oncurrent.futures.Future document
cancel()
Attempt to cancel the call. 
If the call is currently being executed or finished running and cannot 
be cancelled then the method will return False, otherwise the call will 
be cancelled and the method will return True.
```
The cancel API call should return failure.

obi-t4:
> I think we should not interrupt an on-going update, if the ota update has already started and changes are made.

What we'd like to do is cancel update which has already started.
We'd like to support update cancel operation by end-user. For example, if a user starts an update but finds the ETA is very large, he/she wants to cancel updating.

Bodong-Yang:
> > I think we should not interrupt an on-going update, if the ota update has already started and changes are made.
> 
> What we'd like to do is cancel update which has already started. We'd like to support update cancel operation by end-user. For example, if a user starts an update but finds the ETA is very large, he/she wants to cancel updating.

I think we should not interrupt an already started OTA update, interrupting an on-going OTA update equals to an unexpected reboot, and it will result in an corrupted standby slot.

Common OTA solution, like the one in Android or Windows, don't allow user to interrupt or stop OTA update when the update is being applied. The user can only stop the update when the OTA update is still in the preparing stage.
In Android, if the OTA starts to be applied, the user cannot operate the device or use normal way to stop the OTA update.
In Windows, the windows will also prompt a full windows to tell the user to wait for OTA to complete and do not turn off the power.

If user really need to interrupt the on-going update(the update is hanging, etc.), it is not our responsibility(not the ota-client) to stop it, the user should take the risk of their own, and force reboot/reset the system(hard reset or soft reset) by their own.

Also, force kill the thread through low level python API is too risky and introduce unnecessary complexity to the current code base. 
Anyway, force kill the ota-update thread actually equals to an unexpected reboot, if we REALLY want to stop the on-going update, a simple subprocess call to `reboot` is enough.

obi-t4:
>  if we REALLY want to stop the on-going update, a simple subprocess call to reboot is enough.

I see. thanks.

PR #94, opened 2022-01-07
add condition where filesize is None

obi-t4:
# What is this PR?
This PR fixes the issue that ota-client downloads all files from the server w/o copying from active bank if file size attribute is not set.


Youneko (approved):
LGTM!

PR #95, opened 2022-01-11
WIP: Support for multi-subECU OTA update

Bodong-Yang:
Introduce multi-subECU OTA update support for ota-client.
When doing OTA update, the ota-client will wait for all its subECUs finishing their update before finishing up its own OTA update.
The ota-client will keep pulling the status of all subECUs, until it confirms that all the subECUs finish their update.

**WIP: not yet being fully tested!**

PR #97, opened 2022-01-11
[T4PUB-666] multi ota

obi-t4:
# About this PR
modification of https://github.com/tier4/ota-client/pull/95

# issue
https://tier4.atlassian.net/browse/T4PUB-666

# test
multiple ECU update works properly.

# note
ota_client_stub seems complicated, so refactoring is required.

# known bug
```
exception calling callback for <Future at 0x7fc1f4096ee0 state=finished returned NoneType>
Traceback (most recent call last):
  File "/usr/lib/python3.8/concurrent/futures/_base.py", line 328, in _invoke_callbacks
    callback(self)
  File "/usr/lib/python3.8/asyncio/futures.py", line 374, in _call_set_state
    dest_loop.call_soon_threadsafe(_set_state, destination, source)
  File "/usr/lib/python3.8/asyncio/base_events.py", line 764, in call_soon_threadsafe
    self._check_closed()
  File "/usr/lib/python3.8/asyncio/base_events.py", line 508, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed
Exception ignored in: <function OtaClientStub.__del__ at 0x7fc1f6385940>
Traceback (most recent call last):
  File "/ota-client/app/ota_client_stub.py", line 61, in __del__
    self._executor.shutdown()
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 236, in shutdown
    t.join()
  File "/usr/lib/python3.8/threading.py", line 1008, in join
    raise RuntimeError("cannot join current thread")
RuntimeError: cannot join current thread
```


app/extlinux_control.py
```diff
@@ -716,12 +716,16 @@ def get_version(self) -> str:
 
     def boot_ctrl_pre_update(self, version):
         logger.debug("entering pre-update...")
+        self.store_slot_in_use_file(
```
obi-t4, line 719: @Bodong-Yang is this required? fixing bug?

app/ota_client_stub.py
```diff
@@ -134,3 +206,246 @@ def _find_request(update_request, ecu_id):
...
+                self._ensure_subecu_status(
+                    timeout=server_cfg.WAITING_SUBECU_READY_TIMEOUT
+                )
+            )
+            # all subECUs are updated, now the ota_client can reboot
+            logger.debug("all subECUs are ready, set post_update_event")
+            post_update_event.set()
+
+            logger.debug("wait for local ota update to finish...")
+            exp = _future.exception(timeout=server_cfg.LOCAL_OTA_UPDATE_TIMEOUT)
```
obi-t4, line 244: @Bodong-Yang I'll remove this timeout since the current implementation doesn't have a timeout.

app/extlinux_control.py
Bodong-Yang: I think this is a redundant line, I think we can remove it.
Because I notice there is a same line below, write the current slot's `slot_in_use` file.

app/ota_client_stub.py
Bodong-Yang: OK!

app/ota_client_stub.py
```diff
@@ -134,3 +215,270 @@ def _find_request(update_request, ecu_id):
...
+                if not keep_pulling:
+                    return
+            else:
+                # unreachable directly connected subECUs presented
+                logger.info(
+                    f"retry pulling subECUs status due to unreachable subECUs {failed_ecu_list}"
+                )
+
+                """
+                if retry_count > retry:
```
obi-t4, line 459: retry counter is removed since we can't define how long it will take for subecu to reboot for now.

Bodong-Yang (approved):
For the code part, it seems that there is no big issue.
If the test with real machine also passed, I think it is OK to merge the PR!

PR #104, opened 2022-02-08
[T4PB-15011] replace/append partuuid

obi-t4:
# What is this PR?
This PR modifies the extlinux.conf operation:
- update only `PARTUUID={UUID}` part.
- when `PARTUUID={UUID}` part doesn't exist, it is just appended.

# issue
https://tier4.atlassian.net/browse/T4PB-15011

# test
rqx-58g and x86 platform have been confirmed.


app/ota_client.py
```diff
@@ -724,7 +724,7 @@ def _create_regular_file(
         )
 
         if str(reginf.path).startswith("/boot"):
-            dst = boot_standby_path / reginf.path.name
+            dst = boot_standby_path / reginf.path.relative_to("/boot")
```
obi-t4, line 727: to support directory under /boot.

app/extlinux_control.py
```diff
@@ -474,45 +335,28 @@ def is_current_slot_marked_successful(self) -> bool:
...
+                        f"{ma_append.group(1)}APPEND{ma_rest.group(1)}root={self.standby_slot_partuuid}{ma_rest.group(3)}\n"
+                    )
+                else:
+                    # If root=PARTUUID= entry doesn't exist, it is just appended.
+                    out.append(f"{line} root={self.standby_slot_partuuid}\n")
+            else:
+                out.append(f"{line}\n")
+        with open(dst, "w") as f:
+            f.writelines(out)
+            os.fsync(f.fileno())
```
Bodong-Yang, line 359: I think the following is enough:
```python
    def update_extlinux_cfg(self, dst: Path, src: Path):
       
        def _replace(ma: re.Match, repl: str):
            append_l = ma.group(0)
            res, n = re.compile(r'root=[\w\-=]*').subn(repl, append_l)
            if not n:
                res = f"{append_l} {repl}"

            return res

        _repl_func = partial(_replace, repl=self.standby_slot_partuuid)
        dst.write_text(
            re.compile(r'.*APPEND.*').sub(_repl_func, src.read_text())
        )
```

app/extlinux_control.py
```diff
@@ -474,45 +336,19 @@ def is_current_slot_marked_successful(self) -> bool:
...
-        if src and src.is_file():
-            # load template extlinux_cfg from external
-            cfg_text = self.gen_extlinux_cfg(src)
-        else:
-            # generate extlinux from default settings
-            cfg_text = self.gen_extlinux_cfg()
-
-        with open(target, "w") as f:
-            f.write(cfg_text)
+        _repl_func = partial(_replace, repl=self.standby_slot_partuuid)
```
Bodong-Yang, line 350:
```suggestion
        _repl_func = partial(_replace, repl=f"root=PARTUUID={self.standby_slot_partuuid}")
```
Sorry I forgot to prepare the pure PARTUUID into "root=PARTUUID=<PARTUUID>" in my previous comment.

app/extlinux_control.py
```diff
@@ -474,45 +336,19 @@ def is_current_slot_marked_successful(self) -> bool:
...
+        dst.write_text(re.compile(r".*APPEND.*").sub(_repl_func, src.read_text()))
```
Bodong-Yang, line 351:
```suggestion
        dst.write_text(re.compile(r'\n\s*APPEND.*').sub(_repl_func, src.read_text()))
```
I think of the regex again and I think this regex should runs faster than the previous one.

app/extlinux_control.py
Bodong-Yang: Yes I confirm that!

app/extlinux_control.py
```diff
@@ -348,7 +348,7 @@ def _replace(ma: re.Match, repl: str):
             return res
 
         _repl_func = partial(_replace, repl=f"root={self.standby_slot_partuuid}")
-        dst.write_text(re.compile(r"\n\sAPPEND.*").sub(_repl_func, src.read_text()))
```
Bodong-Yang: Emmm you just forgot to add * after \s, no need to change back to use `.*APPEND.*`.

obi-t4:
@Bodong-Yang OTA on the actual board has been confirmed.

Bodong-Yang (approved):

PR #113, opened 2022-03-03
ota_proxy enhancement

Bodong-Yang:
# Introduction
This PR introduces some enhancement and changes over ota_proxy module.
1. ota_cache will only be cleared when all the subecus are ready, and local update is applied,
2. refactor over `ota_proxy.db`,
3. fix, refine and improvements over `ota_proxy`,
4. introduce new mechanisms to sychronizing ota updating state between ota_client and ota_service_stub,
5. (BUG FIX) potential timeout problem caused by blocking queue(on file caching) and connection leaking(broken generator)
6. (BUG FIX) timeout problem caused by `aiohttp` default timeout setting(5mins)
7. ota_client: just in case, add outer retry on `Downloader`; limit the process pool size on `create_regular_files`
8. update test files accordingly

# behavior change
1. proxy_info: default settings are as follow
```yaml
# if no proxy_info.yaml presented,
# we should treat the ecu as main ecu(as gateway),
# and enable ota_proxy without upper proxy
enable_ota_proxy: true
gateway: true
```
2. in subecu, replace scheme to HTTP in the URL unconditionaly
3. (NEW) ota_client: if proxy is enabled locally(either via local ota_proxy or upper_ota_proxy), always replace scheme to HTTP(via Downloader)
4. (NEW) no timeout is applied to the local ota update (`configs.LOCAL_OTA_UPDATE_TIMEOUT` removed)
5. (NEW) introduce custom header `Ota-File-Cache-Control`, allowing ota_client to control the caching behavior of ota_proxy

# TODO
- [x] type annotation and documentation
- [x] provide a set of ota_cache_control headers
- [x] address the potential timeout problem
- [x] new state synchronizing mechanism between ota_client and ota_stub
- [x] for proxy_info.yaml, set `gateway: true` for backward compatibility
- [x] check proxy_info.yaml, if `gateway: true`, unconditionally replace https to http


app/ota_client_stub.py
```diff
@@ -95,14 +97,21 @@ def start(self, enable_cache=False):
...
+        from ota_proxy.config import config as proxy_srv_cfg
+
         with self._lock:
             if not self._closed:
                 # send SIGTERM to the server process
                 self._server_p.terminate()
                 self._closed = True
                 logger.info("ota proxy server closed")
 
+                if cleanup_cache:
```
obi-t4, line 110: I think cleanup is not required since this is called just before switching partition.
How do you think?

ota_proxy/db.py
```diff
@@ -1,43 +1,57 @@
...
+def make_cachemeta_cls(name: str):
+    # set default value of each field as field type's zero value
+    return make_dataclass(
+        name,
+        [(k, v.col_type, v.col_type()) for k, v in cfg.COLUMNS.items()],
+        bases=(_CacheMetaMixin,),
+    )
+
+
+CacheMeta = make_cachemeta_cls("CacheMeta")
```
obi-t4, line 43: I can understand that you don't want to treat COLUMNS members in CacheMeta and OTACacheDB.
But this change reduces code readability slightly, I think.
Easy to understand is one of the important things.
I'd like to hear your opinion.

obi-t4 (commented):
I left some comments.
(The review is not finished yet)

app/ota_client_stub.py
Bodong-Yang: Do you mean I should cleanup the cache after first successful reboot?

obi-t4: after successful reboot, ota cache directory is disappeared since the partition has been switched, isn't it?

ota_proxy/db.py
Bodong-Yang: Hard-coding the table column definition may increase the maintenance complexity.
If I the definition of table change(even the order of the column changed), not only the database operation related code needed to be adjusted, the `CacheMeta` class, and even the code that use `CacheMeta` may need to be adjusted.

Considering the above situation, I decided to dynamically create `CacheMeta` class directly from the db table definiton. Actually it is a simple version of ORM, since our requirement is not so complicated, I directly write one based on dataclass instead of relying on third-party module.

ota_proxy/ota_cache.py
```diff
@@ -35,9 +36,25 @@ def _subprocess_check_output(cmd: str, *, raise_exception=False) -> str:
...
+            return False
+        else:
+            self[url] = None
+            return True
+
+    def unregister(self, url: str):
+        try:
+            del self[url]
+        except KeyError:
+            pass
```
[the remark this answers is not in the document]
obi-t4: I have question, this member is accessed from multiple threads? (if so lock is required)

obi-t4 (commented):
I left some comments.

ota_proxy/db.py
```diff
@@ -55,91 +88,54 @@ def close(self):
...
-            raise sqlite3.OperationalError("connect is closed")
-
-        url_list = [(u,) for u in urls]
-        with self._wlock:
-            cur = self._con.cursor()
-            cur.executemany(f"DELETE FROM {self.TABLE_NAME} WHERE url=?", url_list)
-
+        urls = [(u,) for u in urls]
+        with self._wlock, self._general_query(
+            f"DELETE FROM {self.TABLE_NAME} WHERE url=?", urls
```
obi-t4, line 120: Question: Can we make sure that urls don't contain `;` string? I'd like to make sure that query sanitizing is required or not.

app/ota_client_stub.py
Bodong-Yang: > after successful reboot, ota cache directory is disappeared since the partition has been switched, isn't it?

Yes, so I think the best time to clear cache is the POST_UPDATING stage, before reboot. In this stage, all subECUs are ready and local ota update also finished.

ota_proxy/ota_cache.py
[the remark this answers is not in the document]
Bodong-Yang: with GIL, operating set itself is thread safe. However, since `check url exists` and `add url in the set` in `register` method are not a single atomic operation, it may be edge condition happens, but it will not impact the file caching, so decided to not use lock.

ota_proxy/ota_cache.py
```diff
@@ -215,27 +351,40 @@ def __init__(
...
 
         self._storage_below_hard_limit_event = Event()
         self._storage_below_soft_limit_event = Event()
-        self._on_going_caching = dict()
+        self._on_going_caching = _RegisterDict()
         self._upper_proxy: str = ""
 
+        self._base_dir.mkdir(exist_ok=True, parents=True)
+        _event = Event()
+        _event.set()
```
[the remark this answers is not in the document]
Bodong-Yang: Oh yes you are right, it should be not set by default.

ota_proxy/db.py
Bodong-Yang: using `?` mark with input value is the standard way to assemble query string according to the python3 sqlite3 document. With `?` mark, I think the sqlite3 module will process the input value properly.
``` python3
# copied from python3 sqlite3 document

# Never do this -- insecure!
symbol = 'RHAT'
cur.execute("SELECT * FROM stocks WHERE symbol = '%s'" % symbol)

# Do this instead
t = ('RHAT',)
cur.execute('SELECT * FROM stocks WHERE symbol=?', t)
print(cur.fetchone())

# Larger example that inserts many records at a time
purchases = [('2006-03-28', 'BUY', 'IBM', 1000, 45.00),
             ('2006-04-05', 'BUY', 'MSFT', 1000, 72.00),
             ('2006-04-06', 'SELL', 'IBM', 500, 53.00),
            ]
cur.executemany('INSERT INTO stocks VALUES (?,?,?,?,?)', purchases)
```

ota_proxy/ota_cache.py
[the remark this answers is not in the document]
obi-t4: Entering here from multiple threads OK?
https://github.com/tier4/ota-client/pull/113/files#diff-bf7539f67c1e8add5ca7cafafc3886db722d3e80105caa19c4d9ce99584e4e1cR615-R620

```
                    # failed to get the chance to cache
                    fp, meta = await self._open_fp_by_requests(
                        url, cookies, extra_headers
                    )
                    res = OTAFile(url=url, fp=fp, meta=meta)
```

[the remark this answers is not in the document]
Bodong-Yang: > Entering here from multiple threads OK? https://github.com/tier4/ota-client/pull/113/files#diff-bf7539f67c1e8add5ca7cafafc3886db722d3e80105caa19c4d9ce99584e4e1cR615-R620
> 
> ```
>                     # failed to get the chance to cache
>                     fp, meta = await self._open_fp_by_requests(
>                         url, cookies, extra_headers
>                     )
>                     res = OTAFile(url=url, fp=fp, meta=meta)
> ```

Yes, it will not cause problems.
Actually the edge condition is when 2 threads simultaneously successfully get the chance to cache file, that will not cause problem, but just result in a file being cached twice. New copy will override old copy.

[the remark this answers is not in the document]
obi-t4: If this cache operation `just result in a file being cached twice.`, is safe operation, I think it is OK,
I think at least a comment is necessary, but I think it is easier to understand by adding a lock than by reading that comment.
Why don't you add lock?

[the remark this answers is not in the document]
obi-t4: When I read above comment that `lock is not required since...`, I think I'll try to submit to add lock and remove that comment.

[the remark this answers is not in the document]
Bodong-Yang: > If this cache operation `just result in a file being cached twice.`, is safe operation, I think it is OK, I think at least a comment is necessary, but I think it is easier to understand by adding a lock than by reading that comment. Why don't you add lock?

I try the implementation with lock, the performance is suffered obviously. Since we don't get too much improvement from adding lock, I decided to not add lock. In short, the reason is performance penalty.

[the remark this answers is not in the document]
obi-t4: OK, Could you add those comments?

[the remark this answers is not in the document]
Bodong-Yang: Sure!

obi-t4:
@Bodong-Yang Normal update was OK! so I'll try to shutdown during the update.

Bodong-Yang:
Test environment:
```yaml
##### mainecu #####
# network: 
ip_addr: 10.0.0.2
# ecu_info.yaml
ecu_id: "autoware"
ip_addr: "10.0.0.2"
secondaries:
  -  ecu_id: "p1"
     ip_addr: "10.0.0.11"

# proxy_info.yaml(not presented is also OK)
enable_ota_proxy: true
local_server: ["10.0.0.2", 8082]

##### subecu #####
# network: 
ip_addr: 10.0.0.11
# ecu_info.yaml
ecu_id: "p1"
ip_addr: "10.0.0.11"

# proxy_info.yaml(must presented)
enable_ota_proxy: true
upper_ota_proxy: "http://10.0.0.2:8082"
local_server: ["10.0.0.11", 8082]

##### update requests #####
- ecu_id: "autoware"
  version: "789.x"
  url: "http://10.0.0.1:8080"
  cookies: {"test": "my-cookie-main"}
- ecu_id: "p1"
  version: "789.x"
  url: "http://10.0.0.1:8080"
  cookies: {"test": "my-cookie-p1"}
```

Network environment:
```
vboxnet0:
host: 10.0.0.1 # where ota file server located
mainecu: 10.0.0.2
subecu: 10.0.0.11
```

The way I setup the network environment is as follow:
```bash
# on subecu
# to make sure subecu cannot directly connect to the host and the internet
sudo ip r a unreachable 0.0.0.0/0
sudo ip r a unreachable 10.0.0.1
```

obi-t4:
- [ ] standalone ECU, w/o proxy_info.yaml, w/ HTTPS: ota-cache works well
- [ ] Main ECU and Secondary ECU, w/ proxy_info.yaml, w/ HTTPS: ota-cache works well

obi-t4:
@Bodong-Yang 
I couldn't find cache data under /ota-data with this scenario:
` standalone ECU, w/o proxy_info.yaml, w/ HTTPS: check whether ota-cache works well`
Did you confirm it? Or now testing?

Bodong-Yang:
> @Bodong-Yang I couldn't find cache data under /ota-data with this scenario: ` standalone ECU, w/o proxy_info.yaml, w/ HTTPS: check whether ota-cache works well` Did you confirm it? Or now testing?

standalone scenario is not tested yet, yesterday I just finished testing multi-ecu environment.

Bodong-Yang:
@obi-t4 The cause of timeout problem may be discovered, I am working on solving it. The problem might be the blocking queue used in ota_proxy to streaming data when caching data, resulting other on-going requests being hanging, causing timeout. This is confirmed in the small scale test, I will use async queue to replace blocking queue in ota_cache module to see if problem solved.

obi-t4:
> standalone scenario is not tested yet

OK, understood. Please confirm it!

Bodong-Yang:
Test environment on standalone test:
```yaml
##### mainecu #####
# network: 
ip_addr: <ip>
# ecu_info.yaml
ecu_id: "autoware"
ip_addr: <ip>

# proxy_info.yaml(not presented is also OK)
enable_ota_proxy: true
gateway: true
local_server: [<ip>, 8082]
```

update request:
```yaml
- ecu_id: "autoware"
  version: "789.x"
  url: "https://<firmware_URL>"
  cookies: <cookie>
```

Bodong-Yang:
Standalone test is passed on the above setup.
Also the following scenarios are tested:
1. Retries on network interruption
2. ota_client indicates ota_proxy to re-cache when corrupted files are detected.

Bodong-Yang:
Make some updates on how ota_proxy signal ota_service that ota_proxy is ready(cache scrub finished).
Test with above testing setups also passed with latest commits.

obi-t4:
standalone ecu is working under the environment:

```
# ecu_info.yaml
format_version: 1
ecu_id: "autoware"

# no proxy_info.yaml
```

obi-t4:
main ecu:
```
root@autoware-ecu:/opt/ota# cat /boot/ota/ecu_info.yaml 
format_version: 1
ecu_id: "autoware"
secondaries:
- ecu_id: "perception1"
  ip_addr: "192.168.20.31"
available_ecu_ids:
- "autoware"
- "perception1"
root@autoware-ecu:/opt/ota# cat /boot/ota/proxy_info.yaml
gateway: true
enable_ota_proxy: true
```

seconrary ecu:
```
root@ubuntu:/opt/ota/client# cat /boot/ota/ecu_info.yaml 
format_version: 1
ecu_id: "perception1"
ip_addr: "192.168.20.31"
root@ubuntu:/opt/ota/client# cat /boot/ota/proxy_info.yaml 
enable_ota_proxy: true
upper_ota_proxy: "http://192.168.20.11:8082"
```

Bodong-Yang:
> main ecu:
> 
> ```
> root@autoware-ecu:/opt/ota# cat /boot/ota/ecu_info.yaml 
> format_version: 1
> ecu_id: "autoware"
> secondaries:
> - ecu_id: "perception1"
>   ip_addr: "192.168.20.31"
> available_ecu_ids:
> - "autoware"
> - "perception1"
> root@autoware-ecu:/opt/ota# cat /boot/ota/proxy_info.yaml
> gateway: true
> enable_ota_proxy: true
> ```
> 
> seconrary ecu:
> 
> ```
> root@ubuntu:/opt/ota/client# cat /boot/ota/ecu_info.yaml 
> format_version: 1
> ecu_id: "perception1"
> ip_addr: "192.168.20.31"
> root@ubuntu:/opt/ota/client# cat /boot/ota/proxy_info.yaml 
> enable_ota_proxy: true
> upper_ota_proxy: "http://192.168.20.11:8082"
> ```

Yes I think this setup should work as expected!

app/ota_client.py
```diff
@@ -49,9 +49,70 @@ def verify_file(filename: Path, filehash: str, filesize) -> bool:
     return file_sha256(filename) == filehash
 
 
+_ExceptionWrapper = type("_ExceptionWrapper", (Exception,), dict())
```
obi-t4, line 52: In my opinion, The following is natural. Dynamic generation is not required, I think.
```python
class _ExceptionWrapper(Exception):
    pass
```

app/ota_client.py
```diff
@@ -105,50 +172,55 @@ def _path_to_url(base: str, p: Union[Path, str]) -> str:
...
     ) -> int:
         url = self._path_to_url(url_base, path)
+        if not headers:
+            headers = dict()
 
-        error_count = 0
         try:
-            response = self._session.get(url, stream=True, timeout=10, cookies=cookies)
+            error_count = 0
+            response = self._session.get(
```
obi-t4, line 200: Let me confirm, this session.get is retried by RETRY_COUNT times with retry_strategy and this __call__ function is also retried with `@partial(_retry, RETRY_COUNT, OUTER_BACKOFF_FACTOR, BACKOFF_MAX)` ?
I think it is ok since more robust.

app/ota_client.py
```diff
@@ -314,6 +386,116 @@ def acquire_staging_storage(self):
...
+    ######## state machine definition ########
+    # states definition
+    _START, _S0, _S1, _S2, _END = (
+        "_START",  # start
+        "_S0",  # cache_scrub_finished
+        "_S1",  # pre_update_finished
+        "_S2",  # apply_update_finished
+        "_END",  # end
+    )
+    _STATE_LIST = ["_START", "_S0", "_S1", "_S2", "_END"]
```
obi-t4, line 412: It is better not to use immediate value:
```suggestion
    _STATE_LIST = [_START, _S0, _S1, _S2, _END]
```

app/ota_client.py
```diff
@@ -314,6 +386,116 @@ def acquire_staging_storage(self):
...
+    # participators definition
+    _P1, _P2 = "ota_service", "ota_client"
+    # which participator can start the fsm
+    _STARTER = _P1
+
+    # input: (<expected_state>, <caller>)
+    # output: (<next_state>)
+    _STATE_SWITCH = {
```
obi-t4, line 420: It seems difficult to read code because S0, S1, S2, P1, P2 variables have no meaning.
If P1 is like P_ota_service, I think it's OK.

app/ota_client.py
```diff
@@ -503,28 +682,19 @@ def _update(
...
-                msg = "cannot ensure all subECUs' are ready, abort current local ota update..."
-                logger.error(msg)
-                raise OtaErrorRecoverable(msg)
+        with fsm.proceed(fsm._P2, expect=fsm._S1) as _next:
+            assert _next == fsm._S2
+            logger.debug("[update] signal ota_service that local update finished")
 
-        logger.info("update finished, entering post-update...")
-        self._update_phase = OtaClientUpdatePhase.POST_PROCESSING
+        logger.debug("[update] leaving update, wait on ota_service and then reboot...")
```
obi-t4, line 696: logger.info is prefer to save log.
```suggestion
        logger.info("[update] leaving update, wait on ota_service and then reboot...")
```

app/ota_client.py
```diff
@@ -710,7 +910,8 @@ def error_callback(e):
                 downloader=self._download,
             )
 
-            with Pool() as pool:
+            _max_workder = min(3, os.cpu_count())
```
obi-t4, line 913: I think this change is for the proxy server not to be congested, but If the number of sub ecu increased, is this value appropriate?

obi-t4 (commented):
I left some comments, but I think there is no big issue.

app/ota_client.py
Bodong-Yang: Sure!

app/ota_client.py
Bodong-Yang: Yes! The outer retry will also retry on the failure of inner `urllib3.Retry`, it is a little bit redundant but I think it is not a very big problem.

app/ota_client.py
Bodong-Yang: If developer needs details about the state machine, they can refer to the comments of `OtaStateSync` though. Anyway I will  rename P1, P2 to P_ota_service and P_ota_client accordingly.

app/ota_client.py
Bodong-Yang: I think it is OK, I have tried small scale test with 5 downloading program with more workers to request files from a single ota_proxy server, and it works as expected.

obi-t4 (approved):
standalone ecu update and multiple ecu update has been confirmed!
LGTM :+1:

PR #115, opened 2022-03-07
add iter_content retry

obi-t4:
# About this PR.
This PR fixes issue when `response.iter_content` raises timeout error, Retry function doesn't retry.
In the previous implementation, retry is done manually so iter_content timeout was handled, but the current implementation is not.

`https://github.com/tier4/ota-client/pull/115?w=1` is easer to see diff.

# issue
https://tier4.atlassian.net/browse/T4PB-15676

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -53,6 +53,7 @@ class Downloader:
     CHUNK_SIZE = 2 * 1024 * 1024  # 2MiB
     RETRY_COUNT = 5
     BACKOFF_MAX = 10
+    BACKOFF_FACTOR = 1
 
     def __init__(self):
         from requests.adapters import HTTPAdapter
@@ -73,7 +74,7 @@ class Downloader:
         retry_strategy = Retry(
             total=self.RETRY_COUNT,
             raise_on_status=True,
-            backoff_factor=1,
+            backoff_factor=self.BACKOFF_FACTOR,
             # retry on common server side errors and non-critical client side errors
             status_forcelist={413, 429, 500, 502, 503, 504},
             allowed_methods=["GET"],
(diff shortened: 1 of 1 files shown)


app/ota_client.py
```diff
@@ -113,36 +114,50 @@ def __call__(
...
+                        for data in response.iter_content(chunk_size=self.CHUNK_SIZE):
+                            hash_f.update(data)
+                            f.write(data)
+                        break
+                    except Exception:  # catch all exceptions
+                        error_count += 1
+                        if error_count > self.RETRY_COUNT:
+                            logger.exception("error: iter_content")
+                            raise
+                        time.sleep(self.BACKOFF_MAX)
```
Bodong-Yang, line 142: If I understand correctly, here we sleep `self.BACKOFF_MAX` for every retry, is that right?

obi-t4: Yes, when iter_content raises, sleep self.BACKOFF_MAX(=10s) every time.

Bodong-Yang: Is it too long for the first retry on `iter_content`? Or should we use a backoff function to generate the for every retry delay here?

obi-t4: get_backoff_time has been implemented.

app/ota_client.py
```diff
@@ -107,42 +109,61 @@ def _path_to_url(base: str, p: Union[Path, str]) -> str:
...
-            raise OtaErrorRecoverable(msg)
-
-        # prepare hash
-        hash_f = sha256()
-        with open(dst, "wb") as f:
-            for data in response.iter_content(chunk_size=self.CHUNK_SIZE):
-                hash_f.update(data)
-                f.write(data)
+
+        while True:
```
Youneko, line 123: [IMO] Retry()によるリトライでリトライされていないパターンを補っているように見えました。[retrying module](https://github.com/rholder/retrying)を使ってリトライ処理を1本にした方が良いように思いましたがどうでしょうか？

pga2rn (approved):
I think it should work as expected!

pga2rn (approved):

app/ota_client.py
pga2rn: The module you referenced is no-longer maintained.😥
And here we already use the urllib3.retry to handle connection problem. But it seems sometime there are timeout problem in iter_content, so another simple retry implemented in this PR to re-do connection with urllib3.retry I think.

obi-t4:
@Bodong-Yang @Youneko 
We have plans to refactor this function, so I'll close this PR.

PR #116, opened 2022-03-11
[T4PB-13999] add to build pb2 whl

Youneko:
# Introduction
Build whl package for protobuf files to install pip command by ota client user.

https://tier4.atlassian.net/browse/T4PB-13999

# Changes
- Add Makefile to build proto file and whl package.
- Add latest whl packages.

# Verification
- [x] install whl package with pip command.
- [x] import python package of protobuf after whl package installation.


proto/Makefile
```diff
@@ -0,0 +1,28 @@
+PB2_OUT_BASE_DIR := pb2
+PB2_OUT_DIR := $(PB2_OUT_BASE_DIR)/otaclient_pb2
+
+.PHONY: all
+all: whl
+
+.PHONY: clean
+clean:
+	@rm -rf build *.egg-info dist $(PB2_OUT_DIR)
```
obi-t4, line 9: I think `@` can be removed since the command log is not so redundant. 
It is better to remove it so that the operation will be clearer.

obi-t4 (commented):
I left some comments. Please have a look :+1:

proto/Makefile
Youneko: > I think `@` can be removed since the command log is not so redundant. It is better to remove it so that the operation will be clearer.

I removed `@` in Makefile.

obi-t4 (approved):
LGTM :+1:

PR #117, opened 2022-03-25
Fix large memory use caused by the use of multiprocessing.imap_unordered.

Bodong-Yang:
# Introduction
Previously, I introduce `multiprocessing.imap_unordered` to accelerate the `RegularInf` parsing.
But after small scale tests with `memory profiling` and `timeit`, I found that comparing to the time it saves(around 10seconds), it consume 200MB~ extra memory, which is not so efficient.
This PR dips the use of `multiprocessing.imap_unordered`, and also reduce some memory footprint for `RegularInf` class.

# Test and result
For one mainecu and one subecu environment, the memory usage on the mainecu drops from `867M~` to `664M~`.


Bodong-Yang:
( I use the previous PR as base to commit this PR, so the commits for this PR are large in number, but it is not a problem as the previous PR has been merged.

app/ota_client.py
```diff
@@ -297,6 +306,8 @@ class PersistentInf(_BaseInf):
     Persistent file information class
     """
 
+    __slots__ = ["path"]
+
```
obi-t4, line 310: adding `del self._left` in the `__init__` is nice to have.

app/ota_client.py
```diff
@@ -246,44 +248,51 @@ def __init__(self, info: str):
...
 
     _pattern = re.compile(r"'(?P<link>.+)((?<!\')',')(?P<target>.+)'")
 
+    __slots__ = ["slink", "srcpath"]
+
     def __init__(self, info):
         super().__init__(info)
         res = self._pattern.match(self._left)
         assert res is not None
+        del self._left
```
obi-t4, line 273: add blank line before `del self._left` same as DirectoryInf and RegularInf. Or remove all blank lines before it for each class.

obi-t4 (commented):
I left some comments, but basically, it is good.

obi-t4:
@Bodong-Yang Commit history in this PR seems to be strange. Merge from main or some operations seem to be required.

Bodong-Yang:
Superseded by #118.

app/ota_client.py
```diff
@@ -231,12 +231,14 @@ class _BaseInf:
...
 
+    __slots__ = ["mode", "uid", "gid", "_left"]
+
     @staticmethod
     def de_escape(s: str) -> str:
         return s.replace(r"'\''", r"'")
 
     def __init__(self, info: str):
-        match_res: re.Match = self._base_pattern.match(info.strip("\n"))
+        match_res: re.Match = self._base_pattern.match(info.strip())
```
obi-t4, line 241: @Bodong-Yang This strip() strips not only `\n` but also ` `?
This should be reverted?

PR #118, opened 2022-04-05
Fix large memory use caused by the use of multiprocessing.imap_unordered.

Bodong-Yang:
### Supersede #117

# Introduction
Previously, I introduce `multiprocessing.imap_unordered` to accelerate the `RegularInf` parsing.
But after small scale tests with `memory profiling` and `timeit`, I found that comparing to the time it saves(around 10seconds), it consume 200MB~ extra memory, which is not so efficient.
This PR dips the use of `multiprocessing.imap_unordered`, and also reduce some memory footprint for `RegularInf` class.

# Test and result
For one mainecu and one subecu environment, the memory usage on the mainecu drops from `867M~` to `664M~`.

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -231,59 +231,76 @@ class _BaseInf:
         r"(?P<mode>\d+),(?P<uid>\d+),(?P<gid>\d+),(?P<left_over>.*)"
     )
 
+    __slots__ = ["mode", "uid", "gid", "_left"]
+
     @staticmethod
     def de_escape(s: str) -> str:
         return s.replace(r"'\''", r"'")
 
     def __init__(self, info: str):
-        match_res: re.Match = self._base_pattern.match(info.strip("\n"))
+        match_res: re.Match = self._base_pattern.match(info.strip())
         assert match_res is not None
+
         self.mode = int(match_res.group("mode"), 8)
         self.uid = int(match_res.group("uid"))
         self.gid = int(match_res.group("gid"))
-
         self._left: str = match_res.group("left_over")
 
 
 class DirectoryInf(_BaseInf):
-    """
-    Directory file information class
-    """
+    """Directory file information class."""
+
+    __slots__ = ["path"]
 
     def __init__(self, info):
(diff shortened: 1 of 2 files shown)


Bodong-Yang:
black is broken on version<22.3.0.
https://github.com/psf/black/issues/2964

app/ota_client.py
```diff
@@ -231,74 +231,93 @@ class _BaseInf:
...
 
+    __slots__ = ["mode", "uid", "gid", "_left"]
+
     @staticmethod
     def de_escape(s: str) -> str:
         return s.replace(r"'\''", r"'")
 
     def __init__(self, info: str):
-        match_res: re.Match = self._base_pattern.match(info.strip("\n"))
+        match_res: re.Match = self._base_pattern.match(info.strip())
```
obi-t4, line 241: @Bodong-Yang This strip() strips not only \n but also (=space)?
This should be reverted?

Bodong-Yang: OK!

obi-t4: > end with white space?

I think info.strip is not only `end with white space`. Is my understanding correct?

obi-t4: Ah, I thought "a b c".strip() returns "abc", but it returns "a b c".
But " a b c ".strip() returns "a b c", so I think strip("\n") should be used.

Bodong-Yang: @obi-t4 Ah I mean it there any case that info string will need to preserved the ending white space? Path itself is wrapped with `'` so I think any white space at the end of info string is useless?

obi-t4: Yes, you're right! strip() is fine!

Bodong-Yang: > Ah, I thought "a b c".strip() returns "abc", but it returns "a b c". But " a b c ".strip() returns "a b c", so I think strip("\n") should be used.

Yes, `strip` only strip alway the prefix/suffix ` ` or `\n`

obi-t4 (approved):
LGTM :+1:

PR #119, opened 2022-04-13
fix: use cp -T not to create {nvme}/boot/boot directory

obi-t4:
# About this PR.
Current implementation copies {standby nvme}/boot to {standby emmc}/boot and {standby emmc}/boot/boot is created as a result.
This PR added -T option not to create boot directory under {standby emmc}/boot.

# issue
https://tier4.atlassian.net/browse/T4PB-16833


Bodong-Yang (approved):
LGTM!

PR #120, opened 2022-04-13
Temporary fix for ota-proxy performance.

Bodong-Yang:
# Introduction
Ota-update is much slower(5x times) with ota-proxy than without ota-proxy.
According to the investigation, the bottleneck is the cache entry committing into sqlite db in ota-proxy's cache implemenation.
This PR provide a first-aid fix to avoid this issue by disable the ota-proxy's cache mechanims, make ota-proxy a pure proxy server.

# Changes
1. cache mechanism is disabled for ota-proxy.
2. fix an issue of ota-cache module initializing.

### DIFF
--- a/app/ota_client_stub.py
+++ b/app/ota_client_stub.py
@@ -71,9 +71,12 @@ class OtaProxyWrapper:
         import uvicorn
         from ota_proxy import App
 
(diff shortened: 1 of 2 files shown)


obi-t4 (approved):
confirmed update with x1 ecu works properly!
LGTM :+1:

PR #123, opened 2022-04-26
fix: illegal param

obi-t4:
# About this PR

This PR fixes illegal parameters when sub ecu is not connected on update requests.

# ticket
https://tier4.atlassian.net/browse/T4PB-16828


Bodong-Yang (approved):
Thank you for digging out this bug!

PR #125, opened 2022-05-11
ota_client: create_regular_files refactor

Bodong-Yang:
# Introduction
This PR refactor the implementation of create_regular_files.
Previous implementation has the following problem:
1. large max memory usage
2. complicated implementation and sync logic
3. multiprocessing on I/O mainly tasks

# Changes
1. schedule and dispatch create_regular_file in thread-pool
2. new implementation for hardlink file handling
3. callback based stats collecting and processing
4. use `time.thread_time` to track the download time cost
5. special handling for empty file on `Downloader`(create file directly without request remote)
6. other minor updates

# TODO
- [x] local test
- [x] standalone vm test

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -624,11 +624,9 @@ class _CreateRegularStatsCollector:
             self.done_event.set()
             raise
 
-    def done(self):
-        self.done_event.set()
 
-
-_WeakRef = TypeVar("_WeakRef")
+class _WeakRef:
+    pass
 
 
 class _HardlinkTracker:
@@ -645,9 +643,15 @@ class _HardlinkTracker:
     def writer_done(self):
         self._first_copy_ready.set()
 
+    def writer_on_failed(self):
+        self._failed.set()
+
     def subscribe(self) -> Path:
         # wait for writer
         while True:
+            if self._failed.is_set():
+                raise ValueError(f"writer failed on path={self.first_copy_path}")
+
             if self._first_copy_ready.is_set():
                 break
             time.sleep(self.POLLINTERVAL)
@@ -664,27 +668,27 @@ class _HardlinkTracker:
 
 class _HardlinkRegister:
     def __init__(self):
-        self._hash_ref_dict: Dict[str, Event] = weakref.WeakValueDictionary()
+        self._lock = Lock()
+        self._hash_ref_dict: Dict[str, _WeakRef] = weakref.WeakValueDictionary()
         self._ref_tracker_dict: Dict[
-            Event, _HardlinkTracker
+            _WeakRef, _HardlinkTracker
         ] = weakref.WeakKeyDictionary()
 
     def get_tracker(
         self, hash_in: str, path: Path, nlink: int
     ) -> "Tuple[_HardlinkTracker, bool]":
-        _ref = self._hash_ref_dict.get(hash_in)
-        if _ref:
-            _ref.wait()  # wait for writer to fully intialized
-            _tracker = self._ref_tracker_dict[_ref]
-            return _tracker, False
-        else:
-            _ref = Event()
-            _tracker = _HardlinkTracker(path, _ref, nlink - 1)
+        with self._lock:
+            _ref = self._hash_ref_dict.get(hash_in)
+            if _ref:
+                _tracker = self._ref_tracker_dict[_ref]
+                return _tracker, False
+            else:
+                _ref = _WeakRef()
+                _tracker = _HardlinkTracker(path, _ref, nlink - 1)
 
-            self._hash_ref_dict[hash_in] = _ref
-            self._ref_tracker_dict[_ref] = _tracker
-            _ref.set()
-            return _tracker, True
+                self._hash_ref_dict[hash_in] = _ref
+                self._ref_tracker_dict[_ref] = _tracker
+                return _tracker, True
 
 
 class _BaseOtaClient(OtaStatusControlMixin, OtaClientInterface):
@@ -1153,23 +1157,29 @@ class _BaseOtaClient(OtaStatusControlMixin, OtaClientInterface):
 
         # case 2: normal file or first copy of hardlink file
         else:
-            if reginf.path.is_file() and verify_file(
-                reginf.path, reginf.sha256hash, reginf.size
-            ):
-                # copy file from active bank if hash is the same
-                shutil.copy(reginf.path, dst)
-                processed.op = "copy"
-            else:
-                processed.errors = downloader(
-                    reginf.path,
-                    dst,
-                    reginf.sha256hash,
-                )
-                processed.op = "download"
+            try:
+                if reginf.path.is_file() and verify_file(
+                    reginf.path, reginf.sha256hash, reginf.size
+                ):
+                    # copy file from active bank if hash is the same
+                    shutil.copy(reginf.path, dst)
+                    processed.op = "copy"
+                else:
+                    processed.errors = downloader(
+                        reginf.path,
+                        dst,
+                        reginf.sha256hash,
+                    )
+                    processed.op = "download"
 
-            if is_hardlink and _is_writer:
-                # writer indicates that the first copy of hardlink is ready
-                _hardlink_tracker.writer_done()
+                    if is_hardlink and _is_writer:
+                        # writer indicates that the first copy of hardlink is ready
+                        _hardlink_tracker.writer_done()
+            except Exception:
+                # signal all subscribers to abort
+                if is_hardlink and _is_writer:
+                    _hardlink_tracker.writer_on_failed()
+                    raise
 
         processed.size = dst.stat().st_size
 
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -9,12 +9,7 @@ import tempfile
 import threading
 import time
 import weakref
-from concurrent.futures import (
-    FIRST_EXCEPTION,
-    ThreadPoolExecutor,
-    Future,
-    wait as concurrent_futures_wait,
-)
+from concurrent.futures import ThreadPoolExecutor, Future
 from contextlib import contextmanager
 from dataclasses import dataclass
 from enum import Enum, unique
@@ -24,7 +19,7 @@ from json.decoder import JSONDecodeError
 from pathlib import Path
 from queue import Queue
 from threading import Event, Lock
-from typing import Any, Dict, List, Tuple, TypeVar, Union
+from typing import Any, Dict, List, Tuple, Union
 from urllib.parse import quote_from_bytes, urlparse, urljoin
 
 from ota_client_interface import OtaClientInterface
@@ -576,7 +571,7 @@ class _CreateRegularStatsCollector:
         _cur_time = time.time()
         while True:
             if self.abort_event.is_set():
-                logger.error(f"abort event is set, collector exits")
+                logger.error("abort event is set, collector exits")
                 break
 
             if self._store.get_processed_num() < total_num:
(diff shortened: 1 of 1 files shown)


Bodong-Yang:
With this PR, mem usage and update speed are improved during testing.

obi-t4:
@Bodong-Yang before moving from multi-processing to multi-threading, we need a performance comparison.

obi-t4:
@Bodong-Yang 
https://tier4.atlassian.net/browse/T4PB-16973

Bodong-Yang:
> @Bodong-Yang before moving from multi-processing to multi-threading, we need a performance comparison.

Wait a second I am preparing a report about the performance improvement. Once data is gathered I will post it here.

Bodong-Yang:
# Test report
### Image used for test
270265 files, 25G, local prepared image(avoid the impact of network speed and reduce test time cost)

### original implementation
1. summary: 
    mem usage: 702M max
    combined CPU usage(3 process): 200~ avg
    real world time cost: 585s
2. detail
```
  ecu_id: "autoware"
  status {
    status: UPDATING
    progress {
      phase: PERSISTENT
      total_regular_files: 270265
      regular_files_processed: 270265
      files_processed_copy: 44317
      files_processed_link: 144
      files_processed_download: 225804
      file_size_processed_copy: 853513657
      file_size_processed_link: 427019589
      file_size_processed_download: 24460327179
      elapsed_time_copy {
        seconds: 12
        nanos: 25000000
      }
      elapsed_time_link {
        nanos: 455000000
      }
      elapsed_time_download {
        seconds: 721
        nanos: 721000000
      }
      total_regular_file_size: 25740860425
      total_elapsed_time {
        seconds: 585
        nanos: 610000000
      }
    }
  }
```
### With PR
1. summary
    mem usage: 75M max,
    combined CPU usage: 130~ avg
    real world time cost: 483s
2. details
```
  ecu_id: "autoware"
  status {
    status: UPDATING
    progress {
      phase: PERSISTENT
      total_regular_files: 270265
      regular_files_processed: 270265
      files_processed_copy: 44317
      files_processed_link: 144
      files_processed_download: 225804
      file_size_processed_copy: 853513657
      file_size_processed_link: 427019589
      file_size_processed_download: 24460327179
      elapsed_time_copy {
        seconds: 4
        nanos: 878000000
      }
      elapsed_time_link {
        nanos: 1000000
      }
      elapsed_time_download {
        seconds: 402
        nanos: 201000000
      }
      total_regular_file_size: 25740860425
      total_elapsed_time {
        seconds: 483
        nanos: 539000000
      }
    }
  }
```

obi-t4:
@Bodong-Yang Let me confirm, original (before this change) implementation is using the number of 3 processes, and this change uses MAX_DOWNLOAD_WORKERS=7 threads.
I think same number should be used to compare the performance. What do you think about it?
And to ignore proxy bottleneck, the proxy should be bypassed.
I'd like to confirm the test environment.

Bodong-Yang:
> @Bodong-Yang Let me confirm, original (before this change) implementation is using the number of 3 processes, and this change uses MAX_DOWNLOAD_WORKERS=7 threads. I think same number should be used to compare the performance. What do you think about it? And to ignore proxy bottleneck, the proxy should be bypassed. I'd like to confirm the test environment.

1. Yes, in my test environment, ota-proxy is disabled.
2. In previous implementation, I have to set max process to 3 due to memory usage and implementation limitations, but currently these limitations are resolved so we can use 7 threads, that is also an improvement. So I think `MAX_DOWNLOAD_WORKERS=7` should be used.

obi-t4:
> Yes, in my test environment, ota-proxy is disabled.

OK!

> In previous implementation, I have to set max process to 3 due to memory usage and implementation limitations, but currently these limitations are resolved so we can use 7 threads, that is also an improvement. So I think MAX_DOWNLOAD_WORKERS=7 should be used.

Let me confirm, memory usage limitation is for VM memory usage limitation?
The previous implementation - before introducing proxy - used the maximum number of processes.
And what is `implementation limitations`?

Bodong-Yang:
> Let me confirm, memory usage limitation is for VM memory usage limitation? The previous implementation - before introducing proxy - used the maximum number of processes. And what is `implementation limitations`?

Yes, the memory usage for previous implementation of ota-client during ota-update is related to the number of process, approximately 200M per process, so using too much process will result in unacceptable high memory usage. But with PR, this limitation is resolved.

Bodong-Yang:
In short, with this PR, the implementation is simpler than before, using much less memory, and less CPU to achieve faster regular_files creation.

obi-t4:
> so using too much process will result in unacceptable high memory usage.

But this `high memory usage software` is now used for x1 production.
We need to compare both. I have an actual EUC so would you like me to check?

obi-t4:
I know that this PR improves performance, so it should be merged, but we need to investigate how much faster it is.

Bodong-Yang:
> > so using too much process will result in unacceptable high memory usage.
> 
> But this `high memory usage software` is now used for x1 production. We need to compare both. I have an actual EUC so would you like me to check?

Yes the previous one is working(because the current type of ECU have enough memory to use), but not very efficient. This PR makes ota-client much more efficient. 

This is not related to which kind of machine we run ota-client on, the previous implementation itself is not so efficient, and this PR resolves the efficiency problems fundamentally by refactoring.

Bodong-Yang:
> I know that this PR improves performance compare to the current implementation, so it should be merged, but we need to investigate how much faster it is.

As for the report I posted previously, without considering network speed(network speed on local test is extremely fast), this PR reduce memory usage by 10times, reduce total average CPU usage from 200%+ to 130%+, and reduce 20% total time cost.

About the improvement of time cost, the real world ota-update time consuming is related to the implementation and the network speed. This PR reduces the time cost related to the implementation.

As for me, according to my local test result, I think this PR improves the efficiency quite a lot and should be merged before ota-update benchmarking.

obi-t4:
> This is not related to which kind of machine we run ota-client on, the previous implementation itself is not so efficient, and this PR resolves the efficiency problems fundamentally by refactoring.

What we need to know is performance **data** compared to the previous one.

obi-t4:
Anyway I'll start review.

Bodong-Yang:
By the way, by just look at the time cost, this PR might not bring **so much** improvement. 
The main idea for this PR is refactoring the previous complicated and not so efficient implementation, resolving the memory usage problem.

obi-t4:
@Bodong-Yang 
What http local server are you using? (I'm afraid if you are using `python -m http.server 8081`, simultaneous requests from client cannot be handled.)

Bodong-Yang:
> @Bodong-Yang What http local server are you using? (I'm afraid if you are using `python -m http.server 8081`, simultaneous requests from client cannot be handled.)

Yes I am using `python -m http.server` as local file server. What do you mean by `simultaneous requests from client cannot be handled`?

obi-t4:
I think `http.sever` handles request by the single thread. It handles request in order, I think.

Bodong-Yang:
> I think `http.sever` handles request by the single thread. It means http.server handles request in order, I think.

No, calling http.server via `python3 -m http.server <port>` launches a `ThreadingHTTPServer` by default.
```python
# from server.py:L1285~1292
...
    # ensure dual-stack is not disabled; ref #38907
    class DualStackServer(ThreadingHTTPServer):
        def server_bind(self):
            # suppress exception when protocol is IPv4
            with contextlib.suppress(Exception):
                self.socket.setsockopt(
                    socket.IPPROTO_IPV6, socket.IPV6_V6ONLY, 0)
            return super().server_bind()
...
```

obi-t4:
Thanks!

obi-t4:
20211021113620-1(v2.1.1) -> 20220128100932-1(v3.0.0)
|  v2.0.8  |  current(v2.5.3)  | This PR |
| 504s | ---- | ---- |

```
v2.0.8
[2022-05-12 11:00:09,756][INFO]-ota_client_service.py:<module>:130,response=ecu {
  ecu_id: "autoware"
  status {
    status: UPDATING
    version: "1.2.3"
    progress {
      phase: PERSISTENT
      total_regular_files: 256190
      regular_files_processed: 256190
      files_processed_copy: 227704
      files_processed_link: 142
      files_processed_download: 28344
      file_size_processed_copy: 18626809572
      file_size_processed_link: 364119364
      file_size_processed_download: 2430739317
      elapsed_time_copy {
        seconds: 1622
        nanos: 479000000
      }
      elapsed_time_link {
        seconds: 51
        nanos: 712000000
      }
      elapsed_time_download {
        seconds: 1356
        nanos: 803000000
      }
    }
  }
}
```

obi-t4:
@Bodong-Yang 
This PR is much faster than the current one!

Bodong-Yang:
> @Bodong-Yang This PR is much faster than the current one!

Thank you for your testing! I think the improvement is the result of resolving the overhead of implementation.
New implementation is much more efficient than before.

app/ota_client.py
```diff
@@ -873,177 +1062,136 @@ def _create_symbolic_links(self, list_file, standby_path):
...
-        processed["size"] = dst.stat().st_size
+        processed.size = dst.stat().st_size
 
+        # set file permission as RegInf says
         os.chown(dst, reginf.uid, reginf.gid)
         os.chmod(dst, reginf.mode)
 
-        end_time = time.time()
-        processed["elapsed"] = end_time - begin_time
+        processed.elapsed = time.thread_time() - begin_time
```
obi-t4, line 1192: In the current implementation, the actual time is returned. But this implementation, sleep time is reduced. The actual elapsed time should be returned.

app/ota_client.py
```diff
@@ -873,177 +1062,136 @@ def _create_symbolic_links(self, list_file, standby_path):
...
-                    self._set_statistics(processed_list)
+        self._statistics.set("total_regular_files", total_files_num)
+
+        _hardlink_register = _HardlinkRegister()
+        # maximum allowd concurrent tasks
+        _se = threading.Semaphore(self.MAX_CONCURRENT_DOWNLOAD_TASKS)
+        # collector that records stats from tasks and update ota-update status
+        _collector = _CreateRegularStatsCollector(
+            self._statistics, _se, interval=self.COLLECT_INTERVAL
+        )
```
obi-t4, line 1086: I'd like to remove relationship between _CreateRegularStatsCollector and _se.
Like:
```
def done_callback(se):
  _collector.callback()
  se.release()
...
fut.add_done_callback(done_callback)
```
Because _se.acquire() is done in the loop, but _se.release() is done in the _CreateRegularStatsCollector class.
To understand control flow easier.

obi-t4 (commented):
I left some comments. But I think there are no big problems.

app/ota_client.py
Bodong-Yang: About this point, I think sleep time should not be counted.
When thread is sleeping, it doesn't mean the whole program is halted, only this thread becomes interrupt-able and releases GIL, so that other thread can take the place to execute. The execution time elapsed of different threads are overlapped.

In other word, during a thread's sleeping, other threads can also do jobs. So it is inaccurate to use real world time elapsed to measure this thread's time cost. Using real world time elapsed will result in duplicated counting of execution time from different threads. 

In summary, I think `total_elapsed_time` should use `time.time()` or `time.perf_count()` to calculate, actual execution time of each task(by thread worker) should use `time.thread_time()` to calculate.

app/ota_client.py
Bodong-Yang: Sure! I will let the callback to take `_se` as input param and release it.

app/ota_client.py
Bodong-Yang: I will mention this in the source code as comment, thank you!

obi-t4 (approved):
LGTM :+1:

PR #126, opened 2022-05-13
Bug fix-20220513

Bodong-Yang:
# Introduction
This PR fix bugs found on the 20220513 benchmark test.

# Bugs and fix
1. when space hardlimit reached, ota-update will failed
    cause: when reached hardlimit, cache will be disable, failing the cache stream
    solution: disable cache stream when storage reach hardlimit
2. ota-client will hang if apply ota-update when status==SUCCESS
    cause: scrub_cache_event is not set under status==SUCCESS or cache_init=True but ota_client_service always wait for the event
    solution: set scrub_cache_event in all needed situation
3. ota-client will fail if the ota-proxy database scheme changed
    solution: database scheme version is recorded now and reflected in the database table name

# Other changes
1. ota_proxy.uvicorn: concurrency limit raises up to 256
2. disk usage limit raises from 60:70 to 70:80
3. ~~unknown OTA-File-Cache-Control headers will be ignored instead of failing the request~~
5. database scheme now has version control, and version is embedded in the table_name

### DIFF
--- a/ota_proxy/server_app.py
+++ b/ota_proxy/server_app.py
@@ -151,7 +151,6 @@ class App:
         # pass cookies and other headers needed for proxy into the ota_cache module
         cookies_dict: Dict[str, str] = dict()
         extra_headers: Dict[str, str] = dict()
-
         # currently we only need cookie and/or authorization header
         # also parse OTA-File-Cache-Control header
         ota_cache_control_policies = set()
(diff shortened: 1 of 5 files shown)


ota_proxy/server_app.py
```diff
@@ -169,12 +170,10 @@ async def _pull_data_and_send(self, url: str, scope, send):
                     # also preserved the raw headers value string
                     extra_headers[OTAFileCacheControl.header.value] = header[1].decode()
                 except KeyError:
-                    await self._respond_with_error(
-                        HTTPStatus.BAD_REQUEST,
-                        f"bad OTA-File-Cache-Control headers: {header[1]}",
-                        send,
+                    logger.debug(
```
obi-t4, line 173: Is it OK to bypass this exception?

Bodong-Yang: Emmm yes, after thing twice I think I should revert this change.

obi-t4 (approved):
LGTM :+1:

PR #127, opened 2022-05-18
New way to limit resource usage for ota-client during ota-update.

Bodong-Yang:
# Introduction
Previously, we directly cap the `max_workers` of threadpool to `7` for all operations(copy, link, download).
That is not so flexible and not so compatible to different machine. This PR changes the limitation mechanism from capping `max_workers` of threadpool, to using `Semaphore` to limit the total number concurrent tasks(copy, link, download), and apply another Semaphore to specially limit the on-going downloading tasks.

# Changes
1. Apply semaphore(set to 128 currently) to all operations(copy, link, download).
2. Apply semaphore(set to 16 currently) to download operation.
3. Above settings and some other settings are configured at `configs.py`.
4. Use dataclass for all Inf class.

# TODO
- [x] regression test on VM.

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -242,20 +242,24 @@ class Downloader:
         return error_count
 
 
+def de_escape(s: str) -> str:
+    return s.replace(r"'\''", r"'")
+
+
+@dataclass
 class _BaseInf:
-    _base_pattern = re.compile(
+    mode: int
+    uid: int
+    gid: int
+    _left: str = field(init=False, compare=False, repr=False)
+
+    _base_pattern: ClassVar[re.Pattern] = re.compile(
         r"(?P<mode>\d+),(?P<uid>\d+),(?P<gid>\d+),(?P<left_over>.*)"
     )
 
-    __slots__ = ["mode", "uid", "gid", "_left"]
-
-    @staticmethod
-    def de_escape(s: str) -> str:
-        return s.replace(r"'\''", r"'")
-
     def __init__(self, info: str):
         match_res: re.Match = self._base_pattern.match(info.strip())
-        assert match_res is not None
+        assert match_res is not None, f"match base_inf failed: {info}"
 
         self.mode = int(match_res.group("mode"), 8)
         self.uid = int(match_res.group("uid"))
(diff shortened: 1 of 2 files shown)


app/ota_client.py
```diff
@@ -237,98 +242,106 @@ def download(
         return error_count
 
 
+def de_escape(s: str) -> str:
+    return s.replace(r"'\''", r"'")
+
+
+@dataclass
```
obi-t4, line 249: I think dataclass decorator is not required since __init__ function is implemented.
The dataclass with __init__ function seems to be complicated for me.
https://stackoverflow.com/questions/54927763/control-initialize-order-when-python-dataclass-inheriting-a-class
I'd like to revert these changes.

app/ota_client.py
```diff
@@ -556,10 +569,12 @@ def __init__(
         self,
         store: OtaClientStatistics,
         *,
+        max_concurrency_tasks: int,
         interval=1,
     ) -> None:
         self.abort_event = threading.Event()
         self.last_error = None
+        self.se = threading.Semaphore(max_concurrency_tasks)
```
obi-t4, line 577: In the previous review, semaphore member is removed from this class, and semaphore is used as callback argument.
Is this change intended?

obi-t4 (commented):
I left some comments, please have a check.

app/ota_client.py
Bodong-Yang: Oh I don't use the dataclass generated __init__ method, but write one, so the problem stated in your link doesn't exist in our case. Actually the behavior is the same as before, and we get the benefits from the dataclass.

app/ota_client.py
Bodong-Yang: Yes, it is intended.
Previously this semaphore is created outside `_CreateRegularStatsCollector`, and used as a parameter for callback, and directly used in the threadpool for loop. 
Now I change to initialize the semaphore in the `_CreateRegularStatsCollector` class, and the this semaphore is accessed as `_CreateRegularStatsCollector` instance's property in the for loop.

What I want to achieve is to let the Collector instance take the whole control of acquire/release semaphore. (Oh for better understanding, maybe I should wrap `_collector.se.acquire()` in line 1107 into a method?)

Bodong-Yang: Ah I may also let the collector control the exception detection.
https://github.com/tier4/ota-client/blob/00ca4b70bb009ee2a67817ac7c4da12626676c59/app/ota_client.py#L1119-L1132

app/ota_client.py
obi-t4: > we get the benefits from the dataclass.

In this change, what is the benefits from the dataclass?

app/ota_client.py
```diff
@@ -237,98 +242,106 @@ def download(
...
-        del self._left
 
+@dataclass
+class RegularInf:
+    mode: int
+    uid: int
+    gid: int
+    nlink: int
+    sha256hash: str
+    size: int
```
obi-t4, line 322:
```suggestion
    size: Optional[int]
```

app/ota_client.py
```diff
@@ -1170,19 +1183,24 @@ def _create_regular_file(
...
                 else:
-                    processed.errors = downloader(
-                        reginf.path,
-                        dst,
-                        reginf.sha256hash,
-                    )
-                    processed.op = "download"
+                    # limit the concurrent downloading tasks
+                    with download_limiter:
+                        if True:
```
obi-t4, line 1188: Is this not required?

app/ota_client.py
obi-t4: > should wrap _collector.se.acquire() in line 1107 into a method

Yes!, I think it is better.

obi-t4 (commented):
I left some comments, but it seems there is no big issue I think.

app/ota_client.py
Bodong-Yang: Currently most important is dataclass will generate `__repr__` , `__eq__` and `__hash__`  for us automatically. 
`__repr__` is for debug purpose, `__eq__` and `__hash__` will allow customize set/dict behavior when storing Inf related class' instance(this feature will be used in later development).

app/ota_client.py
Bodong-Yang: This replaces the previous implementation for logic of limiting concurrent __downloading__ tasks.

Bodong-Yang: The outer semaphore(which the `_collector` is maintaining), is used to limit the number of total concurrent tasks(download, copy, link).

obi-t4: > Is this not required?

Sorry, `if Ture:` is required?

Bodong-Yang: Oh I am sorry, I mistakenly left a debug used line, I will delete it now.

obi-t4 (approved):
LGTM :+1:

Bodong-Yang:
> LGTM +1

Thanks! I will do a quick VM test for regression testing, after finish it I will merge the PR.

Bodong-Yang:
Test on VM passed with `MAX_CONCURRENT_DOWNLOAD` set to 8 and `MAX_CONCURRENT_TASKS` bump to 1024!

PR #130, opened 2022-05-23
Support new RegularInf scheme and cleanup Inf related classes.

Bodong-Yang:
# Introduction
1. This PR introduce the support for new new RegularInf scheme that supports distinguishing hardlink group.
The new format for RegularInf is defined as follow:
```python3
# add support for distinguishing hardlink group
format: mode,uid,gid,link number,sha256sum,'path/to/file'[,size[,inode]]
example: 0644,1000,1000,1,0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef,'path/to/file',1234,12345678

NOTE: size and inode sections are both optional, if inode exists, size must exist.
```
2. relocate Inf related classes to `ota_metadata.py`.
3. introduce tests files for Inf related classes.
4. update hardlink tracker to support hardlink group.

### DIFF
--- a/app/ota_client.py
+++ b/app/ota_client.py
@@ -609,10 +609,20 @@ class _HardlinkRegister:
         ] = weakref.WeakKeyDictionary()
 
     def get_tracker(
-        self, hash_in: str, path: Path, nlink: int
+        self, _identifier: str, path: Path, nlink: int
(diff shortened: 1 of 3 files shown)


obi-t4 (approved):
The code seems OK.

PR #131, opened 2022-05-25
Update proxy_info.py

Bodong-Yang:
# Introduction
This PR introduces updates both to the format/fields of `proxy_info.yaml` and the implementation of `proxy_info.py`.

# Changes
1. introduce new proxy_info format, add field `enable_ota_proxy_cache`, split `local_server` into `ota_proxy_listen_addr` and `ota_proxy_listen_port`.
2. refactor `proxy_info.py`
3. add test files for `proxy_info`
4. adjustment and fix according to the updates at the `proxy_info.py`

# Behavior changes
1. perception ECUs don't require to enable ota-proxy any more, as all the files has been cached by the mainECU. When the perception ECU's ota update is interrupted, it can use the files cached by mainECU to recover the ota update.

# proxy_info format change
### Updated format
```python
    """OTA-proxy configuration.

    NOTE: all the default values are for mainECU!
    Attributes:
        enable_local_ota_proxy: whether to launch a local ota_proxy server, default is True.
        gateway: (only valid when enable_local_ota_proxy==true) whether to enforce HTTPS when local ota_proxy
            sends out the requests, default is True.
        local_ota_proxy_enable_cache: enable cache mechanism on ota-proxy, default is True.
        local_ota_proxy_listen_addr: default is "0.0.0.0".
        local_ota_proxy_listen_port: default is 8082.
        upper_ota_proxy: the upper proxy used by local ota_proxy(proxy chain), default is None(no upper proxy).
    """

    enable_local_ota_proxy: bool = True
    gateway: bool = True
    # to be compatible with mainECU
    local_ota_proxy_enable_cache: bool = True
    upper_ota_proxy: str = None
    local_ota_proxy_listen_addr: str = server_cfg.OTA_PROXY_LISTEN_ADDRESS
    local_ota_proxy_listen_port: int = server_cfg.OTA_PROXY_LISTEN_PORT
```

### New default proxy_info
For the case that `proxy_info.yml` doesn't exist(for mainECU):
"""yaml
enable_local_ota_proxy: true
gateway: true
local_ota_proxy_enable_cache: true
"""

### Proxy_info template for perceptionECU
```yaml
# for internal ecu that doesn't have direct internet connection,
# the following is the template configuration for it.
"""
# internal ecu is not the gateway for the local network.
gateway: false

# set to false if internal ECU doesn't have child ECU to serve.
enable_local_ota_proxy: false

# typically, we can only enable ota cache on the gateway ECU.
enable_local_ota_proxy_cache: false

# for internal ECU, upper_ota_proxy is required, 
# internal ECU will use this proxy to request for ota update.
# upper ota proxy must be an HTTP URL.
upper_ota_proxy: <upper_ota_proxy_URL: str>

# the listen_addr of local_ota_proxy, if not presented, default to 0.0.0.0:8082
local_ota_proxy_listen_addr: "0.0.0.0"
local_ota_proxy_listen_port: 8082
"""
```


obi-t4 (commented):
I left some comments.

app/proxy_info.py
```diff
@@ -9,71 +11,66 @@
...
 """
 
 ##                        subecu                           ##
 """
-# gateway option should not be presented, or set to false
+# gateway must be false for subECU
 gateway: false
-enable_ota_proxy: true
+
+# not mandatory if subECU doesn't have child ECU
```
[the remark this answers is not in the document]
Bodong-Yang: Ah I can rename `enable_ota_proxy` to `enable_local_ota_proxy` to make it clear.

app/proxy_info.py
```diff
@@ -9,71 +11,66 @@
...
+"""
 enable_ota_proxy: true
 gateway: true
+enable_ota_proxy_cache: true
 """
 
 ##                        subecu                           ##
 """
-# gateway option should not be presented, or set to false
+# gateway must be false for subECU
```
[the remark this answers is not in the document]
Bodong-Yang: OK! I will rephrase this sentence.

app/proxy_info.py
```diff
@@ -9,71 +11,70 @@
...
 # for mainecu, proxy_info.yaml is not needed, the following
 # DEFAULT_PROXY_INFO will be used.
 # we should treat the ecu with no proxy_info.yaml
 # as main ecu(as gateway), and enable ota_proxy without upper proxy.
-DEFUALT_PROXY_INFO = """
-enable_ota_proxy: true
+"""
+enable_local_ota_proxy: true
 gateway: true
+enable_local_ota_proxy_cache: true
```
obi-t4, line 23: local_ota_proxy_enable_cache or enable_local_ota_proxy_cache?

obi-t4: I think `enable_local_ota_proxy_cache` is better. Since we are using `enable_local_ota_proxy`.

obi-t4:
I'd like to support the previous name `enable_ota_proxy:`. USB install image has this value.
From the specification point of view, `enable_local_ota_proxy` is OK and we don't need to specify the old name `enable_ota_proxy`.
Do you think can we support it?
After changing the vehicle setting, we can remove this support.

Bodong-Yang:
> I'd like to support the previous name `enable_ota_proxy:`. USB install image has this value. From the specification point of view, `enable_local_ota_proxy` is OK and we don't need to specify the old name `enable_ota_proxy`. Do you think can we support it? After changing the vehicle setting, we can remove this support.

Do you mean support `enable_ota_proxy` as an alias of `enable_local_ota_proxy`, so that old configuration can be parsed correctly?

Bodong-Yang:
Compatibility to `enable_ota_proxy` is added in [e750427](https://github.com/tier4/ota-client/pull/131/commits/e750427e99314867b565bc6c710391001ddbca78).

app/proxy_info.py
```diff
@@ -99,7 +107,28 @@ def parse_proxy_info(proxy_info_file: str = cfg.PROXY_INFO_FILE) -> ProxyInfo:
...
+
+        _proxy_info_dict[_field.name] = _option
+
+    # maintain compatiblity with old proxy_info format
+    for old, new in ProxyInfo._compatibility.items():
+        if old in _loaded:
+            warnings.warn(
+                "'enable_ota_proxy' is replaced by 'enable_local_ota_proxy', "
+                "and the support for 'enable_ota_proxy' option might be dropped in the future "
+                "please use 'enable_local_ota_proxy' instead.",
```
obi-t4, line 127: This should be fine.
```suggestion
            warnings.warn(
                f"'{old}' is replaced by '{new}', "
                f"and the support for '{old}' option might be dropped in the future "
                f"please use '{new}' instead.",
```

Bodong-Yang: Ah yes! I make a generic method but hardcode the value, what a miss!

PR #135, opened 2022-06-24
add log size max handling

obi-t4:
# ticket
https://tier4.atlassian.net/browse/T4PB-15232


obi-t4:
This RP is unnecessary since the current implementation has a local HTTP log server w/ a single thread and sends log events to the cloud once every 2 seconds.
Sending 10_000 logs in 2 seconds or 1MiB logs in 2 seconds to the local HTTP server is difficult from the performance point of view.
So I'll close this PR.

PR #137, opened 2022-07-20
Fix total_regular_file_size doesn't exist & cleanup ota_client.py

Bodong-Yang:
- fix [JIRA] (T4PB-19045) [OTA] [refactor3] total_regular_file_size doesn't exist
- update the way of converting internal status class to compiled v2 proto defined structure
- ota_client.OTAUpdater: move boot_control pre_update execution to _pre_update method

### DIFF
--- a/app/update_stats.py
+++ b/app/update_stats.py
@@ -3,15 +3,19 @@ import time
 from enum import Enum
 from concurrent.futures import ThreadPoolExecutor
 from contextlib import contextmanager
+from google.protobuf.duration_pb2 import Duration
 from queue import Empty, Queue
 from threading import Event, Lock
-from typing import Any, Dict, Generator, List
+from typing import Generator, List
 
 from app.configs import config as cfg
+from app.proto import otaclient_v2_pb2 as v2
 
 
 @dataclasses.dataclass
 class OTAUpdateStats:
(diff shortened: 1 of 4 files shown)


app/update_stats.py
```diff
@@ -146,16 +161,20 @@ def collector(self):
...
                     for st in self._staging:
-                        _suffix = st.op.value
-                        if _suffix in {"copy", "link", "download"}:
+                        if (
+                            isinstance(st.op, RegProcessOperation)
+                            and st.op != RegProcessOperation.OP_UNSPECIFIC
+                        ):
+                            _suffix = st.op.value
                             staging_storage.regular_files_processed += 1
+                            staging_storage.total_regular_file_size += st.size
```
obi-t4, line 170: total_regular_file_size is static value. not dynamic value.

Bodong-Yang: Oh, I miss understanding this field.

Bodong-Yang: fixed in [d72a041](https://github.com/tier4/ota-client/pull/137/commits/d72a041b118aa22bcf01dcf6f50a80a321de2365), the real cause of the problem is that I set the total_regular_file_size before launching the update_stats_collecter.

obi-t4 (approved):
LGTM :+1:

PR #138, opened 2022-07-20
Fix T4PB-19071: ota status become FAILURE before reboot.

Bodong-Yang:
### Main changes
This PR fix T4PB-19071. The cause of the issue is the otaclient shutdown ota updater before reboot, that is not proper behavior.
With this PR, otaclient now only shutdown ota updater on failure(exception), cause the ota status to become FAILURE on update failed.

### Other changes
1. refinement on converting internal use classes to v2 proto compiled classes


Bodong-Yang:
closed due to choose wrong base.

PR #139, opened 2022-07-20
Fix T4PB-19071: ota status return RECOVERABLE before reboot.

Bodong-Yang:
1. Main changes
This PR fix T4PB-19071. The cause of the issue is the otaclient shutdown ota updater before reboot, that is not proper behavior.
With this PR, otaclient now only shutdown ota updater on failure(exception), cause the ota status to become FAILURE on update failed.

2. Other changes

- refinement on converting internal use classes to v2 proto compiled classes
- refinement on status query chain

### DIFF
--- a/app/errors.py
+++ b/app/errors.py
@@ -2,6 +2,8 @@
 import traceback
 from enum import Enum, unique
 
+from app.proto import otaclient_v2_pb2 as v2
+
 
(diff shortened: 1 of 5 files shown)


obi-t4 (approved):
Confirmed the issue is solved.

PR #143, opened 2022-07-28
Introduce wrapper for generated protobuf code

Bodong-Yang:
## Introduction
The generated protobuf code is not so convenient for directly use, according to protobuf documentation, 

> **Note**
> Protocol buffer classes are basically data holders (like structs in C) that don't provide additional functionality; they don't make good first class citizens in an object model.

If we want to bind methods to customize the behaviors of the protobuf classes, the documentation recommends to implement wrappers around the generated classes.

> If you want to add richer behavior to a generated class, the best way to do this is to wrap the generated protocol buffer class in an application-specific class. 

This PR implements a proxy type of wrappers for the generated protobuf code under the `app/proto` folder.
**NOTE**: The implemented wrappers are not yet integrated into the otaclient, the integration will be introduced in a future PR.

### DIFF
--- a/app/proto/__init__.py
+++ b/app/proto/__init__.py
@@ -36,5 +36,6 @@ del _import_proto, _import_from_file
 
 import otaclient_v2_pb2 as v2
 import otaclient_v2_pb2_grpc as v2_grpc
+from . import wrapper
 
-__all__ = ["v2", "v2_grpc"]
+__all__ = ["v2", "v2_grpc", "wrapper"]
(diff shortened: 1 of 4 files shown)


obi-t4:
@Bodong-Yang Could you introduce a unit test for this PR? Since this is a new code.

Bodong-Yang:
> @Bodong-Yang Could you introduce a unit test for this PR? Since this is a new code.

Sure! I will introduce the test files before it mark it as ready for review.

obi-t4 (approved):
LGTM :+1:

PR #146, opened 2022-08-03
Fix memory leak on create_slot module

Bodong-Yang:
## Introduction
A memory leak is detected on `create_standby.common.SimpleTasksTracker`, which is caused by `SimpleTasksTracker` doesn't release retrieved `Future`. This PR fix this issue by properly storing and releasing added `Future` object.


obi-t4 (approved):
LGTM :+1:

PR #163, opened 2022-09-12
Bugfix: otaclient should wait on subecus with POST_PROCESSING status

Bodong-Yang:
## Introduction
This PR fixes the otaclient's behavior that otaclient should wait on subecus with otastatus as `POST_PROCESSING`.

## Changes
1. `INITIAL` and `POST_PROCESSING` otastatus are now set in the main entrance of `OTAUpdater` class(the `execute` method);
2. remove `OTAUpdater._post_update` method, directly call the `_boot_control.post_update` method as the Updater itself doesn't do anything during the post_update.

### DIFF
--- a/otaclient/app/ota_client.py
+++ b/otaclient/app/ota_client.py
@@ -160,7 +160,7 @@ class _OTAUpdater:
             metadata.verify(cert_file.read_bytes())
             return metadata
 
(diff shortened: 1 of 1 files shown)


obi-t4 (approved):
It seems to be OK!

PR #170, opened 2022-10-03
[bugfix] fix ota_proxy handling URL from request improperly

Bodong-Yang:
## Motivation
ota_proxy previously parsed the request URL and only quote the path of the URL again when re-quoting the whole URL, which is improper and unwanted.

ota_proxy should not parse the leftover parts of the URL besides <scheme>://<netloc>, and should treat all the leftover parts as path because a proxy server should not care about the contents of the request and forward it to the remote as it.

This PR fix the issue mentioned above.

## Ticket
(T4PB-21670)[https://tier4.atlassian.net/browse/T4PB-21670]

### DIFF
--- a/otaclient/ota_proxy/ota_cache.py
+++ b/otaclient/ota_proxy/ota_cache.py
@@ -38,7 +38,7 @@ from typing import (
     TypeVar,
     Union,
 )
-from urllib.parse import quote, urlparse
+from urllib.parse import SplitResult, quote, urlsplit
 
 
 from .db import CacheMeta, OTACacheDB, DBProxy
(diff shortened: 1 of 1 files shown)


obi-t4:
@Bodong-Yang Can you provide unit test?

obi-t4 (commented):
PR seems OK, but we need to implement a unit test.

Bodong-Yang:
> @Bodong-Yang Can you provide unit test?

OK!

Bodong-Yang:
Althoug basic unit test for ota_proxy is working right now, it still needs some polishing.

docker/test_base/Dockerfile
```diff
@@ -8,6 +8,7 @@ ENV OTA_CLIENT_DIR="/ota-client"
 ENV OTA_IMAGE_SERVER_ROOT="/ota-image"
 ENV OTA_IMAGE_DIR="${OTA_IMAGE_SERVER_ROOT}/data"
 ENV VENV="${OTA_CLIENT_DIR}/.venv"
+ENV SPECIAL_FILE="path;adf.ae?qu.er\y=str#fragement"
```
obi-t4, line 11: @Bodong-Yang 
Can you add multi-byte character file name?

Bodong-Yang: @obi-t4 Updated in [014e83b](https://github.com/tier4/ota-client/pull/170/commits/014e83bdb1616b00f2c5af288482a011d1cd696a).

obi-t4 (approved):
LGTM :+1:

PR #173, opened 2022-10-20
fix: python path to use venv

obi-t4:
Use venv's python3

### DIFF
--- a/bootstrap/root/etc/systemd/system/otaclient.service
+++ b/bootstrap/root/etc/systemd/system/otaclient.service
@@ -7,7 +7,7 @@ Wants=network-online.target
 
 [Service]
 Type=simple
-ExecStart=/bin/bash -c 'source /opt/ota/.venv/bin/activate && PYTHONPATH=/opt/ota /usr/bin/python3 -m otaclient'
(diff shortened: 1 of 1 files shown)


obi-t4:
I'll convert to this PR to draft.

obi-t4:
changed to ready-for-review.

Bodong-Yang:
@obi-t4 where will the otaclient installed to? In this service unit file it expects `/opt/ota/otaclient`, but I notice that in firmware version `20221013091746-22.04-1st-sub`, otaclient is located at `/opt/ota/client/otaclient`. 
But it seems that the systemd service unit file will be updated when installed by ansible though.

obi-t4:
@Bodong-Yang 

> in firmware version 20221013091746-22.04-1st-sub, otaclient is located at /opt/ota/client/otaclient.

That is true. It is the user's responsibility to deiced where to install it. So 20221013091746-22.04-1st-sub is OK.

But the code **before** this change, /usr/bin/python3 is referenced even /opt/ota/.venv is used. It is completely wrong.

Bodong-Yang (approved):

PR #175, opened 2022-10-20
Fix sign.pem being downloaded twice

Bodong-Yang:
This PR fix the behavior that sign.pem being downloaded twice.
* certificate is not considered as image metafile anymore, get_metafiles(get_img_metafiles) now only return directory, symboliclink, regular and persistent
* certificate is only download once for verifying metadata


obi-t4 (approved):
:+1:

PR #178, opened 2022-11-09
boot_control module update and refinement

Bodong-Yang:
## Introduction
This PR applies minor update and refinement to boot_control module before introducing raspberry pi support to otaclient.

**NOTE: this PR is one of the PRs that for introducing rapsberry pi support into otaclient.**

## Changes
1. boot_control related configs now moved from `app.configs` to `app.boot_control.configs`,
2. define bootloader type enum, implement new bootloader detection function,
3. move boot_control internal error def from `boot_control.common` to `boot_control._errors`,
4. boot_control module folder re-arrange,
5. ota_client_stub now detects bootloader before initializing OTAClient instance, instead of detecting and write a const in configs.py previously,
6. other related minor updates.


otaclient/app/boot_control/selecter.py
```diff
@@ -0,0 +1,80 @@
...
+
+from ..configs import config as cfg
+from .. import log_util
+
+logger = log_util.get_logger(
+    __name__, cfg.LOG_LEVEL_TABLE.get(__name__, cfg.DEFAULT_LOG_LEVEL)
+)
+
+
+def detect_bootloader(raise_on_unknown=True) -> BootloaderType:
```
obi-t4, line 31: Is it possible to detect bootloader type dynamically?
If it is possible, we don't need to ask field integrator guys to add bootloader type in the ecu_info.yaml.

obi-t4 (commented):
I left a comment but I think it seems good!

otaclient/app/boot_control/selecter.py
Bodong-Yang: @obi-t4 This is the fallback mechanism to detect bootloader if `bootloader` info is not available in `ecu_info.yaml`. 
What I am afraid is that the evidence I use to detect bootloader might become unavailable in the future(i.e., the path to the special file changed, etc.).

Although the breaking changes like the one I mentioned above are very unlikely to happen, but I still prefer explicitly configure `bootloader` type in the `ecu_info.yaml`.

About the backward compatibility, we can still keep the `ecu_info.yaml` unchanged and let this function detect bootloader , but in the future production use(x2 platform), I would like to explicitly set the `bootloader` type in `ecu_info.yaml` when initializing/installing on new devices.

obi-t4: > This is the fallback mechanism

I see. The bootloader type handling with ecu_info.yaml is another PR, right? Understood.

obi-t4 (approved):

Bodong-Yang:
This PR will be merged into the main branch when the leftover PRs are ready, one by one.

otaclient/app/ecu_info.py
```diff
@@ -11,46 +11,118 @@
...
+                    _field.default
+                    if _field.default is not MISSING
+                    else _field.default_factory()  # type: ignore
+                )
+            # parsed _option is available
+            else:
+                _ecu_info_dict[_field.name] = _option
+
+        # initialize ECUInfo inst
+        return cls(**deepcopy(_ecu_info_dict))
```
obi-t4, line 98: I think unpacking operator is not required.

Bodong-Yang: Emmm it is required. ECUInfo doesn't take the whole `_ecu_info_dict` for __init__, it takes the kv pairs in the `_ecu_info_dict`, so we need to extract the kv pairs from the dict.

Bodong-Yang:
PR reviews in #181 is resolved in [f2c061e](https://github.com/tier4/ota-client/pull/178/commits/f2c061eed42be4b05a704d3acc293b3eae815f19).

Bodong-Yang:
@obi-t4 #178 is ready for review and merging into main.

obi-t4 (approved):
LGTM :+1:

PR #179, opened 2022-11-09
fix: enable put ota_proxy log to cloud

obi-t4:
# About this PR

This PR supports to put ota-proxy log to AWS cloud.

# ticket
https://tier4.atlassian.net/browse/T4PB-16829

# test

image


Bodong-Yang:
@obi-t4 I think this solution solve the issue correctly by also registering HTTP log handler to `ota_proxy` as `ota_proxy` runs as a separate process and the main otaclient logging module doesn't capture the logging from `ota_proxy`.

But I think the `log_util` itself needs some refactoring, as it is not needed to use global var to store the logger instance, and also we don't need to get logger from `log_util` in every module.

`logging` module itself is a self-contained module that manages all loggers created during the code running. `logging` module itself will hold a single root logger when it initialized, and we can get a sublogger belongs to root logger by `logging.getLogger("sublogger")`. All loggers will be unique and managed by logging module, which means that multiple time calling to `logging.getLogger("logger_name")` will always get the same logger instance(`logging.getLogger()` will always return the same root logger).

Basically, we only need to initializing the root logger once as follow, the following setting will be the root logger setting,
```python
import logging

logging.basicConfig(format=cfg.LOG_FORMAT)
logging.setLevel(level)
if http_logging_host := os.environ.get("HTTP_LOGGING_SERVER"):
    from otaclient.aws_iot_log_server import CustomHttpHandler

    h = CustomHttpHandler(host=http_logging_host, url=_get_ecu_id())
    fmt = logging.Formatter(fmt=cfg.LOG_FORMAT)
    h.setFormatter(fmt)
    logging.addHandler(h)
``` 

And then in each module, we can get a sublogger(with module full path as name) from root logger as follow:
```python
import logging

```

The sublogger will use the setting(include the handler setting) as root logger if no configuration is applied to the sublogger. 
For example, if we want to change the loglevel just for this sublogger, we can do the following thing:
```python
import logging

logger.setLevel(logging.DEBUG)
```
And this `setLevel` setting only applied to this sublogger. On the other hand, if we don't do `setLevel` for this sublogger, then this sublogger will follow the loglevel of root logger.

### Reference
[Advanced tutorial - python logging](https://docs.python.org/3/howto/logging.html#logging-advanced-tutorial)
[logging flow - python logging](https://docs.python.org/3/howto/logging.html#logging-flow)

obi-t4:
@Bodong-Yang PR has been updated and confirmed that it works!
for basicConfig, `force=True` seems to be required.
And for the `app` directory, we will fix `log_util.py` in the future.

Bodong-Yang:
@obi-t4 I think the current PR is OK!
In the future, in otaclient, actually we can define the logging setting logic as a function, and inject the function to the process launching entry of ota_proxy(in `ota_client_stub.OtaProxyWrapper.launch_entry`). 
By this method, we don't need to introduce duplicate code into ota_proxy code base.

I will do this right after this PR merged, moving the updated `log_setting.py` module to `otaclient.app` folder, and integrate it into otaclient and ota_proxy.

Bodong-Yang (approved):

PR #186, opened 2022-12-16
[Refine] otaproxy: make ota_cache fully async, resolving blocking

Bodong-Yang:
## Introduction
This PR follows #184 , in #184 ota_cache is switching to asyncio+async generator model, and this PR further improve the implementation, resolving blocking code in asyncio workflow, including sha256 hashing and cache_db accessing. Now sha256 hashing and cache_db accessing are wrapped into a proxy class which provides async interface to resolve blocking.

## Motivation
I have done some small scale test, and found that sha256 blocks the async workflow obviously, especailly when there are many small files. In #184 although cache writing is implemented in async, but `_hash.update()` is still being directly called, which introduces blocking in async.

## Test result
I try 2 tests in different settings as follow.
NOTE: the average download speed is hard to reach the full bandwidth because we have many small files,
but it will reach max bandwidth during downloading large file.
### Test 1(5clients, 43ms, 5ms jitter, 0.1% packet loss, 120Mbit)
1. downloaded files: 147339, ~10.4G,
2. timecost: **1865s**(average 44Mbps, 79 files/s),
3. memory usage: **126M**,
4. finished without errors.

### Test 2(5clients, 43ms, 10ms jitter, 1% packet loss, 8Mbit)
1. downloaded files: 147339, ~10.4G,
2. timecost: **12613s**(average 6.6Mbps, 12 files/s),
3. memory usage: **156M**,
4. finished without errors.

The memory usage is almost the same as in #184, and both way better than before #184. And with this PR, the performance is further better than #184 (2274s), and way better than before #184 (2313s).

### DIFF
--- a/otaclient/ota_proxy/ota_cache.py
+++ b/otaclient/ota_proxy/ota_cache.py
@@ -55,33 +55,12 @@ from .errors import (
     StorageReachHardLimit,
 )
 from .config import config as cfg
+from .utils import wait_with_backoff, AioSHA256Hasher
 
 if TYPE_CHECKING:
     import multiprocessing
 
 logger = logging.getLogger(__name__)
-
-
-def get_backoff(n: int, factor: float, _max: float) -> float:
-    return min(_max, factor * (2 ** (n - 1)))
-
-
-async def wait_with_backoff(
-    _retry_cnt: int, *, _backoff_factor: float, _backoff_max: float
-) -> bool:
-    """
-    Returns:
-        A bool indicates whether upper caller should keep retry.
-    """
-    _timeout = get_backoff(
-        _retry_cnt,
-        _backoff_factor,
-        _backoff_max,
-    )
-    if _timeout <= _backoff_max:
-        await asyncio.sleep(_timeout)
-        return True
-    return False
 
 
 _WEAKREF = TypeVar("_WEAKREF")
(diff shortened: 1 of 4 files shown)


otaclient/ota_proxy/ota_cache.py
```diff
@@ -533,16 +502,14 @@ def __init__(
...
+    async def _finalize_caching(self):
         """Commit cache entry to db and rename tmp cached file with sha256hash
         to fialize the caching."""
-        self._cache_commit_cb(self.meta)
-        # for 0 size file, register the entry to db only,
-        # but if the 0 size file doesn't exist, create one
-        if self.meta.size > 0 or not (self._base_dir / self.meta.sha256hash).is_file():
+        await self._cache_commit_cb(self.meta)
+        # if the file with the same sha256has is already presented, skip the hardlink
+        if not (self._base_dir / self.meta.sha256hash).is_file():
```
obi-t4, line 512: `self.meta.size > 0 or not` seems to be moved, intentional?

Bodong-Yang: Yes, and actually the previous one is wrong and will cause bug. 
The current one `if not (self._base_dir / self.meta.sha256hash).is_file()` is correct.

obi-t4 (approved):
LGTM

PR #188, opened 2022-12-23
[Bugfix] otaproxy: fix AsyncStopIteration errors spamming

Bodong-Yang:
## Introduction
This PR fixes otaproxy spams `cache write failed: AsyncStopIteration` logs caused by `_cache_streamer` doesn't skip empty chunk.
When HTTP chunk encoding is enabled, a empty chunk will be sent to indicate the end of chunk streaming, previously the otaproxy will not skip this empty chunk and also send it to client and the caching generator. And when multiple otaproxy chained together, multiple empty chunks will be sent to the client(one more empty chunk for passing one more otaproxy), causing the problem.

## Changes
1. _cache_streamer: skip the empty chunk

### DIFF
--- a/otaclient/ota_proxy/ota_cache.py
+++ b/otaclient/ota_proxy/ota_cache.py
@@ -144,6 +144,11 @@ class CacheTracker(Generic[_WEAKREF]):
             and not self._writer_failed.is_set()
         )
 
+    def get_cache_write_gen(self) -> AsyncGenerator[int, bytes]:
+        if not self._cache_write_gen:
+            raise ValueError("being called before provider is ready")
+        return self._cache_write_gen
+
     async def _provider_write_cache(
         self, *, storage_below_hard_limit: threading.Event
     ) -> AsyncGenerator[int, bytes]:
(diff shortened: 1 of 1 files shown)


obi-t4:
@Bodong-Yang 
Can we re-produce this `AsyncStopIteration errors` on the current main branch?

Bodong-Yang:
> @Bodong-Yang Can we re-produce this `AsyncStopIteration errors` on the current main branch?

I guess if the `upper_ota_proxy` enables the HTTP chunk streaming, then this behavior can be reproduced?

obi-t4:
@Bodong-Yang 

> I guess if the upper_ota_proxy enables the HTTP chunk streaming, then this behavior can be reproduced?

In other words, how did you debug this feature, how did you confirmed that the issue is resolved?
I'd like to make sure it on this review.

Bodong-Yang:
> @Bodong-Yang
> 
> > I guess if the upper_ota_proxy enables the HTTP chunk streaming, then this behavior can be reproduced?
> 
> In other words, how did you debug this feature, how did you confirmed that the issue is resolved? I'd like to make sure it on this review.

Ah I basically debug by reading code along with the syslog you provided to me, and I come to the solution. I will do test to confirm it, please wait for some time.

Bodong-Yang:
@obi-t4 I can reproduce the issue by chaining 3 otaproxy together, and also confirmed that this PR does fix the problem

obi-t4 (approved):
LGTM :+1:

PR #189, opened 2022-12-23
fix: log-group for ota-proxy

obi-t4:
# About this PR
This PR fixes cloudwatch log group for ota-proxy.
Current implementation stores ota-proxy log to the single `ota-proxy` log group even multiple ecu are targeted.
By this change, ota-proxy log is saved with each ecu's log-group.

current:
all ecus logs in the vehicle for ota-proxy are saved in the same `ota-proxy` log group.
image

by this change:
ota-client and ota-proxy logs are saved in the same log group.
image


Bodong-Yang (approved):
LGTM!

PR #192, opened 2023-01-16
[Refine] logging: filter external logging, tune otaproxy logging

Bodong-Yang:
## Introduction
Previously, when network error occurs, the logging stream will be spammed with too much and too verbose logging messages. The causes are as follow:
1. the `log_setting` applies `loglevel` configuration to the root logger, and by default the `loglevel` is `INFO`. When network errors happen, not only the loggings from otaclient itself being captured, but also the loggings from external modules, especially from `request` module. The loggings from `request` module are not needed and it floods the log recording.
2. the logging in `ota_proxy` is too verbose against normal handled errors.

Also, previously each log is attached by the filename of its module, but now there are modules with same name in the otaclient code base. 
This PR introduces the fix to the above problems.

## Changes
1. log_setting: use `CRITICAL` level for root logger, and apply `<loglevel>` setting and attach greengrass logging handler only to the `otaclient` logger.
2. ota_proxy.server_app: tune logging, only use `logging.exception` when unhandled errors occurs, lower the logging level for common handled errors.
3. otaclient: record the actual package name(normally it will be `otaclient`) in `__init__.py`.
4. configs.LOGGING_FORMAT: use logger name(configured to be the `__name__` of each module) instead of filename.


obi-t4 (approved):
LGTM :+1:

PR #196, opened 2023-02-01
[Refine] otaproxy: only return 500 in unhandled unexpected otaproxy internal error

Bodong-Yang:
## Introduction
Previously otaproxy will return 500 on any unsatisfied request except 4xx(currently 403 and 404) error. This behavior is not proper, and confusing the otaclient.

This PR fixes this behavior as follow:
1. return 500 on internal otaproxy error only.
2. return 502(BAD_GATEWAY) on any connection related error.
3. return 503(SERVICE_UNAVAILABLE) on any aiohttp client error except connection error.


obi-t4 (approved):
LGTM :+1:

PR #198, opened 2023-02-08
[Security] bump cryptography to 39.0.1, pyopenssl to 23.0.0

Bodong-Yang:
## Introduction
Bump cryptography to 39.0.1 and pyopenssl to 23.0.0 due to security concerns, check #197 for details. 
Also remove the deps to `jwt` package as not in use and caused the deps conflicts.

NOTE: pyopenssl version bump is required for cryptography version bump.

### DIFF
--- a/otaclient/requirements.txt
+++ b/otaclient/requirements.txt
@@ -1,6 +1,5 @@
-pyOpenSSL==20.0.1
-jwt==1.2.0
-cryptography==3.4.8
+pyOpenSSL==23.0.0
+cryptography>=39.0.1, <40.0.0
 grpcio==1.48.2
 protobuf==3.20.3
 PyYAML>=3.12


Bodong-Yang:
@obi-t4 It seems that the `jwt` package dependency is introduced a long time ago, but at the time the `requirements.txt` is created(around 20210601), the `jwt` package is not used in `ota_metadata.py` and jwt parsing is done manually without using any extra package.

obi-t4:
> So pyjwt is the correct one that should be included in requirements.txt

Let me confirm, pyjtw package is not required now since `jwt parsing is done manually`.

obi-t4 (approved):
Thanks!

Bodong-Yang:
> Let me confirm, pyjtw package is not required now since jwt parsing is done manually

@obi-t4  Yes it is true for otaclient. 
Also ota-metadat/metadata_gen.py is generating jwt manually without any extra package.

By the way, just curious but why we are generating metadata.jwt manually? When I am checking the source code history for otaclient, it seems that at first `pyjwt` is used but then manually creating is used.

obi-t4:
> why we are generating metadata.jwt manually

Generating means parsing?

> it seems that at first pyjwt is used but then manually creating is used.

I don't know!

Bodong-Yang:
> Generating means parsing?

parsing on otaclient side, and generating on ota-metadata/metadata_gen.py.

obi-t4:
> ota-metadata/metadata_gen.py.

For metadata side, I didn't know about jwt package.

PR #200, opened 2023-02-16
[Refactor] re-implement proto wrapper

Bodong-Yang:
## Introduction
A library that can be used to create pure python implemented wrapper for protobuf messages simply by type annotating without directly depending on compiled protobuf message types. 

The feature of this library are as follow:

1. **for unset/unpopulated field**, aligns the behavior of protobuf message, **wrapper will return the default value according to the field type(int, float will be 0, bool will be false, str will be empty string, enum will be the first defined value in order**, **nested message will be empty message**). 
1. supports **easily creating wrapper with type annotating only** for any generated protobuf message types , custom helper methods can be defined for each created wrappers,
1. **fully supports nested message, repeated fields and `Duration` type**,
1. **fully inter-operatable with protobuf message with `convert` and `export_pb` API,** supports converting protobuf message instance into corresponding wrapper type instance, the converted wrapper instances can be used as normal python instances,
1. simple and unified core APIs, supports exporting wrapper instance back to protobuf message instance, the exported protobuf message instance is equal(each attribute fields have the same value) to the converted one at the beginning,
1. fully type hinted and type annotated.

For more details please check `app/proto/README.md`.

## Motivation
Due to the compiled protobuf message types prohibite the direct inheritance and customizing, previously when handling incoming API request, we directly wrap the incoming protobuf message instance within a simple wrapper(origin message instance is stored in `data` attribute), and all the attribute access is pass-through to the underlaying `data` attribute, and using `typing.cast` method to pretend the wrapper is inherited from the protobuf message type.

This method has several drawbacks, including:
1. cannot handling nested and complicated message very well, 
1. still depends on the protobuf message and directly utilize the protobuf message API,
1. python protobuf is implemented as C++ extension that python runtime cannot fully control,
1. the wrapper is confusing as it exposes both protobuf API and custom helper methods,

Actually it seems that even up to protoc==3.21.12, protobuf==4.21.12, there is a potential memory leak issue in the python protobuf package, we cannot fix the problem as external C++ extension's behavior is out-of python runtime control. Also protobuf python is very complicated internally, we don't want to deal with those complexities in our application code.

To fundamentally solve the problem, this PR introduce a library that can be used to easily generate wrappers simply with type annotation. The provided base class will parse the type annotation, and create wrapper only with the typing information. 

## TODO
- [x] proto wrapper core implementation
- [x] otaclient_v2_pb2 wrapper definition update
- [x] implement test files for new wrapper
- [x] pass all test files
- [x] documentation
- [x] test on VM


Bodong-Yang:
@obi-t4 The PR is updated to cover the review comments, please have a check! Also some minor updates are made at [64e3728](https://github.com/tier4/ota-client/pull/200/commits/64e3728bf778e644c6e22470345d6fba58e4f0c5).

obi-t4 (approved):
LGTM :+1:

PR #201, opened 2023-02-28
[Fix] app.proto: fix related to container field

Bodong-Yang:
This PR fixes some bugs related to container fields.

* fix the behavior when field descriptor for container field handling default value. It should set the value with empty container wrapper instance, not plain list/dict.
* `MessageMappingContainer` now deprecates the `export_pb` and `MessageMappingContainer` should only use `export_to_pb_msg_mapping_container` to export to protobuf.
* the `_FieldBase` abc now marks `__set__` method as abstracted method, all detailed field descriptor implementations must implement the `__set__` method.

More general tests are implemented to cover most of the cases which we support.


obi-t4 (commented):
Is it possible to add test code for this fix?

Bodong-Yang:
> Is it possible to add test code for this fix?

Sure! I will add a test to test the behavior related to default behavior.

otaclient/app/proto/_common.py
```diff
@@ -232,11 +233,27 @@ def convert(
         return res
 
     def export_pb(self) -> Dict[_K, Any]:
+        warnings.warn(
```
obi-t4, line 236: @Bodong-Yang 
Is this warn required?
I think abort is enough.

Bodong-Yang: Abort, do you mean directly return?

obi-t4: No, this function is internal, and if this MessageMapContainer::export_bp is called, it is bug.
So raising exception or something should be done, if I'm undersdanding it correctly.

obi-t4 (approved):
LGTM :+1:

PR #190, opened 2023-01-13
[Refinement] increase MAX_DOWNLOAD_THREAD and DOWNLOADER_CONNPOOL_SIZE_PER_THREAD

Bodong-Yang:
# Introduction
In #183, the MAX_DOWNLOAD_THREAD is decreased from 8 to 3 due to otaproxy and otaclient seem to be not so stable under large amount of connections. 
At that time it seems that this change doesn't reduce too much performance. But it turns out this change does degrade the performance significantly(nearly double the time cost under the same environment). 
But since #186, the otaproxy becomes much more stable and efficient than before. And the local test(check the following comment for details) shows that MAX_DOWNLOAD_THREAD==3 doesn't max out the otaclient's efficiency, and change the MAX_DOWNLOAD_THREAD back to 7 does resume the performance as before #183, meanwhile doesn't increase CPU usage and Memory Usage footprint thanks to optimization over otaproxy and otaclient.
So this PR change the MAX_DOWNLOAD_THREAD of otaclient back to 7.

# Changes
1. increase `MAX_DOWNLOAD_THREAD` back to 7
2. increase `DOWNLOADER_CONNPOOL_SIZE_PER_THREAD` to 10

# Ticket
[T4PB-25078](https://tier4.atlassian.net/browse/T4PB-25078)

[T4PB-25078]: https://tier4.atlassian.net/browse/T4PB-25078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #203, opened 2023-03-06
[Refine] downloader: improve chunk writing and decompression

Bodong-Yang:
## Introduction
This PR refines the implementation of downloader decompressing and saving data chunk.

## Test result
In my local setup, with compression enabled, the time cost of whole ota update reduce from *1088* down to *990*, considering my PC uses nvme and test is done with local prepared ota image, I think the performance in the real environment should also be improved in some extend.

## related issue report
https://tier4.atlassian.net/browse/T4PB-25078, which is handled in https://github.com/tier4/ota-client/pull/190.

### DIFF
--- a/otaclient/app/downloader.py
+++ b/otaclient/app/downloader.py
@@ -31,6 +31,7 @@ from typing import (
     ByteString,
     Callable,
     Dict,
+    Iterator,
     Optional,
     Protocol,
     Tuple,
@@ -140,9 +141,7 @@ class DecompressionAdapterProtocol(Protocol):
     """DecompressionAdapter protocol for Downloader."""
 
     @abstractmethod
-    def stream_decompressor(
-        self, src_stream: Union[IO[bytes], ByteString]
-    ) -> IO[bytes]:
+    def iter_chunk(self, src_stream: Union[IO[bytes], ByteString]) -> Iterator[bytes]:
         """Decompresses the source stream.
 
         This Method take a src_stream of compressed file and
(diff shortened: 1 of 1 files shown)


Bodong-Yang:
This PR only changes `downloader.py`, so It is possible to back-port this PR to v3.2.x.

obi-t4:
@Bodong-Yang 
This PR improves 9% on your environment and that environment is low I/O usage. So why do you think

> the improvement will be significant in real environment.

On the real environment, IO usage is higher than your environment one, and on the high IO usage environment, decompression time can be hided by the IO usage.

Bodong-Yang:
> @Bodong-Yang (This PR itself is OK) This PR improves 9% on your environment and that environment is low I/O usage. So why do you think
> 
> > the improvement will be significant in real environment.
> 
> On the real environment, IO usage is higher than your environment one, and on the high IO usage environment, decompression time can be hided by the IO usage.

My environment is I/O heavy because there is very little diff between my VM image and OTA image(the otaclient will have to download 7GB even with compression enabled, basically downloading the whole image).

But anyway that is only my guess, the test in real environment is still needed.

Bodong-Yang:
( If it is possible, we can backport this PR to v3.2.0, and prepare a image for x1 team to test.

Bodong-Yang:
@obi-t4 I remember that we have handled similar issue(https://tier4.atlassian.net/browse/T4PB-25078) for v3.2.2, and this issue also existed in v3.2.0(download thread num is too small(3), the fix(#190) is to enlarge download thread to 7), I think that is the main cause.

But anyway this PR also improves the performance, so it is better to apply it also.

obi-t4:
> I remember that we have handled similar issue(https://tier4.atlassian.net/browse/T4PB-25078) for v3.2.2, and this issue also existed in v3.2.0(download thread num is too small(3), the fix(https://github.com/tier4/ota-client/pull/190) is to enlarge download thread to 7), I think that is the main cause.

Yes, this issue has been fixed by changing number of thread to 7 #190 

> But anyway this PR also improves the performance, so it is better to apply it also.

Yes, I think so, I'm waiting your:
> the test in real environment is still needed.

PR itself is OK since this can reduce usage of memory I think.

obi-t4 (approved):
LGTM :+1:

Bodong-Yang:
The improvement is confirmed on v3.2.3!
Test setup: 
- **total_regular_files: 23637537004**
- **processed_files_download: 20703405289**

1. Without this PR, time cost is 1008 seconds
2. with this PR, time cost is 980 seconds, the memory usage is also slightly reduced(10MB less than 1)

I will merge this PR now.

PR #204, opened 2023-03-08
feat: support firmware

obi-t4:
# About this PR

This PR supports firmware updating for rqx-58g.
If firmware.yaml doesn't exit, this function does nothing.
The firmwares are located on the OTA image.

# ticket

https://tier4.atlassian.net/browse/T4PB-26551


obi-t4:
@Bodong-Yang 
This PR is currently draft since this is not confirmed yet on the actual OTA image, but the code itself is fixed.

obi-t4:
@Bodong-Yang I confirmed with https://github.com/tier4/autoware_ecu_system_setup/pull/561. I'll change this PR to ready for review.

otaclient/app/boot_control/_cboot.py
```diff
@@ -485,6 +486,13 @@ def pre_update(self, version: str, *, standby_as_ref: bool, erase_standby=False)
 
     def post_update(self):
         try:
+            # firmware update
+            firmware = Firmware(
+                self.standby_slot_mount_point
+                / Path(cfg.FIRMWARE_CONFIG).relative_to("/")
+            )
+            firmware.update(self._cboot_control.get_standby_slot() == "0")
```
Bodong-Yang, line 494: This line is kind of ambiguous, what about directly passing the slot number(0 or 1) into the update function?

otaclient/app/boot_control/firmware.py
```diff
@@ -0,0 +1,78 @@
...
+    """
+
+    def __init__(self, config_file: Path):
+        self._config_file: Path = config_file
+
+    def update(self, slot_a: bool):
+        if not self._config_file.is_file():
+            # just emit warning for backward compatibility.
+            logger.warning(f"{self._config_file} doesn't exist")
+            return
```
Bodong-Yang, line 58: I think not updating firmware is the normal case(we only update firmware when BSP version migrates), so I think in contrast we should issue warning when we are going to update firmware.

otaclient/app/boot_control/firmware.py
```diff
@@ -0,0 +1,78 @@
...
+            if len(fw["partitions"]) != 2:
+                raise ValueError(f"illegal partitions length: {len(fw['partitions'])=}")
+            self._extract_and_copy(fw, slot_a)
+
+    def _extract_and_copy(self, firmware: Dict, slot_a: bool):
+        parent = self._config_file.parent
+        dctx = zstandard.ZstdDecompressor()
+        ifile = parent / firmware["file"]
+        ofile = firmware["partitions"][0 if slot_a else 1]
+        self._copy(ifile, ofile, dctx.copy_stream)
```
Bodong-Yang, line 73: What about change the input paras to `self, firmware: Dict, slot: int`, and using the input slot number in this function.

otaclient/app/boot_control/firmware.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def _copy(self, ifile: Path, ofile: Path, copy_func: Callable):
+        with open(ifile, "rb") as ifh, open(ofile, "wb") as ofh:
+            logger.info(f"{ifile=}, {ofile=}")
+            copy_func(ifh, ofh)
```
Bodong-Yang, line 78: I think the `_copy` method can be merged into `_extract_and_copy` method?

Bodong-Yang (commented):
@obi-t4 I left some comments for the current implement, please have a check, thank you!

otaclient/app/boot_control/firmware.py
obi-t4: firmware updating is used only for the cboot and only cboot calls this update function, so I think it is natural to emit warning when config(=yaml) doesn't exist.

otaclient/app/boot_control/firmware.py
obi-t4: This is for unit test purpose to mock.

otaclient/app/boot_control/_cboot.py
```diff
@@ -260,6 +262,7 @@ def switch_boot(self):
         slot = self.standby_slot
 
         logger.info(f"switch boot to {slot=}")
+        Nvbootctrl.set_active_boot_slot(slot, target="bootloader")
```
obi-t4, line 265: nvbootctrl with bootloader support

Bodong-Yang: So if we specify `nvbootctrl -t bootloader`, and select the next boot slot to slot a(0), the device will not only boots with rootfs in slot a(0), then it will also use bootloader and firmwares for slot a, is that correct?

Bodong-Yang: So if we specify `nvbootctrl -t bootloader`, and select the next boot slot to slot a(0), the device will not only boots with rootfs in slot a(0), then it will also use bootloader and firmwares for slot a, is that correct?

Bodong-Yang: So if we specify `nvbootctrl -t bootloader`, and select the next boot slot to slot a(0), the device will not only boots with rootfs in slot a(0), then it will also use bootloader and firmwares for slot a, is that correct?

obi-t4: target rootfs and bootloader seems to be independent.
target=rootfs is for rootfs, and target=bootloader is for firmware.
If we specify rootfs active slot=0 and bootloader active slot=0, both boot from slot 0.

Bodong-Yang (dismissed):
Thank you for the testing and implementing!

obi-t4:
@Bodong-Yang code has been fixed and confirmed it works properly.

Bodong-Yang (approved):

PR #207, opened 2023-03-14
[Refactor] re-implement RetryTaskMap

Bodong-Yang:
## Introduction
Previous implementation has a problem, it will hold all the resources related to failed task, even that retry round is finished, due to unexpected reference cycle.
This PR re-implements the `RetryTaskMap` to resolve the following issues & requirements:
- simplify the internal design and implementation.
- better memory efficiency, not consume all input iterable at first retry round.
- resolve memory leak due to reference cycle.
- better error handling on interruption/shutdown.
- caller has more control over the executing tasks.

NOTE: this PR is based on #205 .

## Example
```python
        with ThreadPoolExecutor() as pool:
            _mapper = RetryTaskMap(
                max_concurrent=<max_conccurent_tasks>,
                executor=pool,
                retry_interval_f=<Callable[[int], Any]>,
                max_retry=<max_retry>, 
            )
           # a tuple of bool indicates the task finished without exception or not, 
           # processed entry and Future of processed task will be returned.
            for retry_round, task_result in _mapper.map(
                <func>, <iterable>
            ):
                # task successfully finished
                _is_successful, _entry, _fut = task_result
                if _is_successful:
                    ... # logic on successful finished task
                else:
                    # logic on failed task
                    _fut: Future
                    _entry: Any
                    ...

                if <interrupt_condition>:
                    _mapper.shutdown() # force interrupt the running RetryTaskMap instance and raise RetryTaskMapInterrupted
```

## TODO

- [x] test files passed
- [x] confirm improvement comparing to previous implementation
- [x] VM test

### DIFF
--- a/otaclient/app/create_standby/rebuild_mode.py
+++ b/otaclient/app/create_standby/rebuild_mode.py
@@ -123,13 +123,14 @@ class RebuildMode(StandbySlotCreatorProtocol):
                 ),
                 executor=pool,
             )
-            for _is_successful, _entry, _fut in _mapper.map(
+            for _, task_result in _mapper.map(
                 self._process_regular,
                 self.delta_bundle.new_delta.items(),
             ):
-                if not _is_successful:
+                is_successful, entry, fut = task_result
+                if not is_successful:
                     logger.error(
-                        f"[process_regular] failed to process {_entry=}: {_fut=}"
+                        f"[process_regular] failed to process {entry=}: {fut=}"
                     )
             self.stats_collector.wait_staging()
 
--- a/otaclient/app/create_standby/rebuild_mode.py
+++ b/otaclient/app/create_standby/rebuild_mode.py
@@ -17,10 +17,11 @@ import os
 import shutil
 import time
 from concurrent.futures import ThreadPoolExecutor
+from functools import partial
 from pathlib import Path
 from typing import List, Set, Tuple
 
-from ..common import RetryTaskMap
+from ..common import RetryTaskMap, wait_with_backoff
 from ..configs import config as cfg
 from ..ota_metadata import MetafilesV1, OTAMetadata
 from ..update_stats import (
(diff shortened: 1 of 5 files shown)


otaclient/app/create_standby/rebuild_mode.py
```diff
@@ -112,21 +113,24 @@ def _process_regulars(self):
...
+            for _is_successful, _entry, _fut in _mapper.map(
+                self._process_regular,
+                self.delta_bundle.new_delta.items(),
+            ):
+                if not _is_successful:
                     logger.error(
-                        f"[process_regular] failed to process {_entry=}: {_exp=}"
+                        f"[process_regular] failed to process {_entry=}: {_fut=}"
                     )
-                    raise _exp
```
Bodong-Yang, line 129: This raise is not intended, we only break out when max_retry is reached, and let the RetryTaskMap instance raise the last exception.

Bodong-Yang:
A simple POC to check whether there is memory leak caused by RetryTaskMap:
```python
import gc
import time
from concurrent.futures import ThreadPoolExecutor
from .common import RetryTaskMap

if __name__ == "__main__":
    a_large_object = [a for a in range(12345678)]

    def _always_failed_task(_):
        b = a_large_object  # reference a large object
        raise ValueError  # local variable b will be recorded in the traceback frame

    with ThreadPoolExecutor() as pool:
        _mapper = RetryTaskMap(
            max_concurrent=128,
            executor=pool,
            retry_interval_f=lambda _: time.sleep(0.1),
            max_retry=6,
        )
        for _ in _mapper.map(_always_failed_task, range(2000)):
            pass

    print("done!")
    # the result should be 1, otherwise it means RetryTaskMap still holds reference
    # to a_large_object.
    gc.get_referrers(a_large_object)
    # after del, if no memory leak(above statement returns 1), the memory should be
    # release immediately, otherwise memory leak confirmed.
    del a_large_object
```

The previous implementation cannot passed the above test, the `gc.get_referrers` is not 1, which indicates that the failed tasks Futures is still held by `RetryTaskMap`.
The implementation in this PR passes the test.

Bodong-Yang:
The performance is also improved with this PR, under the same condition, the update time cost reduced from 1000+ to 980~ seconds, memory usage also reduced a little(~10M). 
Most importantly, after each OTA update retry, more memory released comparing to previous implementation.

Test configuration(image condition):
```
    total_files_size_uncompressed: 23637537004
    total_files_num: 294680
    total_download_files_num: 184948
    total_download_files_size: 20703405289
    total_remove_files_num: 25721
    processed_files_num: 294680
    processed_files_size: 23637537004
```

obi-t4 (approved):
LGTM :+1:

PR #208, opened 2023-03-23
[Fix&refine] new way to collect downloader stats, resolve edge condition on downloading breakout condition

Bodong-Yang:
## Introduction
Previously, the downloading retry strategy(implemented by **RetryTaskMap** with **keep_failing_timer**) used by otaclient breaks out when **no more succeeded files keeps longer than <timeout_setting>**, there is a potential edge condition that all downloading threads are stuck by downloading large files, resulting in unwanted pre-mature breakout.

This PR fixes this edge condition by introducing another breakout condition, **no more traffic in downloader/downloader becomes inactive longer than \<limit\>**, and change the breakout logic.

## Inactive breakout logic
The download tasks group(outer for loop) has `last_active_timestamp`, which indicates the latest unix timestamp for latest activity(either a successful task, or traffic by downloader). 
If the download tasks group stalls(becomes inactive) longer than \<limit\>, the download tasks group will breakout.
`last_active_timestamp` will be updated by either a succeeded task, or downloader's `last_active_timestamp`, and the newer value will be used.

## Changes
1. **downloader**: now has 3 new properties: `downloaded_bytes`, `downloader_active_seconds` and `last_active_stamp`, these 3 properties are updated per second, and the stats updating starts from the downloader being initialized.
2. **otaclient status API** now report `downloaded_bytes(download_bytes in v1)`, `downloading_elapsed_time(elapsed_time_download in v1)` by accessing the corresponding properties of downloader.
3. **otaclient.download_files & ota_metadata._process_text_base_otameta_file** now have new breakout logic on exceeded inactive duration.
4. **update_stats**: do not update `downloaded_bytes` and `downloading_elapsed_time` anymore, these fields are updated by otaclient status API handler now.

## Other changes
1. fix/update test files accordingly.
2. configs: change `DOWNLOAD_GROUP_BACKOFF_FACTOR` from 0.1 to 1, each retry round should have longer initial interval.
3. configs: rename `DOWNLOAD_GROUP_NO_SUCCESS_RETRY_TIMEOUT` -> `DOWNLOAD_GROUP_INACTIVE_TIMEOUT`.

## Ticket
fix: [[OTA] downloader/RetryTaskMap: edge condition related to 5mins keep_failing breakout](https://tier4.atlassian.net/browse/T4PB-27222)

### DIFF
--- a/otaclient/app/ota_client.py
+++ b/otaclient/app/ota_client.py
@@ -199,7 +199,6 @@ class _OTAUpdater:
             cur_stat.size = _local_copy.stat().st_size
             return cur_stat
 
-        keep_failing_timer = time.time()
         with ThreadPoolExecutor(thread_name_prefix="downloading") as _executor:
             _mapper = RetryTaskMap(
                 title="downloading_ota_files",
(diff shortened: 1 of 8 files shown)


otaclient/app/ota_client.py
```diff
@@ -219,31 +213,32 @@ def _download_file(entry: RegularInf) -> RegInfProcessedStats:
                 executor=_executor,
             )
             for _, task_result in _mapper.map(_download_file, download_list):
-                # task successfully finished
                 is_successful, entry, fut = task_result
+                # reset keep_failing timer on succeeded task or
+                # downloader is active during the last <POLL_INTERVAL> period.
+                if is_successful or self._downloader.is_downloader_active:
```
obi-t4, line 219: I think `self._downloader.is_downloader_active` is not appropriate condition. Imagine if number of thread is 1.
keep_failing_timer should be reset if at least one byte can be donwloaded with the `download_file` function.

obi-t4 (commented):
I left a comment, please have a check.

otaclient/app/ota_client.py
Bodong-Yang: Yes you are right! Let me reconsider a better condition.

Bodong-Yang: Updated in [698e01a](https://github.com/tier4/ota-client/pull/208/commits/698e01abae42ec4afe42b9d3395ba3fcbfed396b).

Bodong-Yang:
@obi-t4 In [698e01a](https://github.com/tier4/ota-client/pull/208/commits/698e01abae42ec4afe42b9d3395ba3fcbfed396b), the breakout condition is updated to **downloader keeps stuck/inactive longer than <limit>**.

More precisely, in the for loop:
1. if yielded task succeeded, continue,
2. if yielded task failed, check the downloader's `last_active_timestamp`, if the duration is longer than <limit>, shutdown the map and breakout raising last exception.

obi-t4 (commented):
sequence seems to be OK, I'm testing now.

obi-t4 (approved):
LGTM :+1:

PR #209, opened 2023-03-27
fix: fix proto version

obi-t4:
This PR fix .proto_hash and .version manually.


Bodong-Yang (approved):
Thanks for the quick fix!

PR #210, opened 2023-03-28
fix: status phase mapping

obi-t4:
I'd like to change status v1 and v2 phase mapping:

[before chaging]

https://user-images.githubusercontent.com/58019445/228157738-8e6647e3-c018-4203-a9f4-e5c890e2c5c3.mp4

After changing phase is more natual.


Bodong-Yang (approved):
Thank you for the fix! Calculating delta phase including the preparation of local copy, so it should be mapped to v1's REGULAR.

PR #213, opened 2023-04-06
[Fix] fix downloader last_active_timestamp recording

Bodong-Yang:
The downloading thread is marked as active before the connection is made, which actually make the active flag meaningless. 
This PR fixes the issue by only set the downloading thread active flag after the connection is established.


obi-t4 (approved):
LGTM :+1:

PR #216, opened 2023-04-20
[Fix] downloader: fix potential download_bytes incorrect record

Bodong-Yang:
## Introduction
Starts from introducing compression support for OTA image, `download_bytes` record when downloading compressed OTA file is done by querying raw response(which is `urllib3.response.HTTPResponse`) `tell` method. However, there is report that indicates `download_bytes` record is not correct in some case. 
This PR introduces another mechanism, which collects the `download_bytes` for each downloading task by our code instead of using raw response' `tell` method.

## Mechanism
A new `HTTPResponseWrapper` is introduced, which wrap around the raw response `read` method. This method collects the actual number of bytes read from the underlying raw response object. We use this wrapper to get the actual bytes read from raw response instead of using raw response's tell method.

## Ticket
https://tier4.atlassian.net/browse/RT4-4198

### DIFF
--- a/otaclient/app/downloader.py
+++ b/otaclient/app/downloader.py
@@ -29,21 +29,17 @@ from hashlib import sha256
 from os import PathLike
 from pathlib import Path
 from typing import (
-    IO,
     Any,
-    ByteString,
     Callable,
     Dict,
     Iterator,
     Optional,
     Protocol,
     Tuple,
-    Union,
 )
 from requests.adapters import HTTPAdapter
 from urllib.parse import urlsplit
 from urllib3.util.retry import Retry
-from urllib3.response import HTTPResponse
 
 from .configs import config as cfg
 from .common import OTAFileCacheControl, wait_with_backoff
--- a/otaclient/app/downloader.py
+++ b/otaclient/app/downloader.py
@@ -386,6 +386,11 @@ class Downloader:
                             _traffic_on_wire - traffic_on_wire
                         )
                         traffic_on_wire = _traffic_on_wire
+                    # double check the wrapper's bytes_read in case we left anything
+                    self._traffic_report_que.put_nowait(
+                        wrapped_resp.tell() - traffic_on_wire
+                    )
+                    traffic_on_wire = wrapped_resp.tell()
                 # un-compressed file
                 else:
                     for data_chunk in resp.iter_content(chunk_size=self.CHUNK_SIZE):
--- a/otaclient/app/downloader.py
+++ b/otaclient/app/downloader.py
@@ -144,8 +144,11 @@ class DecompressionAdapterProtocol(Protocol):
     """DecompressionAdapter protocol for Downloader."""
 
     @abstractmethod
-    def iter_chunk(self, src_stream: Union[IO[bytes], ByteString]) -> Iterator[bytes]:
+    def iter_chunk(self, src_stream) -> Iterator[bytes]:
         """Decompresses the source stream.
+
+        Params:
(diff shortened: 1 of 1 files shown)


otaclient/app/downloader.py
```diff
@@ -158,10 +161,25 @@ class ZstdDecompressionAdapter(DecompressionAdapterProtocol):
...
 
 
+class _HTTPResponseWrapper:
+    def __init__(self, resp: urllib3.HTTPResponse) -> None:
+        self._resp = resp
+        self._bytes_read = 0
+        self.retry_count = len(resp.retries.history) if resp.retries else 0
+
+    def read(self, size: int) -> bytes:
+        data = self._resp.read(size)
```
obi-t4, line 171: @Bodong-Yang Let me confirm, if the file size is 2GB, 2GB is loaded onto the main memory?

obi-t4: Ah, sorry, this just read chunk size, not whole size.

Bodong-Yang: This wrapper is only used by decompressor's `iter_chunk` method, and decompressor will only read a chunk of data each time from this wrapper. (The read size is `zstandard.DECOMPRESSION_RECOMMENDED_INPUT_SIZE`, the default recommended value of zstandard's `read_to_iter` stream decompressing method).

Bodong-Yang: Basically this wrapper is a stub for raw response, wraps the raw response's read method, tracking the actual number of bytes we read from the raw response.

otaclient/app/downloader.py
```diff
@@ -337,12 +346,11 @@ def _download_task(
         #       the number will be of the decompressed file)
         _hash_inst, downloaded_file_size = self._hash_func(), 0
         # NOTE: traffic_on_wire is the number of bytes we directly downloaded from remote
-        traffic_on_wire = 0
-        _err_count = 0
+        traffic_on_wire, err_count = 0, 0
```
obi-t4, line 349: Sometimes you change variable name with `_` or w/o `_`.
Please revert this change if you don't have rule.
If there is no explicit rule, this variable name will be modified in the next PR.

otaclient/app/downloader.py
```diff
@@ -354,29 +362,33 @@ def _download_task(
...
+                wrapped_resp = _HTTPResponseWrapper(resp.raw)
+                err_count = wrapped_resp.retry_count
 
                 # support for compresed file
                 if decompressor := self._get_decompressor(compression_alg):
-                    for _chunk in decompressor.iter_chunk(resp.raw):
-                        _hash_inst.update(_chunk)
-                        _dst.write(_chunk)
-                        downloaded_file_size += len(_chunk)
+                    for data_chunk in decompressor.iter_chunk(wrapped_resp):
```
obi-t4, line 370: _chunk -> data_chunk change is required? If not required, please revert it.

obi-t4 (commented):
I left a comment, but code itself seems to be OK.

Bodong-Yang:
@obi-t4 I still cannot confirm/reproduce the issue on my local setup, also cannot confirm that raw response's `tell` method will return actual data but record 0 on bytes read when reading the `urllib3.response.HTTPResponse` source code.  (`HTTPResponse` does the same thing as we do actually, when reading data from further lower level(socket), it will record the data's len first, and then return the data, these logics are implemented together in the `read` method.)

So I will turn this PR into draft, until I can confirm the issue and its cause.

PR #217, opened 2023-04-26
[Fix] ecu_info: not loading default config when ecu_info.yaml is invalid in some cases

Bodong-Yang:
## Introduction
This PR fixes `ECUInfo` doesn't load default config when `ecu_info.yaml` is valid yaml but not valid ecu_info. 
This is caused by assignment before assertion statement.

NOTE: this PR is splited from #212. 

## Changes
1. fix not loading default config when `ecu_info.yaml` is a valid yaml file but not valid ecu_info(yaml array, empty file, etc.),
2. refactor test files, cover the cases mention above.

### DIFF
--- a/otaclient/app/ecu_info.py
+++ b/otaclient/app/ecu_info.py
@@ -61,14 +61,19 @@ class ECUInfo:
     def parse_ecu_info(cls, ecu_info_file: Union[str, Path]) -> "ECUInfo":
         ecu_info = deepcopy(DEFAULT_ECU_INFO)
         try:
-            _ecu_info = yaml.safe_load(Path(ecu_info_file).read_text())
-            assert isinstance(ecu_info, Dict)
+            ecu_info_yaml = Path(ecu_info_file).read_text()
+            _ecu_info = yaml.safe_load(ecu_info_yaml)
+            assert isinstance(_ecu_info, Dict)
             ecu_info = _ecu_info
+        except (yaml.error.MarkedYAMLError, AssertionError) as e:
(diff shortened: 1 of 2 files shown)


otaclient/app/ecu_info.py
```diff
@@ -61,13 +61,14 @@ class ECUInfo:
     def parse_ecu_info(cls, ecu_info_file: Union[str, Path]) -> "ECUInfo":
         ecu_info = deepcopy(DEFAULT_ECU_INFO)
         try:
-            ecu_info = yaml.safe_load(Path(ecu_info_file).read_text())
+            _ecu_info = yaml.safe_load(Path(ecu_info_file).read_text())
             assert isinstance(ecu_info, Dict)
```
obi-t4, line 65: asserting `ecu_info` at this line is strange. Can you add test for this PR?

obi-t4 (commented):
I left a comment.

otaclient/app/ecu_info.py
Bodong-Yang: Sure! This line is to prevent the case that ecu_info.yaml is valid yaml but not valid ecu_info file(like top level is array, empty file, etc.).

Bodong-Yang:
@obi-t4 The new tests file that covers the new error cases is ready! 
Also actually the previous commit doesn't fully fix the problem, so I also refix it again, and let the error logging provides more details. Now the latest commit should work.

obi-t4 (approved):

PR #211, opened 2023-04-04
[Fix] delta calculating: verify file when copying from active slot to standby slot

Bodong-Yang:
## Introduction
Previously, in `DELTA_CALCULATING` OTA stage, the local copy is prepared after the file in active slot is verified, there is chance that the file is updated after the hash verification. This PR fixes this edge condition by verifying file during the copy. 

## Other change
1. create_standby: when preparing local copy, also take size information into account if the information is available(by saved active slot's OTA image meta).
1. otaclient: fix not resetting `last_failure_traceback` when new update/rollback request is received.
1. common: add a helper method `create_tmp_fname`.

## Ticket
[T4PB-27531](https://tier4.atlassian.net/browse/T4PB-27531)

[T4PB-27531]: https://tier4.atlassian.net/browse/T4PB-27531?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #218, opened 2023-04-26
[Refinement] refinement related to otaproxy launching

Bodong-Yang:
## Description
This PR introduces exposed helper functions for external launching otaproxy, supports configuring `cache_dir` and `cache_db_file` location for standalone otaproxy CLI.

### Motivation
1. Previously the standalone otaproxy startup cli (\_\_main__.py) **doesn't support configure cache_dir and cache_db_file location**, make the launching and testing standalone otaproxy harder as we have to modify the `otaproxy/config.py` every-time. 
2. Previously a wrapper for launching otaproxy in subprocess is implemented `ota_client_stubs`. This wrapper does the job of constructing otaproxy components, configuring ota_cache and otaproxy uvicorn server and laucnhing it in subprocess. **This reduces `ota_client_stub` module's cohesion and introduce deep coupling with otaproxy package, which is not ideal.**

### NOTE
The helper functions introduced by this PR is not yet used in main branch, it is used in refactored `ota_client_stub.py` in #212 . 

## Check list

- [x] local test is passed. 

## Changes
at `otaclient.ota_proxy` package:
1.  `__init__.py`: add `run_otaproxy` helper function for setting up and starting otaproxy server, add wrapper function `subprocess_start_otaproxy` helper function for starting otaproxy as subprocess in one call.
4. `__main__.py`: add `--cache-dir` and `--cache-db-file` options for CLI interface. 

## Behavior changes

- [x] No.

## Breaking change

- [ ] Yes
- [x] No

### DIFF
--- a/otaclient/ota_proxy/__init__.py
+++ b/otaclient/ota_proxy/__init__.py
@@ -98,7 +98,7 @@ def subprocess_start_otaproxy(*args, **kwargs) -> SpawnProcess:
     mp_ctx = multiprocessing.get_context("spawn")
     otaproxy_subprocess = mp_ctx.Process(
         target=partial(_subprocess_main, *args, **kwargs),
-        daemon=True,  # kill otaproxy if otaclient exists
+        daemon=True,  # kill otaproxy if the parent process exists
     )
     otaproxy_subprocess.start()
     return otaproxy_subprocess
--- a/otaclient/ota_proxy/__init__.py
+++ b/otaclient/ota_proxy/__init__.py
@@ -13,9 +13,127 @@
 # limitations under the License.
 
 
+import asyncio
+import logging
+import multiprocessing
+from functools import partial
+from multiprocessing.context import SpawnProcess
+from pathlib import Path
+from typing import Callable
+
 from .cache_control import OTAFileCacheControl
 from .server_app import App
 from .ota_cache import OTACache, OTACacheScrubHelper
 from .config import config
 
-__all__ = ("App", "OTACache", "OTACacheScrubHelper", "OTAFileCacheControl", "config")
+logger = logging.getLogger(__name__)
+
+
+__all__ = (
+    "App",
+    "OTACache",
+    "OTACacheScrubHelper",
+    "OTAFileCacheControl",
+    "config",
+    "subprocess_start_otaproxy",
+)
+
+
+def _subprocess_main(
+    host: str,
+    port: int,
+    *,
+    init_cache: bool,
+    cache_dir: str,
+    cache_db_f: str,
+    upper_proxy: str,
+    enable_cache: bool,
+    enable_https: bool,
+    subprocess_init: Callable,
+):
+    import uvloop
+
+    # ------ pre-start callable ------ #
+    subprocess_init()
+    # ------ scrub cache folder if cache re-use is possible ------ #
+    should_init_cache = (
+        init_cache or not Path(cache_dir).is_dir() or not Path(cache_db_f).is_file()
+    )
+
+    if not should_init_cache:
+        scrub_helper = OTACacheScrubHelper(cache_db_f, cache_dir)
+        try:
+            scrub_helper.scrub_cache()
+        except Exception as e:
+            logger.error(f"scrub cache failed, force init: {e!r}")
+            should_init_cache = True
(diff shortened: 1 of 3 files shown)


obi-t4:
@Bodong-Yang 
Could you explain why this PR is necessary? Or we will reject this PR.

Bodong-Yang:
@obi-t4 (I also add the following lines into the PR comments)
1. Previously the standalone otaproxy startup cli (\_\_main__.py) doesn't support configure `cache_dir` and `cache_db_file` location, make the launching and testing standalone otaproxy harder as we have to modify the `config.py` everytime. so in this PR these options are provided as cli params.
2. **As a standalone python package, otaproxy package should expose helper functions for the user to launch otaproxy without knowing the implementation details(abstraction).** Previously, `ota_client_stub` launch otaproxy by directly importing internal components like `App` and `OTACache` classes from otaproxy, construct them into server instance, and launch it by `ota_client_stub` itself, **this reduce the cohesion** of `ota_client_stub`(it should only focusing on otaclient service API handling), **increase the coupling level with otaproxy package**(it constructs otaproxy server instance by itself), which is not ideal. This PR introduces some helper functions that user can use to launch otaproxy, without the need to let user compose otaproxy server instance by itself.

obi-t4:
@Bodong-Yang I could understand the motivation and I'm thinking of introducing PR template.
I think we need motivation field for refinement PR.
I'd like to introduce the paragraph on the motivation section like:
- current behavior
- the problem about current behavior
- how to improve/fix the problem by this change

I can understand your motivation description, but to understand it more easily when we read this PR back later.

otaclient/ota_proxy/__init__.py
```diff
@@ -13,9 +13,127 @@
...
+    to the _subprocess_main function, and then execute the function in
+    a subprocess.
+    check _subprocess_main function for more details.
+    """
+
+    # run otaproxy in async loop in new subprocess
+    mp_ctx = multiprocessing.get_context("spawn")
+    otaproxy_subprocess = mp_ctx.Process(
+        target=partial(_subprocess_main, *args, **kwargs),
+        daemon=True,  # kill otaproxy if otaclient exists
```
obi-t4, line 101: This function doesn't care about parent process is otaclient or not.
```suggestion
        daemon=True,  # kill otaproxy if the parent process exists
```

obi-t4:
@Bodong-Yang 
> this reduce the cohesion of ota_client_stub(it should only focusing on otaclient service API handling), increase the coupling level with otaproxy package(it constructs otaproxy server instance by itself), which is not ideal. 

But this PR doesn't fix this issue. Step-by-step PR is OK, so I hope the next PR fix this problem.

Bodong-Yang:
> But this PR doesn't fix this issue. Step-by-step PR is OK, so I hope the next PR fix this problem.

@obi-t4 Ah this is PR is split from #212 , so this PR only provides these functions, and the actual use is in the refactored `ota_client_stub.py` in #212 .

obi-t4:
@Bodong-Yang 
Is it possible to split `the actual use is in the refactored ota_client_stub.py` by another PR?

Bodong-Yang:
> @Bodong-Yang Is it possible to split `the actual use is in the refactored ota_client_stub.py` by another PR?

@obi-t4 It is difficult as the refactored `ota_client_stub.py` in #212  use the helper function `subprocess_start_otaproxy` in the first place(which means it doesn't use the legacy code in the beginning).

Bodong-Yang:
( Also as we are introducing total refactoring of `ota_client_stub` in #212 , it is a meaningless to do modification on main branch's `ota_client_stub.py` I think.

obi-t4:
OK, you need to add the description that the use of `subprocess_start_otaproxy` is done in #211.

obi-t4:
@Bodong-Yang 
Can you rewrite the motivation section as follows:
```
- current behavior
- the problem about current behavior
- how to improve/fix the problem by this change
```

Bodong-Yang:
@obi-t4 The PR description is updated to align the refinement PR template, thank you!

obi-t4 (approved):
LGTM :+1:

obi-t4 (approved):
LGTM :+1:

PR #222, opened 2023-05-11
[Fix & Refinement] ota_metadata: implement validation over metadata.jwt

Bodong-Yang:
## Description

This PR introduces the missing validation mechanism when parsing `metadata.jwt`, explicitly define default value for optional field, or raise exception on missing must_set field in `metadata.jwt`. For case of invalid `metadata.jwt`, new OTA error type for `metadata.jwt` invalid is defined in this PR.

### metadata.jwt format
```
# check the document in https://tier4.atlassian.net/l/cp/PCvwC6qk for more details
Version1 JWT payload layout(revision2):
[
    {"version": 1}, # must
    {"directory": "dirs.txt", "hash": <sha256_hash>}, # must
    {"symboliclink": "symlinks.txt", "hash": <sha256_hash>}, # must
    {"regular": "regulars.txt", "hash": <sha256_hash>}, # must
    {"persistent": "persistents.txt", "hash": <sha256_hash>}, # must
    {"certificate": "sign.pem", "hash": <sha256_hash>}, # must
    {"rootfs_directory": "data"}, # must
    {"total_regular_size": "23637537004"}, # optional, added in revision1, default is `0`
    {"compressed_rootfs_directory": "data.zstd"} # optional: added in revision2, default is `""`
]
```
> **Warning**
> `total_regular_size` in `metadata.jwt` is str, we need to convert it into int.

### validation mechanism
The validation includes the following operations:
1. for **must_set field**, check if any is missing and raise exception on missing,
2. for **optional field**, set proper default value for it if missing,
3. for metafiles field(`directory`, `symbolicklink`, `regular`, `persistent`, `certificate`), check input field value's format.

### error handling
`ota_metadata` module now internally defines `MetadataJWTPayloadInvalid` and `MetadataJWTVerificationFailed`. 
For external OTA errors definition, corresponding OTA recoverable error `MetadataJWTInvalid` and `MetadataJWTVerficationFailed` are defined.
`ota_client` now catches the internal `ota_metadata` exceptions, wraps it into corresponding OTA recoverable errors, and reports them.

## Check list

- [x] test file that covers the bug case(s) is implemented:
    1. default value for missing optional field is correctly set.
    2. exception is raised on invalid `metadata.jwt`.
    3. parsing over valid `metadata.jwt` is OK.
- [x] local test is passed. 
- [x] confirm bug cannot be reproduced with this PR on VM environment. 

## Behavior changes

- [x] Yes, internal behaivor (will not impact user experience). Please check **Bug fix** section below for more details.

### 1. ota_client's status API doesn't respond when `total_regular_size` field is missing in metadata.jwt

Ticket: https://tier4.atlassian.net/browse/RT4-4767

#### Current behavior(cause of the bug)

 `ota_metadata` doesn't define proper default value for missing `total_regular_size`(`None` is used). when ota_client status API constructing response, passing `None` to `total_regular_files_size_uncompressed` field in `StatusResponseECUV2` causing `TypeError`, resulting in status API not responding.

Also, if `metadata.jwt` missing any metafile fields(`directory`, `symbolicklink`, `regular`, `persistent`, `certificate`), it will cause exceptions at where these fields are being called, but the exception will be hard to reveal the real cause(fields missing in `metadata.jwt`).

#### Behaivor after fix

1. for valid `metadata.jwt` that missing optional fields, proper default value will be set after parsing:
    a. `total_regular_size` -> 0,
    b. `compressed_rootfs_directory` -> `""`,
2. for invalid `metadata.jwt`(either invalid json, or missing must_set fields), `MetadataJWTPayloadInvalid` OTA recoverable error will be reported.
3. OTA recoverable error `BaseOTAMetaVerificationFailed` now is renamed to `MetadataJWTVerficationFailed`.

## Related links & ticket

Ticket: https://tier4.atlassian.net/browse/RT4-4767
ota_metadata format: https://tier4.atlassian.net/l/cp/PCvwC6qk

### DIFF
--- a/otaclient/app/errors.py
+++ b/otaclient/app/errors.py
@@ -35,8 +35,9 @@ class OTAErrorCode(Enum):
     E_OTAMETA_VERIFICATION_FAILED = 205
     E_UPDATEDELTA_GENERATION_FAILED = 206
     E_APPLY_OTAUPDATE_FAILED = 207
-    E_BASE_OTAMETA_VERIFICATION_FAILED = 208
+    E_METADATA_JWT_VERIFICATION_FAILED = 208
     E_OTAPROXY_FAILED_TO_START = 209
+    E_INVALID_METADATAJWT = 210
 
     E_OTA_ERR_UNRECOVERABLE = 300
     E_BOOTCONTROL_PLATFORM_UNSUPPORTED = 301
@@ -251,10 +252,16 @@ class ApplyOTAUpdateFailed(OTAErrorRecoverable):
     desc: str = f"{_RECOVERABLE_DEFAULT_DESC}: failed to apply ota update"
 
 
-class BaseOTAMetaVerificationFailed(OTAErrorRecoverable):
+class MetadataJWTVerficationFailed(OTAErrorRecoverable):
     module: OTAModules = OTAModules.API
-    errcode: OTAErrorCode = OTAErrorCode.E_BASE_OTAMETA_VERIFICATION_FAILED
-    desc: str = f"{_RECOVERABLE_DEFAULT_DESC}: verification failed for base otameta"
+    errcode: OTAErrorCode = OTAErrorCode.E_METADATA_JWT_VERIFICATION_FAILED
+    desc: str = f"{_RECOVERABLE_DEFAULT_DESC}: verification failed for metadata.jwt"
+
+
+class MetadataJWTInvalid(OTAErrorRecoverable):
+    module: OTAModules = OTAModules.API
+    errcode: OTAErrorCode = OTAErrorCode.E_INVALID_METADATAJWT
+    desc: str = f"{_RECOVERABLE_DEFAULT_DESC}: invalid metadata.jwt"
 
 
 class OTAProxyFailedToStart(OTAErrorRecoverable):
--- a/otaclient/app/ota_client.py
+++ b/otaclient/app/ota_client.py
@@ -152,7 +152,7 @@ class _OTAUpdater:
         self.updating_version: str = ""
         self.failure_reason = ""
         # init variables needed for update
-        self._otameta: OTAMetadata = None  # type: ignore
+        self._otameta: ota_metadata.OTAMetadata = None  # type: ignore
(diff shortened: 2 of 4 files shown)


Bodong-Yang:
@obi-t4 I notice that `total_regular_size` is string in `metadata.jwt`, I think it should be int? Like `{"total_regular_size": 123456}`. But as it is already widely used, so `ota_metadata` will handle this and do the conversion.

obi-t4:
> @obi-t4 I notice that `total_regular_size` is string in `metadata.jwt`, I think it should be int? Like `{"total_regular_size": 123456}`. But as it is already widely used, so `ota_metadata` will handle this and do the conversion.

currently total_regular_size is saved as string.
https://github.com/tier4/ota-metadata/blob/main/metadata/ota_metadata/metadata_sign.py#L70-L71
I'd like make this behavior as specification.

Bodong-Yang:
@obi-t4

> currently total_regular_size is saved as string.

Any reason to use string type instead of numeric type? I didn't take part in reviewing the the ota_metadata documentation so I would like to confirm the intention of this behavior.

> I'd like make this behavior as specification.

Can you update the documentation at https://tier4.atlassian.net/l/cp/PCvwC6qk related to this issue?
Also `compressed_rootfs_directory` is also not documented.

Bodong-Yang:
> @Bodong-Yang
> 
> > 1. for valid `metadata.jwt` that missing optional fields, proper default value will be set after parsing:
> >    a. `version` -> `<UNKNOWN_VERSION: -1>`,
> 
> version filed must be present, not optional.

Updated in [8b69e70](https://github.com/tier4/ota-client/pull/222/commits/8b69e7096a3ac55cd36a9549f40abeae596b379c), now version field is mandatory.

obi-t4:
> Any reason to use string type instead of numeric type? I didn't take part in reviewing the the ota_metadata documentation so I would like to confirm the intention of this behavior.

I think int is appropriate, but for now string is used in the production.

obi-t4 (approved):
LGTM :+1:

PR #223, opened 2023-05-15
[Chore] update .flake8 and pyproject.toml for linting configuration

Bodong-Yang:
## Description
This PR updates `.flake8` and `pyproject.toml` for linting configuration. 
1. flake8: add comments for why ignores E266, E501 and E203.
2. flake8: use proper glob patterns to only match generated proto files, use `extend-exclude` instead of `exclude` to apply the extra excluded config on top of defaults, not overriding it.
3. black: remove `include` config, use default value, as black will automatically search for `otaclient` package.
4. black: use proper glob patterns to only match generated proto files in `extend-exclucede`.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,11 @@
 [flake8]
 max-line-length = 88
+# ignore the following Errors:
+#   E266(too many leading '#'):
+#       sometimes we use multiple # for separting sections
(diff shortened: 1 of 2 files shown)


obi-t4 (approved):
:+1:

PR #229, opened 2023-06-08
[Fix] rpi_boot: call os.sync on all important files replacement

Bodong-Yang:
## Description

There is report that rpi_boot doesn't replace `config.txt` properly in the 1st reboot after firmware update, causing 2nd reboot still boots back to old slot. I read through the source code and found that, in `common.replace_atomic`, `os.sync` is called only after `shutil.copy`, `os.replace` is not covered by `os.sync`. 
This PR fix the `common.replace_atomic` method, while insert more `os.sync` on important file preparing like linux images moving and after firmware update to ensure the changes are write to the disk.

## Check list

- [x] local test is passed. 

## Bug fix

1. `common.replace_atomic`: now `os.sync` is called AFTER `os.replace`, ensure that replace operation is also written to the disk.

## Other changes

1. rpi_boot: add more loggings when important files are prepared.

## Related links & ticket

[RT4-5266](https://tier4.atlassian.net/browse/RT4-5266)

[RT4-5266]: https://tier4.atlassian.net/browse/RT4-5266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/otaclient/app/boot_control/_rpi_boot.py
+++ b/otaclient/app/boot_control/_rpi_boot.py
@@ -301,8 +301,8 @@ class _RPIBootControl:
                 replace_atomic(self.config_txt_standby_slot, self.tryboot_txt)
                 logger.info(
                     "finalizing boot configuration,"
-                    f"replace {self.config_txt=} with {self.config_txt_active_slot}, "
-                    f"replace {self.tryboot_txt=} with {self.config_txt_standby_slot}"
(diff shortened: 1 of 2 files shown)


obi-t4 (approved):
:+1:

Bodong-Yang:
Still need to further confirm, not going to merge it yet.

PR #230, opened 2023-06-09
[Refinement] create_standby: shuffle the calculated download_list

Bodong-Yang:
## Description

This PR introduces the change that the calculated and generated download_list by `DeltaGenerator` will be shuffled before being used to download the needed OTA files.

### Motivation

Considering the situation that multiple ECUs that run the same image are going to update to the same new image.
It is highly possible that otaclients on these ECUs are downloading the same files at once from time to time during the whole OTA. This will reduce the cache efficiency as almost no cache hit will occur.

Currently otaproxy already supports a feature called **simultaneously cache streaming**, which when multiple request to the same file occurs, otaproxy will only request download from remote server once, and teeing the data chunks to multiple client. 

However this feature of otaproxy doesn't resolve the issue of low cache hit, and cache teeing itself introduces extra overhead(teeing data chunks, managing on-going cache writting/subscription, etc.). I think this feature should only serve as last resort to deal with same file requested problem, and we need to try our best to avoid the same file requested problem in the first place.

## Check list

- [x] local test is passed. 

## Changes

1. `create_standby.common`: `DeltaGenerator.calculate_and_process_delta` API now will shuffle the generated `download_list` before return the whole generated delta.

## Behavior changes

- [x] No.

## Breaking change

- [x] No.

### DIFF
--- a/otaclient/app/create_standby/common.py
+++ b/otaclient/app/create_standby/common.py
@@ -15,6 +15,7 @@
 
 r"""Common used helpers, classes and functions for different bank creating methods."""
 import os
+import random
 import time
 from concurrent.futures import (
     Future,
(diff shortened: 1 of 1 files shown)


obi-t4 (approved):
LGTM :+1:

obi-t4:
@Bodong-Yang 
Did you confirmed that this change improves the performance?

Bodong-Yang:
> @Bodong-Yang Did you confirmed that this change improves the performance?

At least the shuffle doesn't introduce noticeble performance drop, the performance increase depends on real world condition, so it is hard to tell.

PR #233, opened 2023-06-13
[Fix] rpi_boot: clear leftover switching_boot_flag_file from previous interrupted OTA

Bodong-Yang:
## Description

This PR fixes otaclient on rpi device incorrectly skipping firmware update and finalizing switching boot if unwanted leftover switching_boot_flag file from previous interrupted OTA presented.

Also check #234  for how the flag file might be leftover.

### rpi_boot control mechanism introduction

considering we have 2slots, A and B, currently we are in slot A, and going to OTA to slot B:
1. (slot A) apply OTA to slot B,
2. (slot A) OTA updating slot B finished, replace tryboot.txt with config.txt_slot_b , and then try boot to slot B(1st reboot)
3. (slot B) detect we are after 1st reboot(flag file not presented), call update-firmware , replace config.txt with config.txt_slot_b to finalize switching boot, create flag file, and then reboot(2st reboot),
4. (slot B) detect we are after 2nd reboot(flag file presented), delete flag file, finished. 

## Check list

- [x] tested on real rpi device.

## Bug fix

1. [RT4-5316](https://tier4.atlassian.net/browse/RT4-5316): on rpi device, in some conditions the otaclient will skip the finalizing of the boot files, resulting in booting back to old slot.

### Current behavior(cause of the problem)

1. If somehow the flag file from previous interrupted OTA still preseneted, after step 2, otaclient will mistakenly think it should be in step 4, resulting in firmware not updated and switching boot not finalized. At this moment, reboot the device again will result in booting back to old slot. 

### Behaivor after fix

1. on rpi device, at the begging of every OTA update(at pre-update), otaclient will cleanup the flag file unconditionally,
2. on rpi_device, after otaclient finishes startup, if ota_status is not `UPDATING/ROLLBACKING`, otaclient will cleanup the switching boot flag file unconditionally. 

## Related links & ticket

1. [ticket: RT4-5316](https://tier4.atlassian.net/browse/RT4-5316)
2. [investigation report](https://tier4.atlassian.net/l/cp/1C2Wf8Z6)

[RT4-5316]: https://tier4.atlassian.net/browse/RT4-5316?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


obi-t4 (approved):
LGTM :+1:

PR #234, opened 2023-06-14
[Fix] boot_control: reboot subprocess call should also terminate otaclient process, refine common.OTAStatusFilesControl

Bodong-Yang:
## Description

This PR fixes potential bug that rpi_boot might change ota_status to `SUCCESS` before 2nd reboot takes effect, resulting in post-2nd reboot cleanup logic not being executed after 2nd reboot into new slot, resulting in flag file not being deleted. Check [investigation report](https://tier4.atlassian.net/l/cp/1C2Wf8Z6) for more details, and also check #233 for the bug caused by leftover flag file.

## Check list

- [x] test on real rpi device.

## Changes list

1. `common.reboot`: add `sys.exit(0)` after subprocess reboot call to ensure otaclient process terminates after reboot is called. This is expected behavior but not guaranteed previously.
2. `common.OTAStatusFilesControl`: refine `_init_ota_status_files` method while maintain the same behavior, except that if in `UPDATING` but not `is_switching_boot`, OTA will be consider as failed.
3. `rpi_boot.finalize_switching_boot`: although in normal cases it will not be reachable, return False instead of True after reboot is called for 2nd reboot, as a failed reboot should be treated as failure.

## Bug fix

1. on rpi device, if reboot doesn't take effect fast enough, rpi_boot might change ota_status to `SUCCESS` before 2nd reboot kill otaclient process, resulting in post-2nd reboot cleanup logic not being executed after 2nd reboot into new slot, resulting in flag file not being deleted.

## Behavior after fix

1. now `CMDHelper.reboot` function will also call `sys.exit(0)` after subprocess reboot call to ensure otaclient process terminates at the point of `CMDHelper.reboot`.

## Related links & ticket

1. [ticket: RT4-5316](https://tier4.atlassian.net/browse/RT4-5316)
4. [investigation report](https://tier4.atlassian.net/l/cp/1C2Wf8Z6)


otaclient/app/boot_control/_common.py
```diff
@@ -454,45 +470,49 @@ def _init_ota_status_files(self):
...
-        else:
-            self._ota_status = _loaded_ota_status
-            # detect failed reboot, but only print error logging currently
-            if _loaded_slot_in_use != self.active_slot:
                 logger.error(
-                    f"boot into old slot {self.active_slot}, "
-                    f"but slot_in_use indicates it should boot into {_loaded_slot_in_use}, "
-                    "this might indicate a failed finalization at first reboot after update/rollback"
+                    "finalization failed on first reboot during switching boot"
                 )
```
obi-t4, line 501: This line
```
raise BootControlInitError(_err_msg)
```
is removed, intentional?

Bodong-Yang: Yes, it is intentional, previously if boot init failed, the otaclient will just exit. With this change, it allows otaclient to keep alive and open the grpc server to report failure status.

Bodong-Yang: ( And switch boot failed is not fatal error(comparing to boot control inits failed due to A/B partition layout incorrectly, etc), it should be counted as OTA failure, and reported via status API, instead of just crashing the otaclient I think.

obi-t4 (approved):
LGTM :+1:

PR #236, opened 2023-06-16
[Refinement] otaproxy: remove cache scrubbing, instead only checking over cache db file

Bodong-Yang:
## Description

> **Note**
> This PR is split from #225 . 

This PR removes the ota_cache scrubbing feature from otaproxy subprocess init, instead it introduces the mechanism of checking over the cache db file, and improve handling the case of cache db corrupted.

### Motivation for removing cache scrubbing on startup

Previously, if there are a lot of left-over cache files(from last failed OTA), otaproxy will take a long time to start due to it will scan and validate all cache files against database records(scrubbing) before startup. This will result in blocking other ECUs that requires this otaproxy for a long time.
The original idea is that, it might be damaged files when last OTA failed. But actually the possibility of damaged cache files is low as cache files will only be registered AFTER the caching is finished, interrupted caching will not be registered in cache database.
Moreover, we already have ota_cache control headers, mechanism and otaclient side file validation, otaclient will report to otaproxy if the cache file is invalid, and indicate otaproxy to re-download the file again.
So in conclusion, the benefit of cache scrubbing before otaproxy starts it very limited, comparing to the overhead/delay it introduces, so it is ideal to remove this feature.

## Check list

- [x] test file(s) that covers the change(s) is implemented.
- [x] local test is passed. 
- [x] test on single VM passed.

## Changes

1. `otaproxy._subprocess_main`: remove cache scrub.
2. `otaproxy.ota_cache`: remove `OTACacheSrubHelper` and related code, 
3. `otaproxy.db`: use `time.time()` instead of `datetime.now().timestamp()` to query the current UNIX timestamp,
4. `otaproxy.db`: `OTACacheDB` implements 2 new helper methods, `check_db_file` and `init_db_file`, especially, `check_db_file` will check the db files' integrity,
5. `otaproxy.db`: now `db` module doesn't take care of initializing database file if database file is invalid, the caller of `OTACacheDB` should use newly implemented helper methods `check_db_file` and `init_db_file` to check and init db file, raise exception on invalid db file,
6. `otaproxy.ota_cache.OTACache`: implement `_check_cache_db` helper method,  db file checking and initializing on invalid db file is fully handled by `OTACache.start` API,
7. update test files accordingly.

## Behavior changes

Does this PR introduce behavior change(s)?

- [x] Yes, internal behaivor (will not impact user experience).
- [x] Yes, external behavior (user will notice the change). 

### Previous behavior

[...]

PR #225, opened 2023-05-17
[Refinement] otaproxy: minor refinements and fixes 

Bodong-Yang:
## Description

> **Note**
> This PR contains the changes from #236, it should be merged and reviewed after #236 .

This PR fixes typing in the otaproxy module, applies refinements related to `NULL` type in orm, and refine the main entry method `retrieve_file` of `OTACache`.

### Motivation for refinement

Previously, there are some minor leftover thing not yet being done/fix for otaproxy:
1. typing is not corrected annotated in some of the modules,
2. orm.py: doesn't have a proper `NULL` type defined(just alias None type), resulting in special case handling and extra checks/code,
3. ota_cache.py: the main entry, `retrive_file` API method is too complicated and has high code cognitive complexity.

## Check list

- [x] test file(s) that covers the change(s) is implemented.
- [x] local test is passed. 
- [x] test on single VM passed.

## Changes

3. `otaproxy.orm`: implement real `NULL_TYPE` instead of type alias, `NULL_TYPE` type is singleton and returns `None` on initializing, and takes any params when init, behaviors like other data type like `int` or `str`, 
6. `otaproxy.orm`: refinement according to new `NULL_TYPE` implementaion, remove some of the special treatments for previous type alias based `NULL_TYPE` as new `NULL_TYPE` behaves like other built-in data types,
7. `otaproxy.orm`: add more inline comments,
8. `otaproxy.orm`: check field value type before field assignment, if the value type is the same as field type, do not go through type conversion,
9. `otaproxy.ota_cache`: split `_FileDescriptorHelper` into several standalone helper functions, 
10. `otaproxy.ota_cache`: refactor `retrive_file` API method, split different cases handler into standalone helper methods, improve code readability and reduce code cognitive complexity,
11. `otaproxy`: now naming the tmp file predictably via URL, check `otaproxy.utils.TmpCacheFileNaming` for more details,
15. update test files accordingly.

## Behavior changes

Does this PR introduce behavior change(s)?

- [x] Yes, internal behaivor (will not impact user experience).
- [x] Yes, external behavior (user will notice the change). 

### Previous behavior

1. (internal behavior) tmp files for unfinished entry is random string,

### Behavior with this PR

2. (internal behavior) now naming the tmp file predictably via URL, check `otaproxy.utils.TmpCacheFileNaming` for more details.

## Breaking change

Does this PR introduce breaking change?
- [x] No.

PR #237, opened 2023-06-19
[Refinement] otaproxy: minor refinements and fixes 

Bodong-Yang:
## Description

> **Note**
> This PR is based on #236 and expected to be merged into main after #236 is merged.

> **Note**
> This PR superseded #225 .

> **Note**
> If possible, this PR should be included in the next releasing, but if it can not catch the releasing window it is OK to delayed the merge.

This PR fixes typing in the otaproxy module, applies refinements related to `NULL` type in orm, and refine the main entry method `retrieve_file` of `OTACache`.

### Motivation for refinement

Previously, there are some minor leftover thing not yet being done/fix for otaproxy:
1. typing is not corrected annotated in some of the modules,
2. orm.py: doesn't have a proper `NULL` type defined(just alias None type), resulting in special case handling and extra checks/code,
3. ota_cache.py: the main entry, `retrive_file` API method is too complicated and has high code cognitive complexity.

## Check list

- [x] local test is passed. 
- [x] test on single VM passed.

## Changes

1. `otaproxy.orm`: implement real `NULL_TYPE` instead of type alias, `NULL_TYPE` type is singleton and returns `None` on initializing, and takes any params when init, behaviors like other data type like `int` or `str`, 
2. `otaproxy.orm`: refinement according to new `NULL_TYPE` implementaion, remove some of the special treatments for previous type alias based `NULL_TYPE` as new `NULL_TYPE` behaves like other built-in data types,
3. `otaproxy.orm`: decorate `ORMeta` with `dataclass_transfomer` to let the IDE identify `ORMBase` subclass as dataclass,
4. `otaproxy.orm`: add more inline comments,
5. `otaproxy.orm`: check field value type before field assignment, if the value type is the same as field type, do not go through type conversion,
6. `otaproxy.ota_cache`: split `_FileDescriptorHelper` into several standalone helper functions, 
7. `otaproxy.ota_cache`: refactor `retrive_file` API method, split different cases handler into standalone helper methods, improve code readability and reduce code cognitive complexity,
8. `requirement.txt`: bump `typing_extension` to 4.6.3,
9. update test files accordingly.

## Behavior changes

- [x] no.

## Breaking change

- [x] No.


otaclient/ota_proxy/utils.py
```diff
@@ -39,3 +45,32 @@ async def hexdigest(self) -> str:
         return await asyncio.get_running_loop().run_in_executor(
             self._executor, self._hashf.hexdigest
         )
+
+
+class TmpCacheFileNaming:
```
Bodong-Yang, line 50: New naming scheme is introduced to naming the tmp ongoing caching file predictably with the requested URL.
This makes ota_proxy aware of the relationship between a tmp file and the URL it caches for.

This is intended for the future feature of resume downloading OTA file.

otaclient/requirements.txt
```diff
@@ -15,4 +15,4 @@ aiohttp==3.8.1
 aiofiles==22.1.0
 zstandard==0.18.0
 pycurl==7.45.1
-typing_extensions==4.5.0
\ No newline at end of file
+typing_extensions==4.6.3
```
Bodong-Yang, line 18: This is for the `dataclass_transformer` decorator used by `orm.ORMeta`.

otaclient/ota_proxy/orm.py
```diff
@@ -133,6 +154,7 @@ def check_type(self, value: Any) -> bool:
         return self.type_checker(value)
 
 
+@dataclass_transform()
 class ORMeta(type):
     """This metaclass is for generating customized <TableCls>."""
```
Bodong-Yang, line 159: With `dataclass_transform`, IDE now can correctly identify the `ORMBase` derived subclass as dataclass.

otaclient/ota_proxy/orm.py
```diff
@@ -90,29 +100,40 @@ def __get__(self, obj, objtype: type) -> FV:
...
             return getattr(obj, self._private_name)  # access via instance
-        return self  # access via class, return the descriptor
+        return self  # return descriptor instance by default
 
     def __set__(self, obj, value: Any) -> None:
         if self._skipped:
-            return
-        # handle dataclass's default value setting behavior and NULL type assignment
-        if isinstance(value, type(self)) or self.field_type == NULL_TYPE:
+            return  # ignore value setting on skipped field
```
obi-t4, line 113: Is this __set__ function compatible with previous version, especially `value is None`?
Or any logical fix is introduced?

Bodong-Yang: The logic is the same, just the implementation is simplified. If the input `value` is None, or field type is `NULL_TYPE`, or dataclass initializing, the default value will be set.

This method is refined and simplify as `NULL_TYPE` is no-longer type alias of python's built-in `None` type. Previously in runtime `type(None)` is the same as `NULL_TYPE`, resulting we have to implement special mechanism to distinguish them when we have input value as `None`. 
This is not required anymore with this PR as `NULL_TYPE` becomes real type that different from `None` type.

Previously:
`type(None) == NULL_TYPE`, so we cannot use judgement like `self.field_type is NULL_TYPE`,
With PR:
`type(None) != NULL_TYPE`, `self.field_type is NULL_TYPE` can be used now.

obi-t4 (approved):
:+1:

PR #244, opened 2023-07-19
[Feature] extend ota-file-cache-control header

Bodong-Yang:
## Description

> **Note**
> This PR is the first PR in a series of PRs(#244,  #245 , #246 ) that introduce extra feature to otaproxy. Each PR is implemented based on the previous PR, so these PRs should be reviewed and merged into main branch one by one.

> **Note**
> This PR has not yet updated otaproxy to support the extended header but just maintain the forward compatibility for it. Check the #245  for otaproxy feature update.

This PR introduces extension to `ota-file-cache-control header`, implements a set of helper functions and types to use this header in coding, update related otaclient modules (`downloader.py` and `ota_metadata.py`) to support such extension(with the new implemented helper code).

### Revision to header def

Two new directives for `ota-file-cache-control` header are introduced:
1. **file_sha256**: the sha256 hash value of **the original plain requested OTA file**(without any compression).
2. **file_compression_alg**: the compression applied to the OTA file resource, currently only `zst` is valid for this field, or empty(not presented in the header string) if no compression is applied.

## Check list

- [x] test file(s) that covers the change(s) is implemented.
- [x] local test is passed. 
- [x] design docs/implementation docs are prepared.

## Documents

1. [OTA cache mechanism design for otaproxy](https://tier4.atlassian.net/l/cp/yWPaqr8z)

## Changes

1. `app`: remove `ota-file-cache-control` definition in `otaclient.app.common` module, but directly use the one from `otaproxy` module,
2. `downloader._transfer_invalid_retrier`: use helper funcs from `otaproxy.cache_control` module to handle `ota-file-cache-control` header,
3. `downloader._prepare_header`: split the request header preparing logic as standalone method, inject OTA file digest(and compression_alg) information into request header via `ota-file-cache-control` header,
4. `downloader._check_against_cache_policy_in_resp`: check `ota-file-cache-control` header from upper otaproxy's response, check the following **Behavior with this PR** section for more details,
5. `downloader._prepare_url`: split URL preparing logic as standalone method,
6. `downloader.get_req_retry_counts`: define helper method for getting retry count from `requests.Response`,
7. `ota_metadata`: use helper funcs from `otaproxy.cache_control` module to handle `ota-file-cache-control` header,
8. `ota_metadata`: fix to inform-use-only `_persist_pa`,
9. `ota_proxy.cache_control`: implement new `ota-file-cache-control` header as dataclass, with several helper methods,
[...]

PR #245, opened 2023-07-19
[Feature & Refinement] otaproxy: OTA file sha256 aware, support new header 

Bodong-Yang:
## Description

> **Note**
> This PR is the first PR in a series of PRs(#244,  #245 , #246 ) that introduce extra feature to otaproxy. Each PR is implemented based on the previous PR, so these PRs should be reviewed and merged into main branch one by one.

> **Note**
> This PR extends otaproxy's behavior based on extended `ota-file-cache-control` header introduced in #244 .

This PR introduces refinement to otaproxy internal implementation based on the design document, adds support to the new `ota-file-cache-control` header definition and extends otaproxy's behavior with it.

### OTA file sha256 awareness for otaproxy

With header extended, now otaproxy has the ability to know the `file_sha256` and `file_compression_alg` for the requested OTA file. 

This PR lets otaproxy aware of `file_sha256` of each cache file's corresponding OTA file, uniquely manage cache entries with OTA file's `file_sha256` information. **This brings another benefit that otaproxy can actively store only one cache file for unique OTA file during multiple OTA images being downloaded, ** saving much disk space, reducing number of cached files, cache db size and eventually improving the otaproxy's performance.

In contrast, previous implementation which calculating sha256 for each cache entry(hash for cache file is meaningless as we don't use this value for validation purpose but only for unique identifying each cache file). With `file_sha256` of original OTA file being provided, we can use it instead.

### Revision to file size bucket

According to new investigation to the file size distribution in OTA image(https://tier4.atlassian.net/l/cp/uPeK0y23), file size bucket for LRU is updated.

### Cache DB schema update

To implement `file_sha256` awared otaproxy, cache db schema also updated:
1. add `file_sha256` and `file_compression_alg` field, switching to use `file_sha256` as primary key, 
2. remove unnecessary `content_type` field(we are not file server).

This is the 4th revision to the database schema.

### Compatibility concern

No compatibility breaking will occur on introducing new behavior to otaproxy.

#### request from old otaclient to new otaproxy

For old otaclient that doesn't support extended header, otaproxy will fallback to use **URL based sha256** to generate unique identifier for the requested resource, and identify the cached entry.

#### response from new otaproxy to old otaclient

By checking `ota_proxy.cache_control` module in v3.5.1:
https://github.com/tier4/ota-client/blob/d093c7b2b28bab2cbed9ac78f54392dfd9f6b529/otaclient/ota_proxy/cache_control.py#L44-L51

header parser will ignore unknown directives, and only read in known one.

## Check list

- [x] local test is passed. 

## Documents

1. [OTA cache mechanism design for otaproxy](https://tier4.atlassian.net/l/cp/yWPaqr8z)
4. [OTA proxy implementation document](https://tier4.atlassian.net/l/cp/0m2gYA7s)

## Code changes

1. `ota_proxy._consts`: pre_define some common used consts,
2. `ota_proxy.configs`: update LRU file size bucket definition, update `TABLE_DEFINITION_VERSION -> v4`, add `URL_BASED_HASH_PREFIX=="URL_"` for entry that uses `URL based sha256` when real `file_sha256` for corresponding OTA file is not available,
3. `ota_proxy.db`: update db schema, rename `sha256` -> `file_sha256`, and now it means the sha256 of original OTA file; define new `export_headers_to_client` to dump metadata as header string for use in response; update `lookup_entry` to not hardcoded to use `url` field to find entry and update `last_access` field,
4. `ota_proxy.ota_cache`: define `create_new_meta` helper function, it reads in information from client's request and upper otaproxy/remote server's response during proxy, and create cache meta for to-be-created cache file,
5. `ota_proxy.ota_cache.CacheTracker`: refine `CacheTracker`; now expects generic unique `cache_identifier` instead of tmp file name; now it does the cache finalizing; now provider tracker becomes ready when tmp cache file is opened; now tracker checks disk usage event instead, and abort on caching happens inside tracker,
6. `ota_proxy.ota_cache.CachingRegister`: refine `CachingRegister`, use pre-defined unique identifier(`file_sha256` or `URL based sha256`) to identify tracker inst,
7. [FIX] `ota_proxy.ota_cache.LRUCacheHelper`: `lookup` and `remove` methods now takes `file_sha256` instead; fix `rotate_cache` not implemented as design doc,
8. `ota_proxy.ota_cache.cache_streaming`: simplify `RemoteOTAFile` class to `cache_streaming` function, transfer most of its feature to `CacheTracker`, and now it only focus on teeing data chunks from upper to client and cache tracker,
9. `ota_proxy.ota_cache`: remove `open_remote`, re-implement the functionality in `OTACache._retrieve_file_by_downloading`,
10. `ota_proxy.ota_cache.OTACache`: refactor `_background_check_free_space`; now issues logging when disk space availability level changed,
11. [FIX] `ota_proxy.ota_cache.OTACache`: fix `_cache_entries_cleanup` not distinguishing the case of **no need to rotate cache** and **failed to rotate cache**,
12. `ota_proxy.ota_cache.OTACache`: refine `retrieve_file` API and each request handling helper functions, refactor client request parsing, upper server response parsing, preparing response to client,
13. `ota_proxy.server_app`:  refine error handling, split exception handling logic from main `_pull_data_and_send` into two context manager functions,
14.  `ota_proxy.server_app`: server_app no-longer parses the contents of client request header, but directly passthrough header to ota_cache; also now it takes header dict from ota_cache, and directly pass it back to client,
15. `ota_proxy.utils`: remove unused `AIOSHA256Hasher`, add `url_based_hash` helper function,
16. fix up test files.

## Behavior changes

Check more details from implementation document(https://tier4.atlassian.net/l/cp/0m2gYA7s).

### Previous behavior

1. otaproxy manages cache files uniquely by each cache file's sha256, sha256 hash calculating is done during caching, 
2. LRU cache rotation might work unexpected in some conditions,
3. server_app takes the responsibility of pre-processing client's request header, and constructing response header back to client,

### Behavior with this PR

1. otaproxy now manage each cache file uniquely by corresponding OTA file's sha256 value,
2. otaproxy no-longer calculates the sha256 hash for cache file,
3. LRU cache rotation's behavior aligns with implementation document,
4. ota_cache now encapsulates all details related to caching mechanism, and return to server_app with file descriptor and header dict,
5. server_app now only focus on bridging ota_cache with uvicorn, stream data chunk from file descriptor, response HTTP with header from ota_cache, and capturing the exceptions raised from ota_cache module.

## Breaking change

- [x] Yes.

1. cache db schema is updated, old ota_cache folder created and managed by old otaproxy will not be used by new otaproxy.

## Related links & tickets

1. [RT4-5484](https://tier4.atlassian.net/browse/RT4-5484)

[RT4-5484]: https://tier4.atlassian.net/browse/RT4-5484?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -144,13 +144,16 @@ ecu_info.yaml is the setting file for ECU configuration.
 
 - available_ecu_ids (optional)
 
-  This field specifies a list of all ECU ids, including directly connected, indirectly connected and itself.
+  This field specifies a list of all ECU ids to be updated.  
   Only the main ECU should have this information.  
-  If this field is not specified, value of `ecu_id` is used as this value.  
+
+  NOTE: For backward-compatibility, if this `available_ecu_ids` filed does not exist, the ecu_id of the main ecu is considered the entry for available_ecu_ids.
+
+  NOTE: for web.auto user, web.auto agent will use `available_ecu_ids` to generate update request, only ECUs listed in `available_ecu_ids` will be included in the update request list.
 
   NOTE: The difference between secondaries and available_ecu_ids:
 
-  `secondaries` lists the directly connected children ECUs, available_ecu_ids consists of all children ECU ids(including directly connected, indirectly connected and itself).
+  `secondaries` lists the directly connected children ECUs, available_ecu_ids consists of all children ECU ids(including directly connected, indirectly connected and itself) to be updated.
 
 #### The default setting
 
--- a/docs/SERVICES.md
+++ b/docs/SERVICES.md
@@ -3,6 +3,12 @@
 
 ## Table of Contents
 
+- [ota_metafiles.proto](#ota_metafiles-proto)
+    - [DirectoryInf](#otaclient-DirectoryInf)
+    - [PersistentInf](#otaclient-PersistentInf)
+    - [RegularInf](#otaclient-RegularInf)
+    - [SymbolicLinkInf](#otaclient-SymbolicLinkInf)
+  
 - [otaclient_v2.proto](#otaclient_v2-proto)
     - [RollbackRequest](#OtaClientV2-RollbackRequest)
     - [RollbackRequestEcu](#OtaClientV2-RollbackRequestEcu)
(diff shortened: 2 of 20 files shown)


obi-t4:
@Bodong-Yang 
> **This brings another benefit that otaproxy can actively store only one cache file for unique OTA file during multiple OTA images being downloaded

In the previous implementation, 

> In contrast, previous implementation which calculating sha256 for each cache entry(hash for cache file is meaningless as we don't use this value for validation purpose but only for unique identifying each cache file).

If so, the previous implementation is also `only one cache file for unique OTA file`?

Bodong-Yang:
> @Bodong-Yang
> 
> > **This brings another benefit that otaproxy can actively store only one cache file for unique OTA file during multiple OTA images being downloaded
> 
> In the previous implementation,
> 
> > In contrast, previous implementation which calculating sha256 for each cache entry(hash for cache file is meaningless as we don't use this value for validation purpose but only for unique identifying each cache file).
> 
> If so, the previous implementation is also `only one cache file for unique OTA file`?

@obi-t4 Yes, previous implementation also store only one copy for unique OTA file, but the problem is both otaproxy and otaclient cannot benefit from it.

The previous implementation manages the cache file by **cache files' sha256(NOTE: not the hash of original uncompressed OTA file)**, but in the database the primary key is URL, and cache entry looks up is using URL.

So if the same file from different image is requested, although the corresponding cache file with the same hash might exist in the cache, the otaproxy will still need to re-download it again, until otaproxy fully downloads the whole file and calculate the hash of it, it will not know their is already a same file presented in the cache folder, only thing it can do is to not save the same file again.

So what I mean `hash for cache file is meaningless` for both otaclient and otaproxy is that, otaclient only knows the hash value of the **original plain uncompressed OTA file**, so if the compression is applied, then the `hash value for cache entry` is not equal to the `hash value for uncompressed OTA file`. Also otaproxy doesn't know the requested OTA file's cache value until it finishes the downloading, so it cannot pre-actively serving the file and avoid same OTA file downloading.

obi-t4:
> The previous implementation manages the cache file by cache files' sha256(NOTE: not the hash of original uncompressed OTA file), if compression is applied to the file, cache file sha256 will be different from original uncompressed OTA file.

Ah, I see. Thanks!

obi-t4:
@Bodong-Yang Let me confirm one thing about compatibility.

If there are two ota-client under the new ota proxy:
```
new ota-proxy - new ota-client (this should work)
              - old ota-client (in the secondary ecu)
```

And if are there two ota-client under the old ota-proxy:
```
old ota-proxy - old ota-client (this should work)
              - new ota-client (in the secondary ecu)
```
Both scenario work?

Bodong-Yang:
> @Bodong-Yang Let me confirm one thing about compatibility.
> 
> If there are two ota-client under the new ota proxy:
> 
> ```
> new ota-proxy - new ota-client (this should work)
>               - old ota-client (in the secondary ecu)
> ```
> 
> And if are there two ota-client under the old ota-proxy:
> 
> ```
> old ota-proxy - old ota-client (this should work)
>               - new ota-client (in the secondary ecu)
> ```
> 
> Both scenario work?

Yes, both scenario should work.
for `new otaproxy <-> old otaclient`, new otaproxy internally will fallback to use **URL based hashing** to manage cache file, and lookup cache file by URL(behave like the previous implementation).
for `old otaproxy <-> new otaclient`, as the code I paste in the PR description, the old otaclient's ota-file-cache-control header parser will just ignore unknown directives in the otaclient's request, and it will work as it.

obi-t4 (commented):
I left some comments but basically it is OK.

Bodong-Yang:
@obi-t4 PR is updated up to [c1a817f](https://github.com/tier4/ota-client/pull/245/commits/c1a817f941c9c84908aa6ef263521fa042c3e7e3), thank you!

obi-t4 (approved):
LGTM :+1:

PR #248, opened 2023-07-31
doc: update document especially available_ecu_ids

obi-t4:
## Description

Update document especially availale_ecu_ids.
Also docs/SERVICES.md has been updated according to DEVELOPMENT.md

## Check list

N/A (no test are needed)

- [ ] local test is passed. 

### Current behavior

Current specification for available_ecu_ids is unclear.

### Behaivor after fix

This PR adds more descriptions.
And adds the description who need to filtering out the update reqest.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -144,13 +144,15 @@ ecu_info.yaml is the setting file for ECU configuration.
 
 - available_ecu_ids (optional)
 
-  This field specifies a list of all ECU ids, including directly connected, indirectly connected and itself.
+  This field specifies a list of all ECU ids to be updated.  
   Only the main ECU should have this information.  
-  If this field is not specified, value of `ecu_id` is used as this value.  
+  For compatibility, if this `available_ecu_ids` filed does not exist, the ecu_id of the main ecu is considered the entry for available_ecu_ids.
+
+  NOTE: The OTA client user (e.g. Agent) has responsibility for filtering out update requests by using this ecu list.
 
   NOTE: The difference between secondaries and available_ecu_ids:
 
-  `secondaries` lists the directly connected children ECUs, available_ecu_ids consists of all children ECU ids(including directly connected, indirectly connected and itself).
+  `secondaries` lists the directly connected children ECUs, available_ecu_ids consists of all children ECU ids(including directly connected, indirectly connected and itself) to be updated.
 
 #### The default setting
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -146,9 +146,10 @@ ecu_info.yaml is the setting file for ECU configuration.
 
   This field specifies a list of all ECU ids to be updated.  
   Only the main ECU should have this information.  
-  For compatibility, if this `available_ecu_ids` filed does not exist, the ecu_id of the main ecu is considered the entry for available_ecu_ids.
 
-  NOTE: The OTA client user (e.g. Agent) has responsibility for filtering out update requests by using this ecu list.
+  NOTE: For backward-compatibility, if this `available_ecu_ids` filed does not exist, the ecu_id of the main ecu is considered the entry for available_ecu_ids.
+
+  NOTE: for web.auto user, web.auto agent will use `available_ecu_ids` to generate update request, only ECUs listed in `available_ecu_ids` will be included in the update request list.
 
   NOTE: The difference between secondaries and available_ecu_ids:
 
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -144,13 +144,15 @@ ecu_info.yaml is the setting file for ECU configuration.
 
 - available_ecu_ids (optional)
 
-  This field specifies a list of all ECU ids, including directly connected, indirectly connected and itself.
+  This field specifies a list of all ECU ids to be updated.  
   Only the main ECU should have this information.  
-  If this field is not specified, value of `ecu_id` is used as this value.  
+  For compatibility, if this `available_ecu_ids` filed does not exist, the ecu_id of the main ecu is considered the entry for available_ecu_ids.
```
Bodong-Yang, line 149: I think this line should also prefixed with **NOTE**(or even **WARNING**), as user might think empty `available_ecu_ids` means not update any ECUs.
Also, I think this is `backward-compatibility`?

docs/README.md
```diff
@@ -144,13 +144,15 @@ ecu_info.yaml is the setting file for ECU configuration.
...
+
+  NOTE: The OTA client user (e.g. Agent) has responsibility for filtering out update requests by using this ecu list.
```
Bodong-Yang, line 151: What about refining as follow:
```
NOTE: for web.auto user, web.auto agent will use `available_ecu_ids` to generate update request, only ECUs listed in `available_ecu_ids` will be included in the update request list.
```

as otaclient can work not only with web.auto agent, but also other implementation that using otaclient. 
For other implementation, they can use other kinds of implementation to generate the update request. The use of `available_ecu_ids` is required by web.auto agent, but it is not a requirement to other implementation that uses otaclient.

obi-t4:
@Bodong-Yang 
update in https://github.com/tier4/ota-client/pull/248/commits/a00549d677d672fa554004c5081ae89b31d5946a, thanks!

docs/README.md
Bodong-Yang: @obi-t4 After some consideration, I agree that all OTA client user should follow the spec of `available_ecu_ids`, a.k.a, `available_ecu_ids` should represent the ECUs to be updated.

Bodong-Yang: ( We can add another line says that `alternative client implementation can also use `available_ecu_ids`, but it is not mandatory.`

obi-t4:
@Bodong-Yang 

> What if some ECUs presented in secondary_ecus field, but not listed in available_ecu_ids, should OTA client include the status of these ECUs in the status response? In other word, should otaclient tracks ECU that listed in secondary_ecus but not in available_ecu_ids?

`secondary_ecus` and `available_ecu_ids` are the independent information.
All `secondary_ecus` responses that can respond should be included in the Status response.
available_ecu_ids is the information for ota-client user(=caller, aka agent).

OTA client doesn't have a ECU filtering policy.

> Another question is, should we expect that secondary_ecus and available_ecu_ids are synced? A.k.a, available_ecu_ids are always either the same as secondary_ecus, or available_ecu_ids==<my_ecu_id> + secondary_ecus?

they are the independent information.

Bodong-Yang:
> @Bodong-Yang
> 
> > What if some ECUs presented in secondary_ecus field, but not listed in available_ecu_ids, should OTA client include the status of these ECUs in the status response? In other word, should otaclient tracks ECU that listed in secondary_ecus but not in available_ecu_ids?
> 
> `secondary_ecus` and `available_ecu_ids` are the independent information. All `secondary_ecus` responses that can respond should be included in the Status response. available_ecu_ids is the information for ota-client user(=caller, aka agent).
> 
> OTA client doesn't have a ECU filtering policy.
> 
> > Another question is, should we expect that secondary_ecus and available_ecu_ids are synced? A.k.a, available_ecu_ids are always either the same as secondary_ecus, or available_ecu_ids==<my_ecu_id> + secondary_ecus?
> 
> they are the independent information.

@obi-t4 Let me double confirm it, otaclient status API should include status from ECUs that listed in `secondary_ecus`, even they are not listed in `available_ecu_ids` and not going to receive OTA update, is that correct?

Also, what about my self ECU? Should status API always include self ECU's status in `ecu` status list even it is not listed in `available_ecu_ids` and not going receive OTA update?

Bodong-Yang (approved):
Thank you for the updates!

PR #253, opened 2023-08-30
[Fix] grub: fix unintended behavior when updating grub_default

Bodong-Yang:
## Description

Current implementation of grub_default file updating method has pitfall when option is specified multiple times in the original grub_default file.

For example, if the input `/etc/default/grub` has the following content:
```
GRUB_TIMEOUT=6
GRUB_TIMEOUT=12
GRUB_TIMEOUT=32
```
with current implementation, the updated grub default will be as follow:
```
GRUB_TIMEOUT=0 # this is the value pre-defined in otaclient code
GRUB_TIMEOUT=12
GRUB_TIMEOUT=32
```
This behavior is not expected as the expected pre-defined value `GRUB_TIMEOUT=0` is not honored and being overridden.

The proper behavior is that the predefined value should have the highest priority, and not impacted by any following up duplicated lines, and the output should be as follow:
```
GRUB_TIMEOUT=0 # this is the value pre-defined in otaclient code
```

## Check list

- [x] local test is passed. 

### Current behavior

When option is specified multiple times, all duplicated option lines will be preserved as it. If predefined default value is defined for this option, the predefined value will only set to the first presented option line, resulting in predefined value not being used(override by later option line).

### Behaivor after fix

The grub_default file is updated with the following rules:

1. predefined default value has the highest priority, and overrides any presented options in the original grub_default,
2. option that specified multiple times will be merged into one, and the latest specified value will be used, or predefined default value will be used if predefined default value is defined.


obi-t4 (approved):

yn-mrse:
image

yn-mrse:
test_codeが妥当であると思われる。
かつそれがパスしているので問題ないと判断しました。

PR #258, opened 2023-10-27
[Fix] grub: workaround for OTA failure/otaclient crashing if rootfs is not sda

Bodong-Yang:
## Description

This PR applies a workaround to deal with otaclient crashing and OTA failed when rootfs device is not assigned name as `sda`.

With this PR, the slot naming is decoupled with the actual device name, the slot name will be fixed to `sda<partition_id>`, and otaclient will search for ota-partition folder with the slot name.
For example, if rootfs device is `nvme0n1`, and `active dev: nvme0n1p3`, `standby dev: nvme0n1p4`, then the slot_id for active slot is still `sda3`, for standby slot is still `sda4`.

**WIth workaround implemented this PR, OTA will work if no matter when the rootfs device's name changed and what name is assigned to rootfs device.**

> [!WARNING] 
> **Deprecation Warning**: the current slot managing/detecting mechanism design is proved to be not so robust, with the inaccurate assumption that rootfs device will always be assigned name as `sda`. 
> 
> This PR is served as a quick workaround to make otaclient properly functioning if the above assumption is broken due to rootfs device is not name as `sda`. This PR doesn't mean to solve the problem fundamentally, but just for maintaining backward compatibility.

> [!IMPORTANT]
> A new mechanism to managing/detecting slot will be introduced in the future to replace the current mechanism. 

## Check list

- [x] local test files are passed. 
- [x] normal OTA on VM works.
- [x] OTA when rootfs is not sda works.
- [x] rootfs naming changed between OTA reboot works.
- [x] extra ota-partition folder with name other than `ota-partition.sda` doesn't impact otaclient's functionality. 

Check [issue reproduce and testing](https://tier4.atlassian.net/l/cp/UqVYVjM0) report for more details.

### Current behavior

1. otaclient will crash during OTA finalizing switching boot if ota-partition folders other than `ota-partition.sda*` presented,
2. OTA will fail if rootfs device's name changed before and after OTA reboot,
3. OTA status will be lost if rootfs device's name changed, corresponding extra ota-partition folders will be created unexpectedly(i.e., `ota-partition.sdb*`, `ota-partition.nvme0n1p*`, etc).

### Behaivor after fix

1. slot naming is fixed to `sda<partition_id>`,
2. OTA will work and otaclient will not crash even if rootfs device's name changed at any time, otaclient will always locate the correct ota-partition folder.

## Related links & ticket

[RT4-7472](https://tier4.atlassian.net/browse/RT4-7472)

## TODO
[...]

PR #259, opened 2023-11-02
[Refinement] refine error handling, error report on status API now covers boot controller startup failure

Bodong-Yang:
## Description

> [!NOTE]
> This PR is based on previous PR #258.

### Error handling refinements

This PR introduces refinements to otaclient's error handling, including:
1. error report now covers boot controller startup failure, boot controller startup failed will not crash the otaclient anymore, and error information is available via status API,
2. simplify and flatten the error handling structure, regulate the error handling policy regarding each otaclient components,
3. refine and redefine the OTA errors definition, refine error description,
4. refine the logging when OTA error occurs.

### otaclient startup procedure refinements

otaclient core(`otaclient.OTAClient`) and boot controller are launched separately, exceptions during these two components startup is handled and captured, so that the whole otaclient startup procedure can finish and grpc server can be launched.
The error information during startup is recorded and available via status API.

### DEBUG_MODE and `failure_traceback` in status API response

Also, previously `failure_traceback` is always included into the status API response, considering the traffic caused by unpredictable amount of bytes from `failure_traceback` field, this PR introduces a new config option `DEBUG_MODE`(default is False), and `failure_traceback` is only included when `DEBUG_MODE` is on. 

## Check list

- [x] local test is passed. 
- [x] VM test is passed. 

### OTA errors handling structure change

Check [documentation](https://tier4.atlassian.net/l/cp/bXyzLG0i) for OTA errors handling design for more details.

### OTA errors definition change and descriptions update

Most of the errors are changed to `unrecoverable OTA error` now, only `E_NETWORK, E_OTAMETA_DOWNLOAD_FAILED` and `E_OTA_BUSY, E_INVALID_STATUS_FOR_OTAROLLBACK` are still recoverable errors.

Error descriptions are simpler and include necessary basic information for better understanding what is going on.

### otaclient startup workflow change

Previously, otaclient core `OTAClient` launches boot controller when it is launching, but not handling the error raised by boot controller.

With this PR,  a new `otaclient.OTAServicer` is introduced as "otaclient composer". It is in charge of launch boot controller and otaclient core(`otaclient.OTAClient`) separately, and then compose then together as a fully functional otaclient instance.

Exceptions during launching otaclient core and boot controller are captured and handled, and error information is available via status API.

## Behavior changes

- [x] Yes, external behaivor (will impact user experience).

### Previous behavior

1. otaclient will crash if boot controller startup failed, grpc server and otaclient core will not start,
2. `failure_traceback` is always included into status API response.

### Behavior with this PR

1. otaclient will not crash on boot controller startup failed, otaclient grpc server will still launch and error information can be queried from status API,
2. `failure_traceback` is included into the status API response only when DEBUG_MODE is on.

## Breaking change

- [X] No.

## Related links & tickets

[RT4-7473](https://tier4.atlassian.net/browse/RT4-7473)

[RT4-7473]: https://tier4.atlassian.net/browse/RT4-7473?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/otaclient/app/configs.py
+++ b/otaclient/app/configs.py
@@ -184,6 +184,8 @@ class BaseConfig(_InternalSettings):
     # default version string to be reported in status API response
     DEFAULT_VERSION_STR = ""
 
+    DEBUG_MODE = False
+
 
 # init cfgs
 server_cfg = OtaClientServerConfig()
--- a/otaclient/app/errors.py
+++ b/otaclient/app/errors.py
@@ -82,7 +82,7 @@ class OTAError(Exception):
     def failure_errcode_str(self) -> str:
         return f"{self.ERROR_PREFIX}{self.failure_errcode.to_errcode_str()}"
 
-    def get_failure_traceback(self, *, splitter="") -> str:
+    def get_failure_traceback(self, *, splitter="\n") -> str:
         return splitter.join(
             traceback.format_exception(type(self), self, self.__traceback__)
         )
@@ -91,16 +91,19 @@ class OTAError(Exception):
         """Return failure_reason str."""
         return f"{self.failure_errcode_str}: {self.failure_description}"
 
-    def get_error_report(self, title: str) -> str:
-        _traceback = self.get_failure_traceback(splitter="\n")
+    def get_error_report(self, title: str = "") -> str:
+        """The detailed failure report for debug use."""
         return (
             f"\n{title}\n"
             "\n------ failure_reason ------\n"
             f"{self.get_failure_reason()}"
             "\n------ end of failure_reason ------\n"
-            "\n------ failure traceback ------\n"
-            f"failure_traceback: {_traceback}"
-            "\n------ end of failure traceback ------\n"
+            "\n------ exception informaton ------\n"
+            f"{self!r}"
+            "\n------ end of exception informaton ------\n"
+            "\n------ exception traceback ------\n"
+            f"{self.get_failure_traceback()}"
+            "\n------ end of exception traceback ------\n"
         )
 
 
--- a/otaclient/app/errors.py
+++ b/otaclient/app/errors.py
@@ -111,9 +111,7 @@ class OTAError(Exception):
 # ------ Network related error ------
 #
 
-_NETWORK_ERR_DEFAULT_DESC = (
-    "network connection unstable, please check the connection and try again"
-)
+_NETWORK_ERR_DEFAULT_DESC = "network unstable, please check the network connection"
 
 
 class NetworkError(OTAError):
@@ -162,7 +160,7 @@ class InvalidStatusForOTARollback(OTAErrorRecoverable):
 #
 
 _UNRECOVERABLE_DEFAULT_DESC = (
-    "unrecoverable OTA error detected, please contact technical support"
+    "unrecoverable OTA error, please contact technical support"
 )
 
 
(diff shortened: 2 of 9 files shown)


otaclient/app/errors.py
Bodong-Yang, line 1: OTA errors definition is simplified considering the real world condition and rethinking on the end-user's side.

With this PR, only network related errors, OTA busy error(OTA already in-progress) and invalid OTA status(attempt to rollback when ota-status is FAILURE) are treated as recoverable, as these errors can be resolved by user themselves.

Other errors are all listed as unrecoverable, and require user to contact technical supports from us.

otaclient/app/ota_client.py
```diff
@@ -589,115 +615,189 @@ def status(self) -> wrapper.StatusResponseEcuV2:
...
+        _bootctrl_cls = get_boot_controller(ecu_info.get_bootloader())
+        _standby_slot_creator = get_standby_slot_creator(cfg.STANDBY_CREATION_MODE)
+
+        # boot controller starts up
+        try:
+            _bootctrl_inst = _bootctrl_cls()
+        except ota_errors.OTAError as e:
+            logger.error(
+                e.get_error_report(title=f"boot controller startup failed: {e!r}")
+            )
```
Bodong-Yang, line 662: boot controller is launched before otaclient core(OTAClient) starts, and we capture and handle the boot controller startup error here.

Error will be captured and parsed, error report(including full traceback) will be logged, failure reason will be available via status API. When DEBUG_MODE is on, full traceback will also be available via status API.

otaclient/app/ota_client.py
```diff
@@ -589,115 +615,189 @@ def status(self) -> wrapper.StatusResponseEcuV2:
...
+                boot_controller=_bootctrl_inst,
+                create_standby_cls=_standby_slot_creator,
+                my_ecu_id=ecu_info.ecu_id,
+                control_flags=control_flags,
+                proxy=proxy,
+            )
+        except ota_errors.OTAError as e:
+            logger.error(
+                e.get_error_report(title=f"otaclient core startup failed: {e!r}")
+            )
```
Bodong-Yang, line 689: OTAClient startup failure also being captured and parsed here.

otaclient/app/ota_client.py
```diff
@@ -589,115 +615,189 @@ def status(self) -> wrapper.StatusResponseEcuV2:
...
+        )
 
     async def get_status(self) -> wrapper.StatusResponseEcuV2:
-        return await self._run_in_executor(self._otaclient.status)
+        # otaclient is not started due to boot control startup failed
+        if self._otaclient_inst is None:
+            return self._otaclient_startup_failed_status
+
+        # otaclient core started, query status from it
+        return await self._run_in_executor(self._otaclient_inst.status)
```
Bodong-Yang, line 803: If otaclient fails to start, status API will return the pre-built status API response, which is generated based on parsing boot controller startup failure/otaclient core startup failure.

Bodong-Yang: If otaclient is started, then use the status report generated by running otaclient instance.

Bodong-Yang:
Rebase the PR to cleanup unwanted commits.

otaclient/app/configs.py
```diff
@@ -184,6 +184,8 @@ class BaseConfig(_InternalSettings):
     # default version string to be reported in status API response
     DEFAULT_VERSION_STR = ""
 
+    DEBUG_MODE = False
+
```
Bodong-Yang, line 188: About DEBUG_MODE, when enabled, currently otaclient will behave as follow:
1. `failure_traceback` will also be reported via status API(normally only being logged).

In the future, more features like logging to file when DEBUG_MODE is on might be supported.

obi-t4:
> Also, previously failure_traceback is always included into the status API response, considering the traffic caused by unpredictable amount of bytes from failure_traceback field, this PR introduces a new config option DEBUG_MODE(default is False), and failure_traceback is only included when DEBUG_MODE is on.

In this PR, why do you change this behavior? If this required changes, please write the reason, and why the previous code was OK to include trace back?

Bodong-Yang:
> > Also, previously failure_traceback is always included into the status API response, considering the traffic caused by unpredictable amount of bytes from failure_traceback field, this PR introduces a new config option DEBUG_MODE(default is False), and failure_traceback is only included when DEBUG_MODE is on.
> 
> In this PR, why do you change this behavior? If this required changes, please write the reason, and why the previous code was OK to include trace back?

Sure! 

So when `failure_traceback` was introduced in status API response format v2 many months ago, **I was thinking of exposing the full detailed error messages to the end-user side to help end-user debugging the problem**.
But now I realize that, **in normal case, we should not rely on end-user to debug/fix things on their own when unrecoverable errors occur, they should call for technical support and it is our responsibility to fix problem for them**. 
So the `failure_traceback` message included in the status API response is meaningless for most of the end-user(`failure_reason` that describing what is going on and how should the end-user do in brief is enough).

For technical support like us, when end-user side has problem, the fastest way for us to locate the problem is to check the otaclient logging on cloudwatch, `failure_traceback` in status API is not so useful for us.

Also the length of `failure_traceback` is unpredictable depending on what error is raised, **I am worried about `failure_traceback` with unpredictable length will make the status API response to large and cause unwanted problem.** (I didn't think of this potential issue when introducing this field).

But the `failure_traceback` can benefit the local OTA testing, when developer directly monitors the OTA progress with `status_monitor` helper tool.

With the above considerations, I decide to make `failure_traceback` in status API as developer oriented feature, and only enabled when `DEBUG_MODE` is on.

obi-t4:
@Bodong-Yang 

> the fastest way for us to locate the problem is to check the otaclient logging on cloudwatch

We can see the traceback on the cloudwatch log, right? (I think it is enough)

obi-t4:
If you think:

> I am worried about failure_traceback with unpredictable length will make the status API response to large and cause unwanted problem. 

Do we need to enable traceback in the status API by using DEBUG_MODE flag?
I think otaclient log is enough.

Bodong-Yang:
> @Bodong-Yang
> 
> > the fastest way for us to locate the problem is to check the otaclient logging on cloudwatch
> 
> We can see the traceback on the cloudwatch log, right? (I think it is enough)

Yes, the same contents will be logged using `logger.error`.

Bodong-Yang:
> If you think:
> 
> > I am worried about failure_traceback with unpredictable length will make the status API response to large and cause unwanted problem.
> 
> Do we need to enable traceback in the status API by using DEBUG_MODE flag? I think otaclient log is enough.

When using `status_monitor` helper tool with `failure_traceback` enabled, the `failure_traceback` can be directly shown in the `status_monitor`, I think it is a good developer oriented feature(if the otaclient being tested is running as background service).

otaclient/app/errors.py
```diff
@@ -88,240 +68,194 @@ class OTAError(Exception):
...
+_UNRECOVERABLE_DEFAULT_DESC = (
+    "unrecoverable OTA error, please contact technical support"
+)
 
 
-class InvalidStatusForOTARollback(OTAErrorRecoverable):
-    module: OTAModules = OTAModules.API
-    errcode: OTAErrorCode = OTAErrorCode.E_INVALID_STATUS_FOR_OTAROLLBACK
-    desc: str = f"{_RECOVERABLE_DEFAULT_DESC}: current ota-status indicates it should not accept ota rollback"
+class OTAErrorUnRecoverable(OTAError):
```
obi-t4, line 168:
```suggestion
class OTAErrorUnrecoverable(OTAError):
```
or if we use `UnRecoverable`, then we need to employee `UN_RECOVERABLE`. `Unrecoverable` is natural I think.

obi-t4 (commented):
I left a comment but basicall OK!

otaclient/app/errors.py
Bodong-Yang: Updated in https://github.com/tier4/ota-client/pull/259/commits/16cc689223d26e5adce9534acb09df8c64b8cc8d

obi-t4 (approved):
LGTM :+1:

PR #268, opened 2023-12-26
[Refinement] create_standby: refine copy_tree to support container mode

Bodong-Yang:
## Description

During implementing the support for container mode, I found that `copy_tree.py` module(which is for handling preserving files from active slot to standby slot) doesn't support container mode(a.k.a, it expects that the src path is always rooted at `/`). 

> [!Note]
> The changes in this PR are cherry-picked from the container mode implementation working branch.

This PR implements a drop-in replacement for `copy_tree` module at  `create_standby.common.PersistFilesHandler`, which supports handling files preserving under container mode and simplify the implementation.

## Check list

- [x] local test is passed. 
- [x] VM test is passed.

## Changes

1. `copy_tree` module is re-implemented as `create_standby.common.PersistFilesHandler`,
2. uid/gid mapping logic now is implemented in `utils.unix` package.

## Behavior changes

- [x] No.

## Breaking change

- [x] No.


tests/test_create_standby/test_persist_files_handling.py
Bodong-Yang, line 1: The previous `test_copy_tree.py` is moved to `create_standby/test_persis_files_handling.py`, with test setup now as follow:
creating two folders(`src`, `dst`) to simulate `active slot` and to-be-updated `standby slot`. Specific files/dirs under `src` are required to be preserved from `src` to `dst`, which specified by the canonical path(a.k.a, path rooted at '/').

For example, we create `src/A/B/C` and want to preserve `A/B/C` to `dst`, then we will setup test as follow:
1. `create_files` will create `src` and `dst` folders as `active slot` and `standby slot`,
2. `create_files` will create `A/B/C` under `src` folder,
3. `PersistFilesHandler` is setup with test passwd/group db and `src_rootfs=src, dst_rootfs=dst`,
4. `PersistFilesHandler.preserve_persist_entry` is called with params: `_persist_entry="/A/B/C"`,
5. check `src/A/B/C` against `dst/A/B/C`.

> [!Important]
> **All test assertion cases are all unchanged**(i.e., uid/gid mapping check, mode check , etc), besides the rules of `src as active slot` and `dst as standby slot` are applied.

obi-t4 (approved):
LGTM :+1:

PR #390, opened 2024-09-14
refactor(otaclient_common): retry_task_map now handles edge condition when input task iterable yields no task

Bodong-Yang:
This PR handles an edge condition when input task iterable yields nothing, which might result in the retry_task_map hang.

### DIFF
--- a/src/otaclient_common/retry_task_map.py
+++ b/src/otaclient_common/retry_task_map.py
@@ -128,7 +128,10 @@ class ThreadPoolExecutorWithRetry(ThreadPoolExecutor):
 
     def _fut_gen(self, interval: int) -> Generator[Future[Any], Any, None]:
         finished_tasks = 0
-        while self._total_task_num != -1 and finished_tasks != self._total_task_num:
+        while finished_tasks == 0 or finished_tasks != self._total_task_num:
+            if self._total_task_num < 0:
+                return
+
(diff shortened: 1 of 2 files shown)


src/otaclient_common/retry_task_map.py
```diff
@@ -120,7 +128,10 @@ def _task_done_cb(
 
     def _fut_gen(self, interval: int) -> Generator[Future[Any], Any, None]:
         finished_tasks = 0
-        while True:
+        while finished_tasks == 0 or finished_tasks != self._total_task_num:
```
Bodong-Yang, line 131: finished_tasks == 0 is the initial status.
When finished_tasks become non-zero, it indicates that tasks execution is started, then we track finished_tasks != self._total_task_num to ensure all tasks are executed. Note that self._total_tasks_num only has three kinds of possible value, see L68-L75.

src/otaclient_common/retry_task_map.py
```diff
@@ -120,7 +128,10 @@ def _task_done_cb(
...
+            if self._total_task_num < 0:
+                return
```
Bodong-Yang, line 133: -1 is set by the tasks dispatcher, when input tasks iterable doesn't yield any task. in this case, we should stop the task execution loop immediately.

PR #401, opened 2024-10-16
ci: gen_requirements_txt workflow: use DEPLOY_KEY ssh key

Bodong-Yang:
## Introduction

Due to the restriction set by github, a commit pushed by bot with default GITHUB_TOKEN will not trigger any workflows. This will result in PR requires CI test passed being stale.

This PR introduces to use repo SSH DEPLOY key to do the push to avoid the above issue according to the documentation. See https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs for more details.

Also for a final ballback mechanism, the test CI now has `on pull_request: review_requested` trigger configured.


Bodong-Yang:
Closed for now as the commit pushed with deploy key can only trigger the `on: push` workflows.

Bodong-Yang:
UPDATED: confirm that even with the deploy key being used, just as the  https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs says, the `pull_request` trigger is not fired with git push using deploy key.
But the workaround is working, the review_requested triggered is working fine!

PR #442, opened 2024-12-01
fix(backport v3.8.x): ota_proxy: fix content-encoding header is not passed through in direct proxy mode

Bodong-Yang:
## Introduction

This PR fixes a bug that prevents ota_proxy properly handling request of file being `gzip` by remote HTTP server. The cause of bug is ota_proxy doesn't include `content-encoding` header in response back to client when in direct proxy passthrough mode. 
The bug is fixed by using `multidict` when parsing/preparing headers, so `multidict<7.0,>=4.5` is added to the direct dep of otaclient. Note that `multidict` is used by aiohttp, so it is already a indirect dep of otaclient.

## Tests

- [x] OTA test with ota_proxy in direct proxy mode(by setting 0 hard space limit).
- [x] OTA test with ota_proxy in caching mode(by setting soft & hard space limit to 100).
- [x] OTA test with full local cache presented.
- [x] OTA test with caching mode and LRU cache rotating being enabled in the middle of OTA.

https://tier4.atlassian.net/browse/RT4-13280


src/ota_proxy/config.py
Bodong-Yang, line 1: Still set the soft and hard limit to a reasonable lower values.

PR #446, opened 2024-12-02
fix(ota_proxy): fix background disk space usage checker not working

Bodong-Yang:
## Introduction

This PR fixes otaproxy's background disk space usage checker not working as expected due to checker thread exits early caused by raised exception, resulting in otaproxy LRU cache rotation not being triggered on soft disk space usage reached and otaproxy cache disabled on hard space usage reached.


Bodong-Yang:
Included in #447 .

PR #454, opened 2024-12-16
feat: deps: bump to use simple-sqlite3-orm v0.5.0, update the otaproxy accordingly

Bodong-Yang:
## Introduction

This PR bumps to use simple-sqlite3-orm v0.5.0 for otaproxy, and updates the otaproxy accordingly. Also missing indexes for ota_cache metadata db (previously defined before otaclient v3.7.1) are added back to otaproxy.

Other minor changes:
1. lru_cache_helper: use burst_suppressed_logger for loggings that might be flooded.

## Tests

- [x] normal OTA e2e test with VM.
- [x] fully interrupt OTA, and re-OTA again with OTA cache.
- [x] shutdown network and restart network, try several times during OTA.
- [x] test cache rotating.


src/ota_proxy/db.py
Bodong-Yang, line 1: The missing bucket_id index and last_access index are added back now.

Bodong-Yang, line 1: All the fields of cache_meta table schema are using native data types, so no need to apply validation to any of the fields.
Since in lru_cache_helper we use `table_spec_no_validation` as row_factory specifier, no need to use SkipValidation hint on each field.

PR #455, opened 2024-12-16
docs: update DEVELOPMENT.md

airkei:
### Why
`docs/` directory is no more used, but the instructions in `DEVELOPMENT.md` still refer docs directory.

### What
Update `DEVELOPMENT.md` to describe the current procedure.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -29,7 +29,7 @@ git clone https://github.com/tier4/ota-client.git
 # enable the otaclient python venv
 . venv/bin/activate
 
-# install the downloaded otaclient 
+# install the downloaded otaclient
 pip install -e ota-client
 
 # restart otaclient to launch the target version
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -14,68 +14,6 @@ $ python3 --version
 Python 3.8.10
 ```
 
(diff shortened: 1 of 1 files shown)


Bodong-Yang:
@airkei Thank you for the PR! 👍 
However their is pitfall that, by just installing otaclient you will (probably) not be able to get it work, as we need to pre-configure the system before it can work.

Currently we are using ansible to setup the system and install otaclient. But that ansible playbook, as you know, is not public, so we cannot just copy-paste it to this repository.

Actually I think currently we should just completely remove this doc.

airkei:
@Bodong-Yang Thank you for feedback. As you say, we should not describe the instructions which depends on non-public   setup. On the other hand, I think it's better to describe the instructions as much as possible in markdown.
How about to remove the ECU installation part and keep tests parts like the latest commit?

Bodong-Yang:
> How about to remove the ECU installation part and keep tests parts like the latest commit?

I think this is OK for now. In the future, we can write a detailed installation guide based on the ansible playbook, or even provide a stripped down version of the ansible role that only setup otaclient.

Bodong-Yang (approved):
Let's remove the installation part for now. 👍 
In the future we can either provide a full and detailed guide of setting up device system and otaclient, or even provide a stripped down version of otaclient installation role.

PR #459, opened 2024-12-18
fix: ecu_status: should use any_child_ecu_in_update flag instead of any_in_update to avoid self-lockedown, still wait on any_child_ecu_in_update before reboot

Bodong-Yang:
## Introduction

This PR reverts the change that otaclient waits for any_requires_network instead of any_in_update. This change was introduced after otaclient v3.8.x. 
Also, a logic fix is introduced to use `any_child_ecu_in_update` instead of `any_in_update`.

`any_in_update` will be set when any ECU(including self-ECU) is in UPDATE. If otaclient waits for this flag before OTA reboot(at this point the self-ECU's OTA status is still UPDATING), it will be dead-locked by itself. 
Fix by using `any_child_ecu_in_update` instead, which aligns with the previous implementation at otaclient v3.8.x.

### DIFF
--- a/src/otaclient/ota_core.py
+++ b/src/otaclient/ota_core.py
@@ -579,11 +579,9 @@ class _OTAUpdater:
             )
         )
 
-        # NOTE: we don't need to wait for sub ECUs if sub ECUs don't
-        #       depend on otaproxy on this ECU.
         if proxy_info.enable_local_ota_proxy:
             wait_and_log(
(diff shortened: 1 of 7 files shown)


src/otaclient/ota_core.py
Bodong-Yang, line 1: Revert to the behavior of otaclient v3.8.x and before: wait for all child ECUs finish OTA update.

PR #465, opened 2024-12-20
feat!: reject any OTA update/rollback request on ecu_info.yaml not properly loaded

Bodong-Yang:
## Introduction

This PR implements the requirement of rejecting any OTA update/rollback request when ecu_info.yaml file is broken, and reports the failure_reason as due to ecu_info.yaml is broken.

If ecu_info.yaml is broken and cannot be parsed at startup, otaclient will set the live_ota_status to FAILURE, failure_type to UNRECOVERABLE, with failure_message: `ecu_info.yaml file is broken, please check /boot/ota/ecu_info.yaml. reject any OTA request.`

And for every incoming OTA request, otaclient will reject with failure_type.UNRECOVERABLE for every ECUs listed in the request(as otaclient with default ecu_info.yaml doesn't have contact for the sub ECUs).

For multiple ECU environment, due to when falling down to the default ecu_info, we will only have one entry(which is the main ECU) in the available_ecu_ids list, so the situation will be the same as single ECU environment.

Other changes:
1. status_monitor: now status_monitor will always push to shm when the incoming report is OTAStatusChangeReport, regardless the push interval. 

BREAKING CHANGE: now otaclient will set itself in FAILURE OTA status and reject any OTA requests when ecu_info.yaml is not properly loaded at startup.

## The behavior on FMS console

image

## Tests

- [x] (single ECU environment) confirm that when ecu_info.yaml is broken, on FMS console the OTA will fail with error message stating that ecu_info.yaml is broken.
- [x] confirm that when ecu_info.yaml file presented, normal OTA can perform.
- [x] confirm that the behavior of web.auto agent is expected.

https://tier4.atlassian.net/browse/RT4-9345

### DIFF
--- a/src/otaclient/_status_monitor.py
+++ b/src/otaclient/_status_monitor.py
@@ -323,7 +323,7 @@ class OTAClientStatusCollector:
                     self.load_report(report)
                     and self._status
                     and (
-                        isinstance(report, OTAStatusChangeReport)
+                        isinstance(report.payload, OTAStatusChangeReport)
                         or _now > _next_shm_push
                     )
                 ):
--- a/src/otaclient/_status_monitor.py
+++ b/src/otaclient/_status_monitor.py
@@ -318,7 +318,15 @@ class OTAClientStatusCollector:
                     break
 
                 # ------ push status on load_report ------ #
-                if self.load_report(report) and self._status and _now > _next_shm_push:
+                # NOTE: always push OTAStatus change report
+                if (
+                    self.load_report(report)
+                    and self._status
+                    and (
+                        isinstance(report, OTAStatusChangeReport)
+                        or _now > _next_shm_push
+                    )
+                ):
                     try:
                         self._shm_status.write_msg(self._status)
                         _next_shm_push = _now + self.shm_push_interval
--- a/src/otaclient/grpc/api_v2/servicer.py
+++ b/src/otaclient/grpc/api_v2/servicer.py
@@ -115,15 +115,7 @@ class OTAClientAPIServicer:
         # NOTE(20241220): due to the fact that OTA Service API doesn't have field
         #                 in UpdateResponseEcu msg, the only way to pass the failure_msg
         #                 to upper is by status API.
-        #                 Here we only immediately accept the OTA update request for main ECU,
-        #                 fails other ECUs, and let main ECU's OTA status tells what is going on.
         if not otaclient_cfg.ECU_INFO_LOADED_SUCCESSFULLY:
-            response.add_ecu(
-                api_types.UpdateResponseEcu(
-                    ecu_id=self.my_ecu_id,
-                    result=api_types.FailureType.NO_FAILURE,
-                )
-            )
             for _update_req in request.iter_ecu():
                 response.add_ecu(
                     api_types.UpdateResponseEcu(
--- a/src/otaclient/grpc/api_v2/servicer.py
+++ b/src/otaclient/grpc/api_v2/servicer.py
@@ -21,6 +21,7 @@ import logging
 import multiprocessing.queues as mp_queue
 from concurrent.futures import ThreadPoolExecutor
 
+import otaclient.configs.cfg as otaclient_cfg
 from otaclient._types import (
     IPCRequest,
     IPCResEnum,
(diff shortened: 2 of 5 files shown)


src/otaclient/_status_monitor.py
```diff
@@ -318,7 +318,15 @@ def _status_collector_thread(self) -> None:
...
-                if self.load_report(report) and self._status and _now > _next_shm_push:
+                # NOTE: always push OTAStatus change report
+                if (
+                    self.load_report(report)
+                    and self._status
+                    and (
+                        isinstance(report.payload, OTAStatusChangeReport)
+                        or _now > _next_shm_push
+                    )
+                ):
```
Bodong-Yang, line 329: Previously, if the OTAStatusChangeReport reported too quickly, the newly coming report will be dropped, resulting in the OTA status not being updated.

src/otaclient/ota_core.py
```diff
@@ -701,9 +694,36 @@ def __init__(
...
+            status_report_queue.put_nowait(
+                StatusReport(
+                    payload=OTAStatusChangeReport(
+                        new_ota_status=_boot_ctrl_loaded_ota_status,
+                    ),
+                )
+            )
+
+            self.started = True
+            logger.info("otaclient started")
```
Bodong-Yang, line 726: If ecu_info.yaml is not loaded properly, we will just set the otaclient as NOT started, any requests coming to otaclient will be rejected when otaclient is not started.

Bodong-Yang: Note that when ecu_info.yaml is broken, only the live_ota_status(in the memory) will be set to FAILURE, the `status` file will not be override, as ecu_info.yaml missing/broken is considered a soft failure.

Bodong-Yang:
Confirm the FMS's behavior is expected, now this PR is ready for review.

src/otaclient/grpc/api_v2/servicer.py
```diff
@@ -111,6 +112,22 @@ async def update(
         update_acked_ecus = set()
         response = api_types.UpdateResponse()
 
+        # NOTE(20241220): due to the fact that OTA Service API doesn't have field
+        #                 in UpdateResponseEcu msg, the only way to pass the failure_msg
+        #                 to upper is by status API.
+        if not otaclient_cfg.ECU_INFO_LOADED_SUCCESSFULLY:
```
airkei, line 118: Just comment.
Seems current otaclient just validate file existence and yaml format.
if otaclient expects specific keys, we can validate schema too in future PR.

src/otaclient/grpc/api_v2/servicer.py
```diff
@@ -111,6 +112,22 @@ async def update(
...
+            logger.error(
+                "ecu_info.yaml is not loaded properly, reject any update request"
+            )
+            for _update_req in request.iter_ecu():
+                response.add_ecu(
+                    api_types.UpdateResponseEcu(
+                        ecu_id=_update_req.ecu_id,
+                        result=api_types.FailureType.UNRECOVERABLE,
```
airkei, line 126: Just comment: In future, I would like to extend these error types to share much more precise reason/playbook.

src/otaclient/grpc/api_v2/servicer.py
```diff
@@ -225,6 +242,22 @@ async def rollback(
...
+            )
+            for _rollback_req in request.iter_ecu():
+                response.add_ecu(
+                    api_types.RollbackResponseEcu(
+                        ecu_id=_rollback_req.ecu_id,
+                        result=api_types.FailureType.UNRECOVERABLE,
+                    )
+                )
+            return response
+
```
airkei, line 260: nit: Can we aggregate as a function between `update` and `rollback` ?

airkei (commented):
LGTM, but there is no test code. How about to pass invalid/valid ecu_info.yaml and verify the error code?

src/otaclient/grpc/api_v2/servicer.py
Bodong-Yang: Yes, I also want to do that! :crying_cat_face: 

I just have no idea why the designer of the API doesn't include a field for including failure_msg. Might he thought that the status API can do the job of providing the failure_msg? :face_with_head_bandage: But the failure message of rejecting a request is different from the failure message of the otaclient itself.

src/otaclient/grpc/api_v2/servicer.py
Bodong-Yang: Emmmm I will persist on splitting them into two methods though, for the single-responsibility principle.

Bodong-Yang:
> LGTM, but there is no test code. How about to pass invalid/valid ecu_info.yaml and verify the error code?

Let me take some time for implementing the test for it :+1:

airkei (approved):
LGTM, thank you!!

PR #453, opened 2024-12-11
[DON'T MERGE][DEV] feat: introduce file_table with sqlite3, re-implement create_standby.rebuild_mode

Bodong-Yang:
> [!NOTE]
> Due to the size of this PR is large, this branch becomes a dev branch. Changes will gradually split from this dev branch into separated PRs.

PR #466, opened 2024-12-20
feat: add csv_parser to parse legacy OTA image's metadata CSV files into sqlite3 db 

Bodong-Yang:
## Introduction

> [!NOTE]
> This PR is split from #453 . The PR only introduces the implementation of csv_parser, but not yet integrate it into otaclient code, also the previous legacy OTA image parser implementation is not changed in this PR.

This PR introduces the implementation of parsing legacy OTA image metadata CSV files into file_table entries, and importing the parsed results as sqlite3 tables. 
NOTE that the introduced implementation is not yet integrated into otaclient code, will be integrated in other future PR.


src/ota_metadata/legacy/csv_parser.py
```diff
@@ -0,0 +1,281 @@
...
+def parse_dirs_from_csv_file(
+    _fpath: StrOrPath, _orm: FileTableDirORM, *, cleanup: bool = True
+) -> int:
+    """Compatibility to the plaintext CSV dirs.txt."""
+    _batch, _batch_cnt = [], 0
+    try:
+        with open(_fpath, "r") as f:
+            _idx = 0
+            for _idx, line in enumerate(f, start=1):
+                _new = parse_dirs_csv_line(line)
```
airkei, line 90: If this parser is only for csv, how about to use the default CSV library instead of regex? I imagine the reason to use regex is to re-use the existing parser for text, but feel that csv library for csv parser is more safe.
Another idea is to aggregate regext module as helper function between text and csv. It might be preference, but I don't mind to create/use such utility functions between Classes.
https://docs.python.org/3/library/csv.html

airkei (commented):
Thank you, left one comments. Feel better to use csv library.

Bodong-Yang:
> Feel better to use csv library.

@airkei :cry: I do like to, but again, the CSV format of the legacy OTA image is kind of nonstandard, long-long time ago I did want to use csv lib, but then I gave up.

Bodong-Yang:
( Since ancient time of this repository, the regex was in used to parse the CSV.

airkei (approved):
OK, in that case, I agree with proceeding this approach.
I also feel it might be much more maintainable to isolate legacy modules completely rather than creating utility function...

Bodong-Yang:
> I also feel it might be much more maintainable to isolate legacy modules completely rather than creating utility function...

Could you elaborate your idea with more details?

airkei:
> Could you elaborate your idea with more details?

Sure! currently there is code clone between `parser` and `csv_parser`.
```
    res.mode = int(_ma.group("mode"), 8)
    res.uid = int(_ma.group("uid"))
    res.gid = int(_ma.group("gid"))
    res.slink = de_escape(_ma.group("link"))
    res.srcpath = de_escape(_ma.group("target"))
```
My idea is to create a single function and aggregate these regex parts between parser and csv_parser.
But in this case, if we change the format of either csv or text, we need to change or separate the function. This is in "legacry" modules, will not update so frequently. so I think we don't have to take care much for the cloning. 
This might be relate to the topic between "SOLID" and "Code Clone".

Bodong-Yang:
> My idea is to create a single function and aggregate these regex parts between parser and csv_parser.
But in this case, if we change the format of either csv or text, we need to change or separate the function. This is in "legacry" modules, will not update so frequently. so I think we don't have to take care much for the cloning.
This might be relate to the topic between "SOLID" and "Code Clone".

No worry, this PR is part of my effort to rewrite the `ota_metadata.legacy` package, as we are migrating to use sqlite3 to operate the file_table. 
So later I will remove the old code when the new implementation is ready. The old `parser` module along with other old implementation will be removed.

PR #474, opened 2025-01-10
feat: support both http and grpc for logging

airkei:
### Why
- currently, otaclient doesn't wait for logging server waking up. It cause log lost.
- otaclient support only HTTP interface for logging even though logging server supports both gRPC and HTTP.

### What
- wait for logging server up with HTTP empty message before starting log transmission
- support both http and grpc for logging.
  - otaclient try to use gRPC first. If it is not available, fallback to use HTTP. 
  - In gRPC, support two types(`LOG` and `METRICS`) logging.

## Check list

- [x] local test is passed.
https://tier4.atlassian.net/wiki/spaces/OTA/pages/3481894920/OTA+Metrics

### verified test cases
- verified the following matrix test cases

|  | otaclient v3.8.6 | otaclient with this PR |
|:---|:---:|---:|
|iot-logger v1.2.0 | (HTTP) | HTTP |
|iot-logger v1.3.0 | HTTP | gRPC |

- verified that otaclient wait for iot-logger wake up with this PR.

## Documents

https://tier4.atlassian.net/wiki/spaces/OTA/pages/3481894920/OTA+Metrics

## Changes
- wait for logging server up with HTTP empty message before starting log transmission.
- if gRPC server is available, use gRPC to output log. Otherwise, use the existing HTTP.
- support two types logging
    - `LOG`
    the purpose is to collect execution log. the log will follow `LOG_FORMAT` format.
    - `METRICS`
    the purpose is to collect metrics log. the metrics log is raw format without applying any formatting.
    To emit metrics log, `extra` argument and "log_type" key is necessary.
    The actual METRICS caller part will be implemented in the later PR.
    ```
    (`logging.level(message, extra = {"log_type": _logging.LogType.METRICS})`)
    ```

## Behavior changes

- [ ] No.

## Breaking change

- [ ] Yes.
- [x] No.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,6 +30,7 @@ dependencies = [
   "grpcio==1.70.0",
   "multidict<7.0,>=4.5",
   "msgpack>=1,<1.2",
+  "otaclient_iot_logging_server_pb2 @ https://github.com/tier4/otaclient-iot-logging-server/releases/download/v1.3.0/otaclient_iot_logging_server_pb2-1.0.0-py3-none-any.whl#sha256:e5a2e6474a8b6f5656679877a5df40891e6a65c29830a31faed3d47d9973be60",
   "protobuf>=4.21.12,<5.29",
   "pydantic<3,>=2.10",
   "pydantic-settings<3,>=2.3",
--- a/samples/proxy_info.yaml
+++ b/samples/proxy_info.yaml
@@ -7,3 +7,4 @@ local_ota_proxy_listen_addr: 127.0.0.1
 local_ota_proxy_listen_port: 8082
 # if otaclient-logger is installed locally
 logging_server: "http://127.0.0.1:8083"
+logging_server_grpc: "http://127.0.0.1:8084"
--- a/src/otaclient/_logging.py
+++ b/src/otaclient/_logging.py
@@ -19,13 +19,118 @@ from __future__ import annotations
 import atexit
 import contextlib
 import logging
+import time
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from enum import Enum
 from queue import Queue
 from threading import Event, Thread
-from urllib.parse import urljoin
+from urllib.parse import urljoin, urlparse
 
+import grpc
 import requests
+from otaclient_iot_logging_server_pb2.v1 import (
+    otaclient_iot_logging_server_v1_pb2 as log_pb2,
+)
+from otaclient_iot_logging_server_pb2.v1 import (
+    otaclient_iot_logging_server_v1_pb2_grpc as log_v1_grpc,
+)
+from pydantic import AnyHttpUrl
 
 from otaclient.configs.cfg import cfg, ecu_info, proxy_info
+
+
+class LogType(Enum):
+    LOG = "log"
+    METRICS = "metrics"
+
+
+@dataclass
+class QueueData:
+    """Queue data format for logging."""
+
+    log_type: LogType
+    message: str
+
+
+class Transmitter(ABC):
+    def __init__(self, logging_upload_endpoint, ecu_id: str):
+        self.ecu_id = ecu_id
+
+    @abstractmethod
+    def send(self, log_type: LogType, message: str, timeout: int) -> None:
+        pass
+
+    @abstractmethod
+    def check(self, timeout: int) -> None:
+        pass
+
+
+class TransmitterGrpc(Transmitter):
+    def __init__(self, logging_upload_grpc_endpoint: AnyHttpUrl, ecu_id: str):
+        super().__init__(logging_upload_grpc_endpoint, ecu_id)
+
+        parsed_url = urlparse(str(logging_upload_grpc_endpoint))
+        log_upload_endpoint = parsed_url.netloc
+        channel = grpc.insecure_channel(log_upload_endpoint)
+        self._stub = log_v1_grpc.OTAClientIoTLoggingServiceStub(channel)
+
+    def send(self, log_type: LogType, message: str, timeout: int) -> None:
+        pb2_log_type = (
+            log_pb2.LogType.METRICS
+            if log_type == LogType.METRICS
+            else log_pb2.LogType.LOG
+        )
+        log_entry = log_pb2.PutLogRequest(
+            ecu_id=self.ecu_id, log_type=pb2_log_type, message=message
+        )
+        self._stub.PutLog(log_entry, timeout=timeout)
+
+    def check(self, timeout: int) -> None:
+        # health check
+        log_entry = log_pb2.HealthCheckRequest()
+        self._stub.Check(log_entry)
+
+
+class TransmitterHttp(Transmitter):
+    def __init__(self, logging_upload_endpoint: AnyHttpUrl, ecu_id: str):
+        super().__init__(logging_upload_endpoint, ecu_id)
+
+        _endpoint = f"{str(logging_upload_endpoint).strip('/')}/"
+        self.log_upload_endpoint = urljoin(_endpoint, ecu_id)
+        self._session = requests.Session()
+
+    def send(self, log_type: LogType, message: str, timeout: int) -> None:
+        # support only LogType.LOG
+        self._session.post(self.log_upload_endpoint, data=message, timeout=timeout)
+
+    def check(self, timeout: int) -> None:
+        resp = self._session.head(self.log_upload_endpoint, timeout=timeout)
+        resp.raise_for_status()
+
+
+class TransmitterFactory:
+    @staticmethod
+    def create(
+        logging_upload_endpoint: AnyHttpUrl,
+        logging_upload_grpc_endpoint: AnyHttpUrl | None,
+        ecu_id: str,
+    ):
+        if not logging_upload_grpc_endpoint:
+            return TransmitterHttp(
+                logging_upload_endpoint=logging_upload_endpoint, ecu_id=ecu_id
+            )
+
+        try:
+            _transmitter_grpc = TransmitterGrpc(
+                logging_upload_grpc_endpoint=logging_upload_grpc_endpoint, ecu_id=ecu_id
+            )
+            _transmitter_grpc.check(timeout=3)
+            return _transmitter_grpc
+        except Exception:
+            return TransmitterHttp(
+                logging_upload_endpoint=logging_upload_endpoint, ecu_id=ecu_id
+            )
 
 
 class _LogTeeHandler(logging.Handler):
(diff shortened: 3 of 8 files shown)


samples/proxy_info.yaml
```diff
@@ -6,4 +6,4 @@ enable_local_ota_proxy_cache: true
 local_ota_proxy_listen_addr: 127.0.0.1
 local_ota_proxy_listen_port: 8082
 # if otaclient-logger is installed locally
-logging_server: "http://127.0.0.1:8083"
+logging_server: "127.0.0.1:8083"
```
airkei, line 9: To change from REST to gRPC

src/otaclient/_logging.py
```diff
@@ -54,7 +80,16 @@ def _thread_main():
...
+                    # convert to the protobuf log type
+                    pb2_log_type = (
+                        log_pb2.LogType.METRICS
+                        if entry.log_type == LogType.METRICS
+                        else log_pb2.LogType.LOG
+                    )
+                    log_entry = log_pb2.PutLogRequest(
+                        ecu_id=ecu_id, log_type=pb2_log_type, message=entry.message
+                    )
+                    stub.PutLog(log_entry)
```
airkei, line 92: Call gRPC function.

tests/test_otaclient/test_logging.py
```diff
@@ -0,0 +1,187 @@
...
+                {"log_type": LogType.LOG},
+                log_pb2.LogType.LOG,
+            ),
+            (
+                "some metrics message",
+                {"log_type": LogType.METRICS},
+                log_pb2.LogType.METRICS,
+            ),
+        ],
+    )
```
airkei, line 151: each test case for logging.

tests/test_otaclient/test_logging.py
```diff
@@ -0,0 +1,187 @@
...
+            log_format_regex = r"\[.*?\]\[.*?\]-.*?:.*?:\d+,(.*)"
+            match = re.match(log_format_regex, _response.message)
+            assert match is not None, "Log message format does not match"
+            extracted_message = match.group(1)
+            assert extracted_message == log_message
+        elif _response.log_type == log_pb2.LogType.METRICS:
+            # Expect the message to be the same as the input
+            assert _response.message == log_message
+        else:
+            pytest.fail(f"Unexpected log type: {_response.log_type}")
```
airkei, line 187: verify each parameter(ecu_id, log_type, message(format/raw)).

tests/test_otaclient/test_logging.py
```diff
@@ -0,0 +1,187 @@
...
+        mocker.patch(f"{MODULE}.cfg", self._cfg)
+
+    @pytest.fixture(autouse=True)
+    def mock_ecu_info(self, mocker: MockerFixture):
+        self._ecu_info = ECUInfo(ecu_id=TestLogClient.ECU_ID)
+        mocker.patch(f"{MODULE}.ecu_info", self._ecu_info)
+
+    @pytest.fixture(autouse=True)
+    def mock_proxy_info(self, mocker: MockerFixture):
+        self._proxy_info = ProxyInfo(
```
airkei, line 96: override each file only for test

airkei, line 96: override each file only for test

tests/test_otaclient/test_logging.py
```diff
@@ -0,0 +1,187 @@
...
+            logging_server=TestLogClient.OTA_CLIENT_LOGGING_SERVER
+        )
+        mocker.patch(f"{MODULE}.proxy_info", self._proxy_info)
```
airkei, line 99: override each file only for test

src/otaclient/_logging.py
```diff
@@ -13,38 +13,136 @@
...
+    LOG = "log"
+    METRICS = "metrics"
+
+
+@dataclass
+class QueueData:
+    """Queue data format for logging."""
+
+    log_type: LogType
+    message: str
```
airkei, line 51: changed queue from str to `QueueData`.

tests/test_otaclient/test_log_setting.py
```diff
@@ -17,26 +17,44 @@
...
+            _logging.LogType.LOG,
+        ),
+        (
+            "emit one metrics entry",
+            {"log_type": _logging.LogType.METRICS},
+            _logging.LogType.METRICS,
+        ),
+    ],
+)
+def test_server_logger(test_log_msg, test_extra, expected_log_type):
```
airkei, line 44: add test cases for metrics.

src/otaclient/_logging.py
```diff
@@ -13,38 +13,136 @@
...
+
+class TransmitterFactory:
+    @staticmethod
+    def create(logging_upload_endpoint: AnyHttpUrl, ecu_id: str):
+        try:
+            _transmitter_grpc = TransmitterGrpc(logging_upload_endpoint, ecu_id)
+            _transmitter_grpc.check(timeout=3)
+            return _transmitter_grpc
+        except Exception:
+            return TransmitterHttp(logging_upload_endpoint, ecu_id)
```
airkei, line 118: Factory Class.

src/otaclient/_logging.py
```diff
@@ -13,38 +13,136 @@
...
+    def __init__(self, logging_upload_endpoint, ecu_id: str):
+        self.ecu_id = ecu_id
+
+    @abstractmethod
+    def send(self, log_type: LogType, message: str, timeout: int) -> None:
+        pass
+
+    @abstractmethod
+    def check(self, timeout: int) -> None:
+        pass
```
airkei, line 64: Base Class.

src/otaclient/_logging.py
```diff
@@ -13,38 +13,136 @@
...
+        )
+        log_entry = log_pb2.PutLogRequest(
+            ecu_id=self.ecu_id, log_type=pb2_log_type, message=message
+        )
+        self._stub.PutLog(log_entry, timeout=timeout)
+
+    def check(self, timeout: int) -> None:
+        # health check
+        log_entry = log_pb2.HealthCheckRequest(service="")
+        self._stub.Check(log_entry)
```
airkei, line 90: Implementation Class for gRPC.

src/otaclient/_logging.py
```diff
@@ -13,38 +13,136 @@
...
+        self.log_upload_endpoint = urljoin(_endpoint, ecu_id)
+        self._session = requests.Session()
+
+    def send(self, log_type: LogType, message: str, timeout: int) -> None:
+        # support only LogType.LOG
+        with contextlib.suppress(Exception):
+            self._session.post(self.log_upload_endpoint, data=message, timeout=timeout)
+
+    def check(self, timeout: int) -> None:
+        pass
```
airkei, line 107: Implementation Class for HTTP.

src/otaclient/grpc/log_v1/otaclient_iot_logging_server_v1_pb2.py
airkei, line 1: https://github.com/tier4/otaclient-iot-logging-server is private repo. So think it's better to put gRPC methods and it's README rather than installing protocol buffer whl.

src/otaclient/_logging.py
```diff
@@ -19,32 +19,181 @@
...
+        try:
+            _transmitter_grpc = TransmitterGrpc(
+                logging_upload_grpc_endpoint=logging_upload_grpc_endpoint, ecu_id=ecu_id
+            )
+            _transmitter_grpc.check(timeout=3)
+            return _transmitter_grpc
+        except Exception:
+            return TransmitterHttp(
+                logging_upload_endpoint=logging_upload_endpoint, ecu_id=ecu_id
+            )
```
airkei, line 133: Factory Class.

src/otaclient/_logging.py
```diff
@@ -19,32 +19,181 @@
...
+        )
+        log_entry = log_pb2.PutLogRequest(
+            ecu_id=self.ecu_id, log_type=pb2_log_type, message=message
+        )
+        self._stub.PutLog(log_entry, timeout=timeout)
+
+    def check(self, timeout: int) -> None:
+        # health check
+        log_entry = log_pb2.HealthCheckRequest()
+        self._stub.Check(log_entry)
```
airkei, line 92: gRPC class.

src/otaclient/_logging.py
```diff
@@ -19,32 +19,181 @@
...
+        self.log_upload_endpoint = urljoin(_endpoint, ecu_id)
+        self._session = requests.Session()
+
+    def send(self, log_type: LogType, message: str, timeout: int) -> None:
+        # support only LogType.LOG
+        self._session.post(self.log_upload_endpoint, data=message, timeout=timeout)
+
+    def check(self, timeout: int) -> None:
+        resp = self._session.head(self.log_upload_endpoint, timeout=timeout)
+        resp.raise_for_status()
```
airkei, line 109: HTTP Class.

src/otaclient/_logging.py
```diff
@@ -19,32 +19,181 @@
...
         def _thread_main():
-            _session = requests.Session()
+            self._wait_for_log_server_up(
+                logging_upload_endpoint=logging_upload_endpoint, ecu_id=ecu_id
+            )
+            _transmitter = TransmitterFactory.create(
+                logging_upload_endpoint=logging_upload_endpoint,
+                logging_upload_grpc_endpoint=logging_upload_grpc_endpoint,
+                ecu_id=ecu_id,
+            )
```
airkei, line 196: Before entering while loop, wait for log server wake up by HTTP **without** timeout.

pyproject.toml
```diff
@@ -30,6 +30,7 @@ dependencies = [
   "grpcio==1.70.0",
   "multidict<7.0,>=4.5",
   "msgpack>=1,<1.2",
+  "otaclient_iot_logging_server_pb2 @ https://github.com/tier4/otaclient-iot-logging-server/releases/download/v1.3.0/otaclient_iot_logging_server_pb2-1.0.0-py3-none-any.whl#sha256:e5a2e6474a8b6f5656679877a5df40891e6a65c29830a31faed3d47d9973be60",
```
airkei, line 33: So far, otaclient-iot-logging-server package is not released into PyPI yet. This will be future task.

samples/proxy_info.yaml
```diff
@@ -7,3 +7,4 @@ local_ota_proxy_listen_addr: 127.0.0.1
 local_ota_proxy_listen_port: 8082
 # if otaclient-logger is installed locally
 logging_server: "http://127.0.0.1:8083"
+logging_server_grpc: "http://127.0.0.1:8084"
```
airkei, line 10: Once new otaclient with this PR is released, we will update `proxy_info.yaml` in each project(main target will be X2).

src/otaclient/_logging.py
```diff
@@ -19,32 +19,181 @@
...
+                self.format(record) if _log_type == LogType.LOG else record.getMessage()
+            )
+            self._queue.put_nowait(QueueData(_log_type, _message))
+
+    def _wait_for_log_server_up(
+        self, logging_upload_endpoint: AnyHttpUrl, ecu_id: str
+    ) -> None:
+        """Wait for the logging server to be up.
+
+        This is a blocking call. It will keep trying to connect to the server until it succeeds.
```
timothylin, line 158: what happens if it does not succeed?

airkei: Good point, wait the success forever :rocket: 
This is the dedicated logging thread, so will not affect to other tasks.
And the later log transmission will be failed even if we exit this while loop by timeout. Thus I think it can wait for the server up without timeout.

timothylin: As long as it does not cause CPU spikes with the infinite retries :)

airkei: Yes, there is `sleep` in this loop. This will prevent to occupy the CPU resource.
When the server is down,  the behavior will be the following.
`POST` -> 1sec sleep(dispatch the resource to other tasks) -> `POST` -> sleep

src/otaclient/_logging.py
```diff
@@ -19,32 +19,181 @@
...
+    """Queue data format for logging."""
+
+    log_type: LogType
+    message: str
+
+
+class Transmitter(ABC):
+    def __init__(self, logging_upload_endpoint, ecu_id: str):
+        self.ecu_id = ecu_id
+
```
Bodong-Yang, line 59: Here I can see that the intention is,  the `ecu_id` will be directly assign to the instance, while the `logging_upload_endpoint` will be parsed by each detailed implementation of `Transmitter`.
But it looks strange to me that, the `__init__` of an ABC base class has partial implementation.

Here I recommend that, we just set the `__init__` as `abstractmethod` also, and this `abstractmethod` is to declare that, the `Transmitter` implementation will take `logging_upload_endpoint` and `ecu_id` args, and completely leave the actual implementation to the detailed subclass.

src/otaclient/_logging.py
```diff
@@ -19,32 +19,181 @@
...
+        log_upload_endpoint = parsed_url.netloc
+        channel = grpc.insecure_channel(log_upload_endpoint)
+        self._stub = log_v1_grpc.OTAClientIoTLoggingServiceStub(channel)
+
+    def send(self, log_type: LogType, message: str, timeout: int) -> None:
+        pb2_log_type = (
+            log_pb2.LogType.METRICS
+            if log_type == LogType.METRICS
+            else log_pb2.LogType.LOG
+        )
```
Bodong-Yang, line 83: Inline if else here is OK, but implementing a `convert` method on `otaclient._logging.LogType` Enum is also a good idea.

Bodong-Yang (commented):
The changes generally look good! I have left some comments for some minor issues, please take a look.

src/otaclient/_logging.py
airkei: Good catch. Fixed.

src/otaclient/_logging.py
airkei: You are right, complete interface will be better than abstract class in this case. Fixed.

pyproject.toml
airkei: Good point.
As long as it is for internal use only, I think we don't have to publish.
In future, if ota-client and it's logger is used by someone outside of TIER IV, I think it's better to register into PyPI as open source project.

Bodong-Yang: ( Oh, that would be a very very future topic. I got your point :smile: .

Bodong-Yang (approved):
LGTM

PR #476, opened 2025-01-15
feat: implement Update metrics logging

airkei:
### Why
Currently, there is no metrics for ota-client

### What
Implement metrics logging for `Update` API.

## Check list

- [x] local test is passed.
https://tier4.atlassian.net/wiki/spaces/OTA/pages/3481894920/OTA+Metrics

Verified that the expected metrics log was output into [CloudWatch](https://ap-northeast-1.console.aws.amazon.com/cloudwatch/home?region=ap-northeast-1#logsV2:log-groups/log-group/$252Faws$252Fgreengrass$252Fedge$252Fap-northeast-1$252F947232557913$252Ffms-prd-edge-otaclient-metrics/log-events/2025$252F05$252F14$252Ffms-prd-edge-d0c63b12-37b3-45d4-a10e-e2a149c9dbd1-Core$252Fautoware) in E2E test.
image

## Changes
Implement metrics logging for `Update` API. The metrics logging works in ota-core process.

## Behavior changes

- [ ] No.

## Breaking change

- [ ] Yes.
- [x] No.

## Related links & tickets
https://tier4.atlassian.net/browse/RT4-13461

### DIFF
--- a/src/otaclient/metrics.py
+++ b/src/otaclient/metrics.py
@@ -19,6 +19,7 @@ import logging
 from dataclasses import asdict, dataclass
 
 from _otaclient_version import __version__
+
 from otaclient._logging import LogType
 from otaclient.configs.cfg import (
     ecu_info,
--- a/src/otaclient/metrics.py
+++ b/src/otaclient/metrics.py
@@ -19,7 +19,6 @@ import logging
 from dataclasses import asdict, dataclass
 
 from _otaclient_version import __version__
-
 from otaclient._logging import LogType
 from otaclient.configs.cfg import (
     ecu_info,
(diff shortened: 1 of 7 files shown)


src/otaclient/ota_core.py
```diff
@@ -742,15 +777,18 @@ def _process_persistents(self, ota_metadata: OTAMetadata):
...
                 payload=OTAUpdatePhaseChangeReport(
                     new_update_phase=UpdatePhase.FINALIZING_UPDATE,
-                    trigger_timestamp=int(time.time()),
+                    trigger_timestamp=_current_time,
                 ),
                 session_id=self.session_id,
             )
         )
+        self._metrics.update(finalizing_update_start_timestamp=_current_time)
+        self._metrics.publish()
```
airkei, line 791: publish the metrics to CloudWatch before rebooting.

src/otaclient/ota_core.py
```diff
@@ -783,6 +821,7 @@ def execute(self) -> None:
             self._boot_controller.on_operation_failure()
             raise ota_errors.ApplyOTAUpdateFailed(_err_msg, module=__name__) from e
         finally:
+            self._metrics.publish()
```
airkei, line 824: publish the metrics in error cases too.

src/otaclient/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+
+        # Metrics
+        ota_image_total_files_size: int = 0
+        ota_image_total_regulars_num: int = 0
+        ota_image_total_directories_num: int = 0
+        ota_image_total_symlinks_num: int = 0
+        delta_download_files_num: int = 0
+        delta_download_files_size: int = 0
+        downloaded_bytes: int = 0
+        downloaded_errors: int = 0
```
airkei, line 71: metrics items.

src/otaclient/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+    def publish(self):
+        """
+        Publishes the metrics data to the metrics server.
+        """
+        if self._already_published:
+            # metrics data has already been published.
+            return
+
+        logger.info(json.dumps(asdict(self.data)), extra={"log_type": LogType.METRICS})
+        self._already_published = True
```
airkei, line 98: outputs as json.

src/otaclient/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def publish(self):
+        """
+        Publishes the metrics data to the metrics server.
+        """
+        if self._already_published:
+            # metrics data has already been published.
+            return
+
+        logger.info(json.dumps(asdict(self.data)), extra={"log_type": LogType.METRICS})
```
airkei, line 97: outputs as json.

src/otaclient/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+            else:
+                logger.warning(f"Key {key} is not found in metrics_data.")
+
+    def publish(self):
+        """
+        Publishes the metrics data to the metrics server.
+        """
+        if self._already_published:
+            # metrics data has already been published.
+            return
```
airkei, line 95: ensure the metrics will be published once for each session.

src/otaclient/ota_core.py
```diff
@@ -269,15 +275,17 @@ def _handle_upper_proxy(self) -> None:
 
     def _process_metadata(self) -> None:
         """Process the metadata.jwt file and report."""
+        _current_time = int(time.time())
```
timothylin, line 278: is this guaranteed to be UTC time?

airkei: Yes. Python's `time.time()` return UNIX time. And we don't have to care about the timezone in UNIX time.
From UNIX time, we can convert it to readable time with timezone information.

timothylin (dismissed):
:+1:

src/otaclient/ota_core.py
timothylin: I just saw the screenshot in the PR. That makes a lot of sense now :smile:

src/otaclient/metrics.py
```diff
@@ -0,0 +1,98 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+
+class OTAMetrics:
+    @dataclass
+    class OTAMetricsData:
+        """
+        Dataclass for storing metrics data.
```
Bodong-Yang, line 35: Here we don't need to wrap `OTAMetricsData` within `OTAMetrics`. For dataclasses, we can use other method to label the `OTAMetricsData` without tempering the data fields.

Suggest the following changes:

```python
@dataclass
class OTAMetricsData:
    """
    Dataclass for storing metrics data.
    These data are collected during the OTA update process, converted to JSON, and expected to be used in subscription filters.
    Thus, flatten structure is used for easy access.
    """

    # data fields definition
    ...

    def __post_init__(self):
        # this variable will not be included in data fields
        self._already_published = False

    def publish(self) -> None:
        """
        Publishes the metrics data to the metrics server.
        """
        if self._already_published:
            return

        logger.info(json.dumps(asdict(self)), extra={"log_type": LogType.METRICS})
        self._already_published = True
```

Bodong-Yang: Also with the above change, we don't need to define dedicated `update` method anymore, just directly do variable assignment is enough.

src/otaclient/metrics.py
```diff
@@ -0,0 +1,85 @@
...
+    ota_image_total_symlinks_num: int = 0
+    delta_download_files_num: int = 0
+    delta_download_files_size: int = 0
+    downloaded_bytes: int = 0
+    downloaded_errors: int = 0
+
+    def __init__(self):
+        # this variable will not be included in data fields
+        self._already_published = False
+
```
Bodong-Yang, line 75: Should be `__post_init__`, not `__init__` :pray: .
Manual defined `__init__` will conflict with the dataclass generated `__init__`.

airkei: Good catch! Fixed.

src/otaclient/metrics.py
airkei: Thank you. Yes, `publish` is enough simple, and we can avoid key/value verification. Fixed based on your advice.

Bodong-Yang (approved):
LGTM!

PR #477, opened 2025-01-21
feat(otaclient_release): add Dockerfile for otaclient application image

Bodong-Yang:
## Introduction

This PR adds dockerfile for creating otaclient application docker image, which is for further creating otaclient squashfs application image. A README.md is also added containing the steps to create zstd compressed otaclient application squashfs image.

https://tier4.atlassian.net/browse/RT4-14061


docker/app_img/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+ARG OTACLIENT_INSTALLATION="/otaclient"
+
+COPY --from=image_build ${OTACLIENT_INSTALLATION} ${OTACLIENT_INSTALLATION}
+
+ENV OTACLIENT_VERSION=${OTACLIENT_VERSION}
+ENV OTACLIENT_INSTALLATION=${OTACLIENT_INSTALLATION}
+ENV OTACLIENT_VENV=${OTACLIENT_INSTALLATION}/venv
+
+RUN set -eux; \
+    ln -s ${OTACLIENT_VENV}/bin/python3 /otaclient_py3; \
```
airkei, line 53: what's the motivation to create symbolic link?

docker/app_img/Dockerfile
```diff
@@ -0,0 +1,64 @@
...
+		\) -exec rm -rf '{}' + \
+	; \
+	\
+    rm -rf \
+        /root/.cache \
+        /tmp/* \
+        /var/lib/apt/lists/* \
+        /var/tmp/*
+
+FROM ${UBUNTU_BASE}
```
airkei, line 41: Why we need multiple stage?(To download python on venv?)
I wonder, if we need to use venv on the docker.

docker/app_img/Dockerfile
```diff
@@ -0,0 +1,64 @@
+ARG UBUNTU_BASE="ubuntu:22.04"
+
+FROM ${UBUNTU_BASE} AS image_build
+
+ARG OTACLIENT_WHL
+COPY ./${OTACLIENT_WHL} /tmp
```
airkei, line 6: Can I change this part to create whl by hatch in Dockerfile in the later PR?

airkei:
It will be better to create test docker images based on this dockerfile.
As next steps, I will change the existing docker-compose, test Dockerfiles, and `release.yml` to create squashfs. Is this same as your expectations?

docker/app_img/Dockerfile
Bodong-Yang: In Dockerfile, we cannot use environmental variable in CMD. I don't want to hardcode python interpreter paths of virtual env into CMD, so I use a symlink instead.

docker/app_img/Dockerfile
Bodong-Yang: In the first stage, the virtual environment is boostrapped with `python3-pip` and `python3-ven` package. 
In the second stage, we only need `python3` package, then the otaclient in the virtual environment can run as expected.

Using virtual env is for making otaclient installation portable and controllable.

Bodong-Yang:
> It will be better to create test docker images based on this dockerfile.

I think it is not so neccessary. This dockerfile is especially optimized for releasing, so I use many measures, including cleanup pycache, multiple-stages build, limit the installed packages number, etc. 
These measures are not needed for docker base image for testing. 

Also, the high prior related to test is not test Dockerfile, but is to rewrite and update many old tests of otaclient.

Bodong-Yang:
> `release.yml` to create squashfs

Yes we are going to build the squashfs with github action and add it to the release :+1:

Bodong-Yang:
@airkei (A little off-topic here) Also I think we can start to apply the otaclient release format to our release, the `manifest.jwt` and all the packages(squashfs for different architecture, update patches) can be generated by github action and push as release assets.

docker/app_img/Dockerfile
airkei: Thank you. You mean we will need to replace the final line?
```
- CMD ["sh", "-c", "${OTACLIENT_VENV}/bin/python3 -m otaclient"]
+ CMD ["/otaclient_py3", "-m", "otaclient"]
```

Bodong-Yang: Oh, my fault, the symlink method is actual not working. I update the CMD but forgot to remove the symlink creation line.

docker/app_img/Dockerfile
airkei: Thank you, understood. The original motivation is to use same environment to avoid issues, especially related to CPU architecture difference between whl and docker for arm.
I will try to separate the steps and use same environment.

airkei:
> (A little off-topic here) Also I think we can start to apply the otaclient release format to otaclient release here, the manifest.jwt and all the packages(squashfs for different architecture, update patches) can be generated by github action and push as release assets.

True. The problem is that private keys / cert are necessary to create `manifest.jwt`. And Github actions can't access these secrets :( So we can create squashfs in Github actions, but will need to create `manifest.jwt` on AWS like codepipeline.

Bodong-Yang:
> True. The problem is that private keys / cert are necessary to create manifest.jwt.

Ah we can only create the payload(the actual contents of the manifest.jwt) here, and let the AWS pipeline use the manifest.json to create manifest.jwt.

PR #480, opened 2025-01-28
chore: add CODEOWNERS

airkei:
Set @tier4/ota team as code owner.


Bodong-Yang (approved):
LGTM!

By the way, in the future I think we can increase the granularity of code owner assignment, i.e., further specify the code owner of each components in otaclient repo.

airkei:
> By the way, in the future I think we can increase the granularity of code owner assignment, i.e., further specify the code owner of each components in otaclient repo.

Yes, we can change the owner per each component. This is just first step.
I imagine we might need to exclude some directories for auto merge. Anyway, let's change it as necessary.

PR #482, opened 2025-01-31
feat: integrate new ota_metadata implementation for legacy OTA image

Bodong-Yang:
## Introduction

> [!NOTE]
> This PR follows #481.

This PR integrates the new legacy OTA image metadata implementation from #481 into otaclient. 
The `create_standy` package is written with utilizing the sqlite3 based filetable implementation from #481 .
The new implementation of how the OTA metadata is downloaded is also integrated into `ota_core` module.


src/otaclient/ota_core.py
```diff
@@ -405,11 +402,62 @@ def _process_persistents(self, ota_metadata: ota_metadata_parser.OTAMetadata):
...
+                            "failed to download metadata", module=__name__
+                        ) from _exc
+
+                    if _response.status_code in [
+                        HTTPStatus.FORBIDDEN,
+                        HTTPStatus.UNAUTHORIZED,
+                    ]:
+                        raise ota_errors.UpdateRequestCookieInvalid(
+                            module=__name__
+                        ) from _exc
```
Bodong-Yang, line 452: Previously when ota_metadata package does the metadata files downloading, we handle 404 and 403 errors there.
Now the ota_core takes the responsibility of handling 404 and 403 during metadata downloading, and we map the corresponding HTTP errors to OTA errors.

src/otaclient/ota_core.py
```diff
@@ -245,74 +246,67 @@ def __init__(
...
-        self._status_report_queue.put_nowait(
-            StatusReport(
-                payload=SetUpdateMetaReport(
-                    total_download_files_num=len(delta_bundle.download_list),
-                    total_download_files_size=delta_bundle.total_download_files_size,
-                    total_remove_files_num=len(delta_bundle.rm_delta),
-                ),
-                session_id=self.session_id,
-            )
+    def _download_file(self, entry: DownloadInfo) -> tuple[int, int, int]:
```
airkei, line 256: Let's define the `@dataclass` for output.
it's more clear than `tuple[int, int, int]:`.

src/otaclient/ota_core.py
```diff
@@ -321,14 +315,20 @@ def _download_file(
...
-
             for _done_count, _fut in enumerate(
-                _mapper.ensure_tasks(_download_file, download_list), start=1
+                _mapper.ensure_tasks(
+                    self._download_file,
+                    resource_meta.iter_resources(
+                        batch_size=cfg.MAX_CONCURRENT_DOWNLOAD_TASKS
+                    ),
+                ),
+                start=1,
```
airkei, line 331: seems `start=1` is unnecessary.

airkei (commented):
Finished initial round review.
This is big change, thus let's add much more tests for new class, function, etc. I believe It will reduce future maintenance cost and realize aggressive change.

src/otaclient/ota_core.py
Bodong-Yang: Although it doesn't affect the result, but here we are counting the items we downloaded, so we need to start from 1.

src/otaclient/ota_core.py
Bodong-Yang: Sure! As the data here is simple, I think NamedTuple is ideal.

src/otaclient/create_standby/rebuild_mode.py
```diff
@@ -16,237 +16,266 @@
...
-        # cleanup on successful finish
-        # NOTE: if failed, the tmp dirs will not be removed now,
-        #       but rebuild mode will not reuse the tmp dirs anyway
-        #       as it requires standby_slot to be erased before applying
-        #       the changes.
-        shutil.rmtree(self._ota_tmp, ignore_errors=True)
+    def rebuild_standby(self) -> None:
+        self._process_dir_entries()
+        self._process_non_regular_files()
+        self._process_regular_files()
```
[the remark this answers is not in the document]
Bodong-Yang: Also tested in e2e way in `tests/test_otaclient/test_create_standby.py`.

src/otaclient/create_standby/_delta_gen.py
```diff
@@ -0,0 +1,282 @@
...
+)
+
+CANONICAL_ROOT = cfg.CANONICAL_ROOT
+
+SHA256HEXSTRINGLEN = 256 // 8 * 2
+DELETE_BATCH_SIZE = 512
+DB_CONN_NUMS = 2
+
+
+class DeltaGenerator:
```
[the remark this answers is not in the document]
Bodong-Yang: Currently DeltaGenerator is tested in e2e way in `tests/test_otaclient/test_create_standby.py`. 
I do want to add specific test for this class, but I think we are not in a hurry.

airkei:
> When in the future we have time, we can increase the granularity of tests, i.e., besides e2e test, we also implement tests for each classes/functions 👍
For now, e2e test at tests/test_otaclient/test_create_standby.py is kind of enough, and let's add more tests in the future.

OK, I still prefer to add test cases when we add new file/class, especially prefer test driven development.
But as you say, at least the coverage of `_delta_gen.py` is enough(91%). So let's postpone to add much more tests for now.

src/otaclient/ota_core.py
Bodong-Yang: Updated in [d30215b](https://github.com/tier4/ota-client/pull/482/commits/d30215bf0d0762a201b5f3e64d13b357828aa1ab).

airkei (approved):
LGTM, thank you!

PR #487, opened 2025-02-07
feat: create_standby.rebuild_mode: file_table at delta_src assisted delta_calculation

Bodong-Yang:
## Introduction

> [!NOTE]
> This PR follows #482 .

This PR adds feature of using file_table database of delta_source(for current rebuild_mode, the active_slot's file_table) to assist and speed up delta calculation. 
With delta_source file_table presented and using sqlite3 query, we can have a list of files from delta_source with digests that we need in the OTA image, and only scan those files. Previously, we have to fully scan the delta_source to get the files that we can re-use.

With this PR, now `create_standby.DeltaGenerator` has two variants, one is the new `DeltaGenWithFileTable`, and the old `DeltaGenFullDiskScan`.
`ota_core` will check if the delta_source's file_table is presented and valid or not, if available, use `DeltaGenWithFileTable` to calculate delta.

This PR bumps to use simple-sqlite3-orm v0.10.0.

## Tests

- [x] pytest covers the changes implemented.
- [x] local pytest passed.
- [x] OTA e2e test with VM passed.
    - [x] with file_table of active_slot presented.
    - [x] without file_table of active_slot presented(the old full disk scan mode).


src/ota_metadata/file_table/_table.py
```diff
@@ -140,6 +140,7 @@ def prepare_target(
             burst_suppressed_logger.exception(
                 f"failed on preparing {self!r}, {_rs=}: {e!r}"
             )
+            raise
```
Bodong-Yang, line 143: Fix the prepare_target only logs exception but forgets to raise the exception.

src/ota_metadata/file_table/_orm.py
```diff
@@ -26,28 +29,70 @@
...
+            FROM base.{FT_REGULAR_TABLE_NAME} AS d2
+            INNER JOIN (
+                SELECT digest
+                FROM main.{FT_REGULAR_TABLE_NAME}
+                WHERE digest != {wrap_value(EMPTY_FILE_SHA256_BYTE)}
+                GROUP BY digest
+            ) AS d1 USING(digest) ORDER BY digest
+        )
+        SELECT * FROM ranked WHERE rown <= {max_entries_per_digest};
+        """
```
Bodong-Yang, line 68: The core sqlite3 query statement to get a list of entries from active_slot's file_table that have digests presented in the OTA image, a.k.a., the files we can re-use.

src/ota_metadata/legacy2/metadata.py
```diff
@@ -76,6 +78,20 @@
...
+        return
+
+    with contextlib.suppress(Exception), contextlib.closing(
+        sqlite3.connect(f"file:{db_f}?mode=ro&immutable=1", uri=True)
+    ) as con:
+        if utils.check_db_integrity(
+            con,
+            table_name=FileTableRegularORM.orm_bootstrap_table_name,
+        ):
+            return db_f
```
Bodong-Yang, line 92: We ensure that the delta_source's file_table database is valid by checking the corresponding table(`ft_regular`) is presented and valid.

src/ota_metadata/legacy2/metadata.py
```diff
@@ -262,12 +278,6 @@ def _prepare_ota_image_metadata(
             )
             symlink_save_fpath.unlink(missing_ok=True)
 
-            # ------ post parsing ------ #
-            ft_regular_orm.orm_create_index(
-                index_name="digest_index",
-                index_keys=("digest",),
-            )
-
         logger.info(
```
Bodong-Yang, line 281: With the new `orm_bootstrap_db` API, now the indexes can be defined at defining the ORM. See `FileTableRegularORM` for more details.

src/ota_metadata/legacy2/metadata.py
```diff
@@ -356,8 +370,31 @@ def iter_dir_entries(self, *, batch_size: int) -> Generator[FileTableDirectories
...
+        self, base_file_table: StrOrPath, *, max_num_of_entries_per_digest: int = 10
+    ) -> Generator[tuple[bytes, list[FileTableRegularFiles]]]:
+        _hash, _cur = b"", []
+        with FileTableRegularORM(self.connect_fstable()) as orm:
+            for entry in orm.iter_common_by_digest(str(base_file_table)):
+                if entry.digest == _hash:
+                    # When there are too many entries for this digest, just pick the first
+                    #   <max_num_of_entries_per_digest> of them.
+                    if len(_cur) <= max_num_of_entries_per_digest:
+                        _cur.append(entry)
```
Bodong-Yang, line 390: To avoid the case of there are a lot of entries that have the same hashes, we only get the first 10 entries, this should be enough for most cases.

src/otaclient/create_standby/_delta_gen.py
```diff
@@ -125,6 +96,122 @@ def __init__(
             cfg.MAX_CONCURRENT_PROCESS_FILE_TASKS
         )
 
+        # put the empty file into copy_dst
+        (copy_dst / EMPTY_FILE_SHA256).touch()
+
```
Bodong-Yang, line 101: The empty file is excluded by sqlite3 query, so we prepare one here.

src/otaclient/ota_core.py
```diff
@@ -566,14 +581,30 @@ def _execute_update(self):
...
+                    f"file_table for active_slot({_base_ft_db}) not found/invalid, use full disk scan for delta calculation!"
+                )
+                delta_calculator = DeltaGenFullDiskScan(
+                    ota_metadata=self._ota_metadata,
+                    delta_src=Path(cfg.ACTIVE_SLOT_MNT),
+                    copy_dst=self._resource_dir_on_standby,
+                    status_report_queue=self._status_report_queue,
+                    session_id=self.session_id,
+                )
+                delta_calculator.calculate_delta()
```
Bodong-Yang, line 607: Explicitly choose which DeltaGenerator to use on whether the file_table database is available or not.

src/ota_metadata/file_table/_orm.py
```diff
@@ -26,28 +29,70 @@
...
+
+        attached_db_schema = "base"
+        stmt = f"""\
+        WITH ranked AS (
+            SELECT d2.*, ROW_NUMBER() OVER (PARTITION BY d2.digest) AS rown
+            FROM base.{FT_REGULAR_TABLE_NAME} AS d2
+            INNER JOIN (
+                SELECT digest
+                FROM main.{FT_REGULAR_TABLE_NAME}
+                WHERE digest != {wrap_value(EMPTY_FILE_SHA256_BYTE)}
```
Bodong-Yang, line 63: Explicitly exclude empty file.

src/ota_metadata/file_table/_orm.py
Bodong-Yang: @airkei Please take a careful look at this query :pray:

src/otaclient_common/downloader.py
```diff
@@ -52,6 +52,9 @@
 logger = logging.getLogger(__name__)
 
 EMPTY_FILE_SHA256 = r"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+EMPTY_FILE_SHA256_BYTE = bytes.fromhex(
+    "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
+)
```
airkei, line 57: This is same as `orm.EMPTY_FILE_SHA256_BYTE`.
Can we use the above in `_delta_gen.py` instead of the duplicated definitions?

airkei (commented):
Overall, looks good to me.
Left some minor comments.

airkei:
- we don't use `raise` at here(=continue delta calculation), is it expected?
https://github.com/tier4/ota-client/blob/cbc4c08d1d36e0164580ecd6de9c3cb5af6fd6fb/src/otaclient/create_standby/_delta_gen.py#L260-L261

Bodong-Yang:
> * we don't use `raise` at here(=continue delta calculation), is it expected?
>   https://github.com/tier4/ota-client/blob/cbc4c08d1d36e0164580ecd6de9c3cb5af6fd6fb/src/otaclient/create_standby/_delta_gen.py#L260-L261

Good catch, for delta calculation, if exception occurs for that file, it means we cannot reuse this file for special reason except hash mismatch(like the file changed during operation), in such case, we just need to log what is happening, and skip this file. 
No need to interrupt the whole delta calculation. If that file should be kept, the worst case is just we need to download it from remote.

src/otaclient_common/downloader.py
Bodong-Yang: Actually this value should not bound to detailed module, let me move it under the otaclient_common package as a shared const for the whole otaclient.

airkei (approved):
LGTM, thank you!

PR #488, opened 2025-02-17
feat: status_monitor: log otaclient status every 60s when active OTA ongoing

Bodong-Yang:
## Introduction

This PR introduces an extra background thread for status_monitor, when active OTA is ongoing, it will log the otaclient status every 60seconds.

### DIFF
--- a/src/otaclient/_status_monitor.py
+++ b/src/otaclient/_status_monitor.py
@@ -43,6 +43,8 @@ burst_suppressed_logger = get_burst_suppressed_logger(f"{__name__}.shm_push")
 
 _status_report_queue: queue.Queue | None = None
 
+LOG_OTA_PROGRESS_INTERVAL = 60  # seconds
+
 
 def _global_shutdown():
     if _status_report_queue:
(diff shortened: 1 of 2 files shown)


src/otaclient/_status_monitor.py
```diff
@@ -329,6 +331,15 @@ def _status_collector_thread(self) -> None:
             except queue.Empty:
                 time.sleep(self.min_collect_interval)
 
+    def _ota_status_logging_thread(self) -> None:
+        while True:
+            # when in active OTA, log OTA progress very <LOG_OTA_PROGRESS_INTERVAL>
+            if (
+                _cur_status := self._status
+            ) and _cur_status.ota_status == OTAStatus.UPDATING:
```
airkei, line 339: How about to  add `OTAStatus.ROLLBACKING` too?

airkei (commented):
left a minor comments.

src/otaclient/_status_monitor.py
Bodong-Yang: `OTAStatus.ROLLBACKING` is just configuring the bootloader to directly switch slot, no changes to rootfs will occur on both slots, so the progress will be very quick(succeed or fail).

Another thing is, actually ROLLBACK is not used and not officially announced. The code is just there, but it is not expected to be used.

airkei (approved):
LGTM

PR #491, opened 2025-02-21
feat: mount tmpfs for each update_session

Bodong-Yang:
## Introduction

Previously, otaclient uses `/run/otaclient` as run dir, and for each update session creates a `update_session-xxx` folder to store the file_table and resource_table, occupying the system `/run` tmpfs mount point space. 
To avoid using up the `/run` space, this PR introduces to use tmpfs for each update session.


Bodong-Yang:
Will be replaced by a new PR.

Bodong-Yang:
Superseded by #582 .

PR #492, opened 2025-02-21
fix: delta_gen: fix semaphore not being reclaimed in full disk scan mode

Bodong-Yang:
## Introduction

This PR fixes an issue which In `_delta_gen.DeltaGenFullDiskScan._process_file`, if src file is an empty file, the method will return directly without releasing the semaphore.

Tests that cover this condition has been added by adding a folder with a lot of empty files in the generated OTA image.


airkei (dismissed):
LGTM, nice test case supporting.

airkei (approved):

PR #498, opened 2025-03-14
fix(v3.8.x): rpi_boot.update_firmware: ensure `/tmp` folder exists when doing flash-kernel

Bodong-Yang:
## Introduction

Starting from otaclient v3.8.x, a new mechanism is used to replace the previous two-stage reboot mechanism for executing `flash-kernel`. This new mechanism will chroot into the standby slot, and call `flash-kernel` on standby slot directly to deploy new dtbs, kernel and initrd img.

However, current OTA image is configured to ignore with pattern `/tmp`, which means even the top-level `/tmp` folder will be ignored. When OTA just updates the standby slot, there will be no `/tmp` folder existed on it, causing `flash-kernel` failed due to `flash-kernel` needs `/tmp` folder during execution.

This PR fixes this problem by ensuring `/tmp` folder on standby slot.

Other changes:
1. now rpi_boot will wait for 6 seconds, and issue a logging before rebooting.

## Tests

- [x] test passes on directly executing the rpi_boot control flow.

### DIFF
--- a/src/otaclient/boot_control/_rpi_boot.py
+++ b/src/otaclient/boot_control/_rpi_boot.py
@@ -298,6 +298,9 @@ class _RPIBootControl:
         sys_mp = target_slot_mp / "sys"
         mounts[str(sys_mp)] = "/sys"
 
+        # NOTE(20250314): ensure that tmp folder exists on standby slot
+        (target_slot_mp / "tmp").mkdir(exist_ok=True)
+
         try:
             for _mp, _src in mounts.items():
                 CMDHelperFuncs.mount(
--- a/src/otaclient/boot_control/_rpi_boot.py
+++ b/src/otaclient/boot_control/_rpi_boot.py
@@ -335,6 +335,7 @@ class _RPIBootControl:
                     env={"FK_FORCE": "yes"},
                 )
                 os.sync()
+                logger.info("flash-kernel succeeded!")
         except subprocess.CalledProcessError as e:
             _err_msg = f"flash-kernel failed: {e!r}\nstderr: {e.stderr.decode()}\nstdout: {e.stdout.decode()}"
             logger.error(_err_msg)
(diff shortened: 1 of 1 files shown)


src/otaclient/boot_control/_rpi_boot.py
```diff
@@ -410,11 +415,15 @@ def prepare_tryboot_txt(self):
...
             raise _RPIBootControllerError(_err_msg) from e
 
-    def reboot_tryboot(self):
+    def reboot_tryboot(self, *, reboot_wait: int = 6):
         """Reboot with tryboot flag."""
-        logger.info(f"tryboot reboot to standby slot({self.standby_slot})...")
+        logger.info(
+            f"tryboot reboot to standby slot({self.standby_slot}) in {reboot_wait=}s ..."
+        )
+        time.sleep(reboot_wait)
```
airkei, line 427: (nits) I wonder, if we can use flush instead of waiting for constant time like the following code.
```
for handler in logger.handlers:
    handler.flush()
```

airkei (approved):
LGTM, thank you :tada: I understood that `flash-kernel` is used only in rasp pi.

Bodong-Yang:
@airkei What do you think above?

src/otaclient/boot_control/_rpi_boot.py
Bodong-Yang: Emmm currently our `_LogTeeHandler` at `otaclient._logging` in otaclient `v3.8.x` doesn't implement the `flush` API.

Bodong-Yang: So for v3.8.x, let's make the thing easier, just wait for 6 seconds.

And in the main branch, let's implement the `flush` API in the otaclient `LogHandler` :+1:

airkei: Yes, this is nits comment. Please feel free to ignore as necessary. And I agree with waiting for constant value for this hot-fix.

> Emmm currently our _LogTeeHandler at otaclient._logging in otaclient v3.8.x doesn't implement the flush API.

Oh..., I understood we need to implement it by ourselves. the original `logging.Handler.flush()` just pass the function...

PR #499, opened 2025-03-14
fix: rpi_boot.update_firmware: ensure /tmp folder exists when doing flash-kernel

Bodong-Yang:
This is the forward port version of fix in #498, check that PR for more details.


airkei (approved):
LGTM, thank you.

PR #502, opened 2025-03-18
fix(otaclient main branch): Add timeout when downloading "metadata.jwt" and *.pem files

Zhenfeng-Sun:
Related JIRA: [RT4-15059](https://tier4.atlassian.net/browse/RT4-15059)

### About the fix
At the beginning of OTA process, we will download "medata.jwt" file and then "certificate" pem file. 
There are cases that these files can not be downloaded correctly. 

For example:
  1, ECU network config is not correct. 
  2, The certificate file is not correct
  etc...
  
When it happens, current behavior is that "otaclient" keep retry endlessly. 
For the end user, he will not be aware of the error and will likely keep waiting for a long time. 

This fix will timeout the retrying process after 5 mins. 
End user will be informed OTA is not successful after that.

[RT4-15059]: https://tier4.atlassian.net/browse/RT4-15059?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ota_metadata/legacy/_parser.py
+++ b/src/ota_metadata/legacy/_parser.py
@@ -638,8 +638,10 @@ class OTAMetadata:
         with NamedTemporaryFile(prefix="metadata_jwt", dir=self.run_dir) as meta_f:
             _downloaded_meta_f = Path(meta_f.name)
 
-            _retry_counter = 0
-            while _retry_counter <= cfg.DOWNLOAD_INACTIVE_TIMEOUT:
+            _retry_duration = 0
+            while (not _shutdown) and (
+                _retry_duration < cfg.DOWNLOAD_INACTIVE_TIMEOUT
+            ):
                 try:
                     self._downloader.download(
                         urljoin_ensure_base(self.url_base, self.METADATA_JWT),
@@ -656,7 +658,7 @@ class OTAMetadata:
                     logger.warning(
                         f"failed to download {_downloaded_meta_f}, retrying: {e!r}"
                     )
-                    _retry_counter += 1
+                    _retry_duration += self.retry_interval
                     time.sleep(self.retry_interval)
 
             _parser = _MetadataJWTParser(
@@ -672,8 +674,10 @@ class OTAMetadata:
             cert_fname, cert_hash = cert_info.file, cert_info.hash
             cert_file = Path(cert_f.name)
 
-            _retry_counter = 0
-            while _retry_counter <= cfg.DOWNLOAD_INACTIVE_TIMEOUT:
+            _retry_duration = 0
+            while (not _shutdown) and (
+                _retry_duration < cfg.DOWNLOAD_INACTIVE_TIMEOUT
+            ):
                 try:
                     self._downloader.download(
                         urljoin_ensure_base(self.url_base, cert_fname),
@@ -700,7 +704,7 @@ class OTAMetadata:
                             raise OTARequestsAuthTokenInvalid from e
 
                     logger.warning(f"failed to download {cert_info}, retrying: {e!r}")
-                    _retry_counter += 1
+                    _retry_duration += self.retry_interval
                     time.sleep(self.retry_interval)
 
             cert_bytes = cert_file.read_bytes()
--- a/src/ota_metadata/legacy/_parser.py
+++ b/src/ota_metadata/legacy/_parser.py
@@ -639,7 +639,7 @@ class OTAMetadata:
             _downloaded_meta_f = Path(meta_f.name)
 
             _retry_counter = 0
-            while not cfg.DOWNLOAD_INACTIVE_TIMEOUT:
+            while _retry_counter <= cfg.DOWNLOAD_INACTIVE_TIMEOUT:
                 try:
                     self._downloader.download(
                         urljoin_ensure_base(self.url_base, self.METADATA_JWT),
@@ -673,7 +673,7 @@ class OTAMetadata:
             cert_file = Path(cert_f.name)
 
             _retry_counter = 0
-            while not cfg.DOWNLOAD_INACTIVE_TIMEOUT:
+            while _retry_counter <= cfg.DOWNLOAD_INACTIVE_TIMEOUT:
                 try:
                     self._downloader.download(
                         urljoin_ensure_base(self.url_base, cert_fname),
--- a/src/ota_metadata/legacy/_parser.py
+++ b/src/ota_metadata/legacy/_parser.py
@@ -77,6 +77,7 @@ from typing_extensions import Self
 
 from ota_metadata.utils.cert_store import CAChainStore
 from ota_proxy import OTAFileCacheControl
+from otaclient.configs.cfg import cfg
 from otaclient_common._typing import StrEnum
 from otaclient_common.common import urljoin_ensure_base
 from otaclient_common.downloader import Downloader
@@ -637,7 +638,8 @@ class OTAMetadata:
         with NamedTemporaryFile(prefix="metadata_jwt", dir=self.run_dir) as meta_f:
             _downloaded_meta_f = Path(meta_f.name)
 
-            while not _shutdown:
+            _retry_counter = 0
+            while not cfg.DOWNLOAD_INACTIVE_TIMEOUT:
                 try:
                     self._downloader.download(
                         urljoin_ensure_base(self.url_base, self.METADATA_JWT),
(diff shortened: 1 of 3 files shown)


airkei (commented):
- I meant, `main` branch has already migrated to `legacy2`. so this change(`legacy`) will not impact to the actual behavior(timeout will not be triggered in metadata downloading).
- When timeout is happened, what is the expected behavior? 
- Let's add unit tests

src/ota_metadata/legacy/_parser.py
```diff
@@ -637,7 +638,8 @@ def _process_metadata_jwt(self) -> _MetadataJWTClaimsLayout:
         with NamedTemporaryFile(prefix="metadata_jwt", dir=self.run_dir) as meta_f:
             _downloaded_meta_f = Path(meta_f.name)
 
-            while not _shutdown:
+            _retry_counter = 0
+            while not cfg.DOWNLOAD_INACTIVE_TIMEOUT:
```
airkei, line 642: `_retry_counter` is not used in this condition.

Zhenfeng-Sun: Oh. 
The test I did it like
１） Test against branch v3.8.x. 
my whl file is version otaclient-3.8.5xxx
It worked fine. 
I saw an exception thrown after 5 min.

2) Then I start to test against main. 
my whl file is version otaclient-3.8.2xxx
It worked fine. 
I saw an exception thrown after 5 min.

**Now I realized that I did not call a "systemctl restart otaclient" after I install otaclient-3.8.2xxx in step 2).
So, that caused the test result is wrong.** 

Let me slow down and check the main branch again.

Zhenfeng-Sun: I am following this manual to do the local test.
https://tier4.atlassian.net/wiki/spaces/OTA/pages/3567092373/Simulate+an+OTA+update+locally

src/ota_metadata/legacy2/metadata.py
```diff
@@ -163,7 +164,7 @@ def _prepare_metadata(
                     dst=_metadata_jwt_fpath,
                 )
             ]
-            condition.wait()  # wait for download finished
+            condition.wait(cfg.DOWNLOAD_INACTIVE_TIMEOUT)  # wait for download finished
```
airkei, line 167:
```suggestion
            if not condition.wait(cfg.DOWNLOAD_INACTIVE_TIMEOUT):  # wait for download finished
                    raise TimeoutError("Download of metadata.jwt timed out")
```

src/ota_metadata/legacy2/metadata.py
```diff
@@ -188,7 +189,7 @@ def _prepare_metadata(
                     digest=cert_hash,
                 )
             ]
-            condition.wait()
+            condition.wait(cfg.DOWNLOAD_INACTIVE_TIMEOUT)
```
airkei, line 192:
```suggestion
            if not condition.wait(cfg.DOWNLOAD_INACTIVE_TIMEOUT):  # wait for download finished
                    raise TimeoutError("Download of metadata.jwt timed out")
```

src/ota_metadata/legacy2/metadata.py
```diff
@@ -263,7 +264,9 @@ def _prepare_ota_image_metadata(
             ]
             with condition:
                 yield _download_list
-                condition.wait()  # wait for download finished
+                condition.wait(
+                    cfg.DOWNLOAD_INACTIVE_TIMEOUT
+                )  # wait for download finished
```
airkei, line 269:
```suggestion
                if not condition.wait(cfg.DOWNLOAD_INACTIVE_TIMEOUT):  # wait for download finished
                        raise TimeoutError("Download of metadata.jwt timed out")
```

src/otaclient_common/downloader.py
```diff
@@ -343,7 +343,6 @@ def close(self) -> None:
         """
         self._session.close()
 
-    @retry_on_digest_mismatch
```
airkei, line 346: In my understanding, `retry_on_digest_mismatch` is to detect hash mismatch(= download was succeeded, but hash was mismatched). `downloading_watchdog` is for timeout by watch dog.
In this change, my question is that what will be happened when download is succeeded but hash mismatch is occurred? Expected behavior will be to retry.

airkei (commented):
Let's add test code, it will reduce the E2E tests by human and be more efficient.

src/otaclient_common/downloader.py
Bodong-Yang: `retry_on_digest_mismatch` implements the `OTA_CACHE_CONTROL` protocol between otaclient and otaproxy.

During downloading, otaproxy will cache the requests from otaclient. If otaclient detects an incomplete/broken downloading by digest check,  `retry_on_digest_mismatch` will inject a header into the request of next downloading request, telling otaproxy to clear the previously cached file.

src/ota_metadata/legacy2/metadata.py
Bodong-Yang: I think add `timeout` to `condition.wait` is unintended. 
Here it means that the download for each file MUST be finished within 5mins, so it will unintentionally fail the downloading if the network is just very slow.

I think we should still set the timeout logic as when network is fully disconnected and there is not network traffic for `DOWNLOAD_INACTIVE_TIMEOUT`.

Bodong-Yang:
https://github.com/tier4/ota-client/blob/51003331f07bc6eeaed084019869350cbf03b6f3/src/otaclient/ota_core.py#L329-L351

Emmmm it seems like I have already tried to implement timeout logic for metadata downloading, we are already using `ThreadPoolExectorWithRetry` here. So the problem is the timeout is not expected(like longer than 5mins)?

Bodong-Yang:
Emmmm do a very quick test, for network fully disconnected case, the timeout logic is already working:
(For faster test I set the timeout to 20s)
```
[2025-03-21 11:21:39,513][ERROR]-otaclient_common.downloader:downloading_watchdog:517,downloader stuck for 20 seconds, abort
[2025-03-21 11:21:39,513][WARNING]-otaclient_common.retry_task_map:_watchdog:121,watchdog failed: ValueError('downloader stuck for 20 seconds, abort'), shutdown the pool and draining the workitem queue on shutdown..
[2025-03-21 11:21:39,555][WARNING]-otaclient_common.retry_task_map:_fut_gen:182,dispatcher exits on threadpool shutdown, finished_tasks=0, self._total_task_num=0
[2025-03-21 11:21:40,291][ERROR]-otaclient.ota_core:_process_metadata:318,failed to prepare ota metafiles: ValueError('generator already executing')
[2025-03-21 11:21:40,291][ERROR]-otaclient.ota_core:execute:777,update failed: OTAMetaDownloadFailed("failed to prepare ota metafiles: ValueError('generator already executing')")
[2025-03-21 11:21:40,291][WARNING]-otaclient.boot_control._grub:on_operation_failure:868,on failure try to unmounting standby slot...
[2025-03-21 11:21:40,359][ERROR]-otaclient.ota_core:_on_failure:888,E101: failed to download OTA meta due to network unstable, please check the network connection
[2025-03-21 11:21:40,359][ERROR]-otaclient.ota_core:_on_failure:889,last error traceback: 
```

Here there is another unrelated bug though, caused by:
https://github.com/tier4/ota-client/blob/51003331f07bc6eeaed084019869350cbf03b6f3/src/otaclient/ota_core.py#L380-L382

I am thinking of throw an exception to interrupt the generator, but the ota_metadata.process_metadata generator is still running(condition.wait()).
generator throw only works when the generator stops at yield statement.

I will create another PR for fixing the above issue.

Bodong-Yang:
Also, for the case of digest mismatch, currently we will just do infinitely retrying on this case as digest mismatch is not network disconnected.

Digest mismatch mostly happens on in-complete downloading, in very rare case it is caused by a broken OTA image.

We can simply configure `ThreadPoolExecutorWithRetry` with a reasonable `max_retry_count` value to implement breaking out on digest mismatch error.

Bodong-Yang:
So I think this PR can be closed, and what we need to do are as follow:

1. open a new PR to configure `max_retry_count` for OTA metadata files downloading.
2. open a new PR to fix the generator issue when breaking out the ota_metadata files downloading.

Zhenfeng-Sun:
The solution is not good. Let me close this PR

PR #503, opened 2025-03-24
fix: otaproxy_ctx: fix otaproxy life-cycle management thread exits at OTA failed

Bodong-Yang:
## Introduction

By design, otaproxy life-cycle management thread should keep running at the background, start otaproxy when there is at least one ECU requires network, and shutdown otaproxy when there is no ECU requires network. 
When OTA failed and another OTA retry triggered, otaproxy ctx should start otaproxy again. However, due to otaproxy ctx thread calls **global_shutdown** when we need to close otaproxy, otaproxy ctx unexpectedly shutdown itself, and can no-longer start otaproxy anymore.

This PR fixes the above issue, only do otaproxy shutdown instead of global_shutdown when we don't need otaproxy for now, ensure that otaproxy ctx still keeps running at the background, and start/stop otaproxy in next time's OTA retry.

## Tests

- [x] test on VM: ensure that otaproxy ctx thread will keep running when OTA failed and start otaproxy at next OTA.

### DIFF
--- a/src/otaclient/_otaproxy_ctx.py
+++ b/src/otaclient/_otaproxy_ctx.py
@@ -24,6 +24,7 @@ import logging
 import multiprocessing as mp
 import multiprocessing.context as mp_ctx
 import shutil
+import threading
 import time
 from functools import partial
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


airkei (approved):
LGTM.
In my understanding, when multiple threads call global shutdown simultaneously, it might lead to unexpected behavior. It prevents to wake up otaproxy for next OTA requests.
This change introduce graceful shutdown of otaproxy by using thread lock, and realize otaproxy rewakeup for next time.

PR #507, opened 2025-03-26
tests: refine tests for retry_task_map

Bodong-Yang:
## Introduction

Previously, we test retry_task_map breaks out on exceeding max retry limit with doing assertion of failure_count also exceeds on main thread. 
This strategy has some problem, as retry_task_map internal breakout happens faster than retry_task_map thread pool delivers the result to the main thread, resulting in assertion failed on main thread sometime.

This PR refines the implementation of tests, now we track the failure_count by each worker thread, so that we can do accurate checking over the failure_count.


airkei (approved):
LGTM, thank you :+1:

PR #508, opened 2025-03-26
fix: ota_core: Add timeout for when downloading metadata file

Zhenfeng-Sun:
### Related JIRA:
https://tier4.atlassian.net/browse/RT4-15059

### What is the fix:
When downloading OTA metadata files. (metadata.jwt, certificate.pem, regulars.txt, dirs.txt, symlinks.txt, persistents.txt), we might encounters some errors.
Current behavior of main branch is as following:
1, When "Connection Error", download will timeout in 5 minutes. (Good Behavior)
2, When “Hash Verification Error“, download will retry forever (NOT Good behavior)

This fix will fix 2, and make it timeout in 5 minutes.

### DIFF
--- a/src/otaclient/ota_core.py
+++ b/src/otaclient/ota_core.py
@@ -341,6 +341,7 @@ class _OTAUpdateOperator:
         _mapper = ThreadPoolExecutorWithRetry(
             max_concurrent=cfg.MAX_CONCURRENT_DOWNLOAD_TASKS,
             max_workers=cfg.DOWNLOAD_THREADS,
+            max_retry_on_entry=cfg.DOWNLOAD_INACTIVE_TIMEOUT,
             thread_name_prefix=thread_name_prefix,
             initializer=self._downloader_workder_initializer,
             watchdog_func=partial(


src/otaclient/ota_core.py
```diff
@@ -341,6 +341,7 @@ def _download_and_process_file_with_condition(
         _mapper = ThreadPoolExecutorWithRetry(
             max_concurrent=cfg.MAX_CONCURRENT_DOWNLOAD_TASKS,
             max_workers=cfg.DOWNLOAD_THREADS,
+            max_retry_on_entry=cfg.DOWNLOAD_INACTIVE_TIMEOUT,
```
airkei, line 344: seems the expected unit of `max_retry_on_entry` parameter is failure "count", not "sec". Is it expected to use this variable?

Zhenfeng-Sun: I referred to the usage of "watchdog_func" part. I thought it is fine that it will retry every second. 
What do  you suggest ?

Zhenfeng-Sun: Another super minor issue. Shall I fix this typo with this PR also ? 
image

src/otaclient/configs/_cfg_configurable.py
```diff
@@ -57,6 +57,9 @@ class _OTAClientSettings(BaseModel):
     DOWNLOAD_THREADS: int = 6
     MAX_CONCURRENT_DOWNLOAD_TASKS: int = 128
     DOWNLOAD_INACTIVE_TIMEOUT: int = 5 * 60  # seconds
+    MAX_RETRY_ON_ENTRY_COUNT: int = (
+        300  # counts. In case of retry every second, it will be 5 minutes
+    )
```
airkei, line 62: Just comment, I think 300 counts is OK. But to be exact, this is not 5 minutes. 
[exponential backoff](https://en.wikipedia.org/wiki/Exponential_backoff) is used for retry waiting time.  Thus, retry interval will be...
```
Retry 1: 0.01 seconds
Retry 2: 0.02 seconds
Retry 3: 0.04 seconds
Retry 4: 0.08 seconds
Retry 5: 0.16 seconds
Retry 6: 0.32 seconds
Retry 7: 0.64 seconds
Retry 8: 1.0 seconds
Retry 9: 1.0 seconds
Retry 10: 1.0 seconds
```

airkei (approved):
LGTM, thank you :+1:

src/otaclient/configs/_cfg_configurable.py
airkei: - 300 times -> 293.59 seconds
- 306 times -> almost 300 seconds

But I think 300 counts is OK. Might be better to change the comments like "almost 5 minutes".

Zhenfeng-Sun: OK. let me change the comment. I need to re-trigger the unit test again anyway. It is unstable

airkei (approved):

Zhenfeng-Sun:
All tests passed and I did a final test locally also. let me merge

PR #509, opened 2025-03-26
refactor: encapsulate data of ecu status

airkei:
### Why
For ota client update, need to pass ECU status from the previous process to next process. So better to encapsulate data for process communication.

### What
Create new  ` _ECUStatusState` class and encapsulate necessary data.

### Tests
Confirmed to pass unit tests.

### DIFF
--- a/src/otaclient/grpc/api_v2/ecu_status.py
+++ b/src/otaclient/grpc/api_v2/ecu_status.py
@@ -43,6 +43,7 @@ import asyncio
 import logging
 import math
 import time
+from dataclasses import dataclass, field
 from itertools import chain
 
 from otaclient._types import MultipleECUStatusFlags, OTAClientStatus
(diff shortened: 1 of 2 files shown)


Zhenfeng-Sun (approved):

Bodong-Yang:
@airkei Split a new type for holding data is good! 

Just some nits, since we need to pass the object across process, we need to serialize the object. Which serialization method you plan to use(like pickle, or dump as JSON, etc.)? 

If using pickle, could you add test to confirm it can be pickled?

airkei:
> Just some nits, since we need to pass the object across process, we need to serialize the object. Which serialization method you plan to use(like pickle, or dump as JSON, etc.)?
If using pickle, could you add test to confirm it can be pickled?

@Bodong-Yang That's good point, because I'm still considering which format and communication way is better in this case.
- format

| | pickle  | json |
| ------------- | ------------- | ------------- |
| serialize | easy  | complicated(need to serialize `StatusResponseEcuV2` by ourselves)  |
| debuggability | low | high(can read)  |
| security | low | high  |
| size | good(binary)  | not good(string)  |

- process communication way

| | environment variable  | pipe |
| ------------- | ------------- | ------------- |
| implementation | easy  |  a bit complicated  |
| supportable format | only json | both pickle and json  |
| security | low | high  |
| restriction | there is maximum string length | /  |

I'm considering to use "json" and "pipe". Do you have any preferences?

Bodong-Yang:
> I'm considering to use "json" and "pipe". Do you have any preferences?

@airkei Agree! Then let's implement the JSON serialization/deserialization in another PR :+1: This PR is ready to go.

Bodong-Yang (approved):

PR #510, opened 2025-03-26
refactor: aggregate duplicated code in servicer

airkei:
### Why
The motivation is to add client update command. Before that, organize and remove duplicated code in servicer.

### What
Aggregate the duplicated code for update and rollback in servicer.

### Tests
Confirmed to pass the unit tests.

### DIFF
--- a/src/otaclient/grpc/api_v2/servicer.py
+++ b/src/otaclient/grpc/api_v2/servicer.py
@@ -191,6 +191,9 @@ class OTAClientAPIServicer:
         logger.info(f"receive request: {request}")
         response = response_type()
 
+        # NOTE(20241220): due to the fact that OTA Service API doesn't have field
+        #                 in UpdateResponseEcu msg, the only way to pass the failure_msg
+        #                 to upper is by status API.
         if not otaclient_cfg.ECU_INFO_LOADED_SUCCESSFULLY:
             logger.error("ecu_info.yaml is not loaded properly, reject any request")
             for _req in request.iter_ecu():
@@ -237,19 +240,19 @@ class OTAClientAPIServicer:
             tasks.clear()
 
         # second: dispatch update request to local if required by incoming request
-        if update_req_ecu := request.find_ecu(self.my_ecu_id):
-            if isinstance(update_req_ecu, api_types.UpdateRequestEcu) and (
+        if req_ecu := request.find_ecu(self.my_ecu_id):
+            if isinstance(req_ecu, api_types.UpdateRequestEcu) and (
                 request_cls is UpdateRequestV2
             ):
                 # update
-                new_session_id = gen_session_id(update_req_ecu.version)
+                new_session_id = gen_session_id(req_ecu.version)
                 local_request = request_cls(
-                    version=update_req_ecu.version,
-                    url_base=update_req_ecu.url,
-                    cookies_json=update_req_ecu.cookies,
+                    version=req_ecu.version,
+                    url_base=req_ecu.url,
+                    cookies_json=req_ecu.cookies,
                     session_id=new_session_id,
                 )
-            elif isinstance(update_req_ecu, api_types.RollbackRequestEcu) and (
+            elif isinstance(req_ecu, api_types.RollbackRequestEcu) and (
                 request_cls is RollbackRequestV2
             ):
                 # rollback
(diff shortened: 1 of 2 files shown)


Zhenfeng-Sun (dismissed):

Bodong-Yang:
@airkei The idea is greate, but the implementation needs to be polished. Please consider how to properly type the newly added aggregated functions :pray:

Bodong-Yang:
Suggest that instead of providing `request_cls, local_handler, request_cls, remote_call, response_type, response_type, response_ecu_type` as args, what about just checking the type of `request`, and let the `_handle_request` decides what to do.

Something like:
```python
  def _handle_request(self, request: api_types.UpdateRequest | api_types.RollbackRequest):
        if isinstance(request, api_types.UpdateRequest):
            local_handler=self._local_update,
            request_cls=UpdateRequestV2,
            remote_call=OTAClientCall.update_call,
            response_type=api_types.UpdateResponse,
            response_ecu_type=api_types.UpdateResponseEcu,
        else:
            ...
```

By this the code will be simple and readability will be improved.

airkei:
@Bodong-Yang Thank you for your feedback.
I understood how it is difficult to aggregate the methods of the different inputs/outputs in Python... Especially in the combination of static code analysis and dynamic typing language :disappointed:

In the latest commit, introduced fixes.
- set `update_acked_ecus` only for `update`. It means main ECU will not wait for others in `rollback`.
- added overload and type hint for each function
- for overload and type hint, let me pass `request_cls` as arg instead of judging from `request`.

airkei:
Added test file for servicer.
- update
Verified that `on_ecus_accept_update_request` is called.
https://github.com/tier4/ota-client/blob/12377ae5a32fd6a5b29bcb47121ce18fd4cb3dda/tests/test_otaclient/test_grpc/test_api_v2/test_servicer.py#L260

- rollback
Verified that `on_ecus_accept_update_request` is not called.
https://github.com/tier4/ota-client/blob/12377ae5a32fd6a5b29bcb47121ce18fd4cb3dda/tests/test_otaclient/test_grpc/test_api_v2/test_servicer.py#L435

Zhenfeng-Sun (approved):

Bodong-Yang:
> I understood how it is difficult to aggregate the methods of the different inputs/outputs in Python... Especially in the combination of static code analysis and dynamic typing language 😞

@airkei 

This time the typing is hard due to:
1. Container message like `UpdateResponse` and its contained type `UpdateResponseEcu` currently are not related to each other from python typing point of view. Extra typings are needed for these two types(like `list[str]`, announcing that the container type has `str` as element type).
2. `UpdateResponse` and `RollbackResponse`, they behave similarly (both have same helper methods), but these helper methods takes different contained type, so we still need to narrow down the types before we can typing the methods call.
3. I also think of using `overload` to type the aggregated function(`_handle_request`), but the limitation is it only checks the input of function call, inside the function itself we still need to manually reveal the type(with `isinstance`) to make type checker happy.

You have already done a good job, changes of PR is the best we can do to the typing for now.

Bodong-Yang:
( Also due to how python typing works, actually providing `request_cls, local_handler, request_cls, remote_call, response_type, response_type, response_ecu_type` as args has no difference to using `isinstance` over `request` and at runtime from typing point of view. 
For both case, inside `_handle_request`, you still need to reveal the type from Union.

Bodong-Yang (approved):

airkei:
@Bodong-Yang Thank you for your feedback! I also tried to pass only `request` as arg and use `isinstance` internally based on your adivce. In this case, caller side complains for the ambiguous **output** :cry:
branch: https://github.com/tier4/ota-client/commit/d25f817840f338b342e1864ebac2471212b78f9b
```
Type "UpdateResponse | RollbackResponse" is not assignable to return type "UpdateResponse"
  Type "UpdateResponse | RollbackResponse" is not assignable to type "UpdateResponse"
    "RollbackResponse" is not assignable to "UpdateResponse"Pylance[reportReturnType](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportReturnType)
(method) def _handle_request(request: UpdateRequest | RollbackRequest) -> CoroutineType[Any, Any, UpdateResponse | RollbackResponse]
```

image

Bodong-Yang:
>  caller side complains for the ambiguous output 😢
@airkei  Oh since we type `_handle_request` return value as Union, we need to narrow it down to one of the type:

```python
async def update(self, request: api_types.UpdateRequest) -> api_types.UpdateResponse:
    # here we know that for update request, we will get UpdateResponse from _handle_request, so type hint the res as UpdateResponse to narrow down the Union to concrete type
    res: api_types.UpdateResponse = await self._handle_request(request=request)
    return res
```

PR #511, opened 2025-03-27
fix: otaclient :  fix vulnerability reported by Synk

Zhenfeng-Sun:
### Related JIRA
https://tier4.atlassian.net/browse/RT4-15378

### About this Fix
As the vulnerability report suggested, we change to use cryptography@44.0.1
image


airkei:
How about to use `44.0.2` instead? `44.0.2` is the latest in the original range, and no breaking changes.
https://cryptography.io/en/latest/changelog/

Bodong-Yang (approved):

PR #514, opened 2025-03-28
fix: ota-client : Update grpcio version 

Zhenfeng-Sun:
### Related JIRA
https://tier4.atlassian.net/browse/RT4-15432

### About This Fix
As being adviced here
https://security.snyk.io/vuln/SNYK-PYTHON-GRPCIO-9486468
image

Change grpcio to version listed here
https://pypi.org/project/grpcio/#history

### Note : 
lastest version is : 1.71.0
But it is not supported for ubuntu 18.04 and 20.04. 
So we use 1.70.0

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -27,7 +27,7 @@ dependencies = [
   "anyio>=4.5.1,<5",
   "aiohttp>=3.10.11,<3.12",
   "cryptography==44.0.2",
-  "grpcio==1.71.0",
+  "grpcio==1.70.0",
   "multidict<7.0,>=4.5",
   "msgpack>=1,<1.2",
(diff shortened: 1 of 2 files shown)


airkei (commented):
Code change is LGTM.
This is new version which we don't use before, thus suggest to do the actual OTA test by using VM and describe test result in the PR description.

Bodong-Yang:
Basically it should be OK if unit tests are passed, but as @.airkei says, for safety first, please use VM to confirm this grpc version is working.

Zhenfeng-Sun:
Very bad. Seems we can not update after the update
image

Bodong-Yang (approved):

airkei (approved):

PR #516, opened 2025-03-28
feat: add ecu status serialization by pickle

airkei:
### Why
in otaclient update, ecu status should be handover between process.

### What
This PR introduces the serialization of ecu status by pickel.
#### Why Pickel?

| | pickle  | json |
| ------------- | ------------- | ------------- |
| serialize | easy  | complicated(need to serialize `StatusResponseEcuV2` by ourselves)  |
| debuggability | low | high(can read)  |
| security | low | high  |
| size | good(binary)  | not good(string)  |
| maintainability | good  | too bad  |

First, try to serialize into json. But in the implementation, realized it's not good in terms of maintainability.
Because ecu status data include multiple types(int, dict, set, object like `StatusResponseEcuV2`, event flag). We need to check each type of variable then need to serialize them in proper way. When we add new variable with new type in future, we need to change the serialization method too to support the new type. it's not good in terms of maintainability.
Thus, suggest to use pickle by prioritizing maintainability.

## Check list

- [x] local test is passed.
- [ ] design docs/implementation docs are prepared.

## Behavior changes

- [x] No.

## Breaking change

- [ ] Yes.
- [x] No.

### DIFF
--- a/src/otaclient/grpc/api_v2/ecu_status.py
+++ b/src/otaclient/grpc/api_v2/ecu_status.py
@@ -42,6 +42,7 @@ from __future__ import annotations
 import asyncio
 import logging
 import math
+import pickle
 import time
 from dataclasses import dataclass, field
 from itertools import chain
@@ -64,7 +65,7 @@ ACTIVE_POLLING_INTERVAL = 1  # seconds
 
 
 @dataclass
-class _ECUStatusState:
+class ECUStatusState:
     """State container for ECUStatusStorage."""
 
     # ECU status flags
@@ -98,6 +99,55 @@ class _ECUStatusState:
     # Update request tracking
     last_update_request_received_timestamp: int = 0
 
+    def to_pickle(self):
+        """Serialize the ECUStatusState instance to a byte string using pickle.
+
+        Returns:
+            bytes: Pickled representation of the ECUStatusState instance
+        """
+        state_dict = {}
+        for key, value in self.__dict__.items():
+            if key == "ecu_status_flags":
+                # Pickle doesn't support serialization of Event objects
+                # Extract the state of each flag instead of the Event objects
+                flags_state = {
+                    "any_child_ecu_in_update": value.any_child_ecu_in_update.is_set(),
+                    "any_requires_network": value.any_requires_network.is_set(),
+                    "all_success": value.all_success.is_set(),
+                    # Add any other flags that might be present in MultipleECUStatusFlags
+                }
+                state_dict[key] = flags_state
+            else:
+                state_dict[key] = value
+        return pickle.dumps(state_dict)
+
+    def from_pickle(self, pickled_data):
+        """Deserialize a pickled byte string to restore the ECUStatusState instance.
+
+        Args:
+            pickled_data (bytes): Pickled representation of the ECUStatusState instance.
+        """
+        state_dict = pickle.loads(pickled_data)
+        for key, value in state_dict.items():
+            if key == "ecu_status_flags":
+                # Restore the state of each flag
+                if value.get("any_child_ecu_in_update", False):
+                    self.ecu_status_flags.any_child_ecu_in_update.set()
+                else:
+                    self.ecu_status_flags.any_child_ecu_in_update.clear()
+
+                if value.get("any_requires_network", False):
+                    self.ecu_status_flags.any_requires_network.set()
+                else:
+                    self.ecu_status_flags.any_requires_network.clear()
+
+                if value.get("all_success", False):
+                    self.ecu_status_flags.all_success.set()
+                else:
+                    self.ecu_status_flags.all_success.clear()
+            else:
+                setattr(self, key, value)
+
 
 class ECUStatusStorage:
     def __init__(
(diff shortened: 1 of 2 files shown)


src/otaclient/grpc/api_v2/ecu_status.py
```diff
@@ -64,7 +65,7 @@
 
 
 @dataclass
-class _ECUStatusState:
+class ECUStatusState:
```
airkei, line 68: this class will be used outside of this file, thus changed to public.

airkei:
close due to unnecessary PR

src/otaclient/grpc/api_v2/ecu_status.py
```diff
@@ -98,6 +99,48 @@ class _ECUStatusState:
...
     last_update_request_received_timestamp: int = 0
 
+    def to_pickle(self):
+        """Serialize the ECUStatusState instance to a byte string using pickle.
+
+        Returns:
+            bytes: Pickled representation of the ECUStatusState instance
+        """
+        state_dict = {}
+        for key, value in self.__dict__.items():
```
Bodong-Yang, line 109: `dataclasses` has `asdict` method, we can use this method to dump the dataclass into dict object.

Bodong-Yang (commented):
Overall it looks OK, leave one comment.

src/otaclient/grpc/api_v2/ecu_status.py
airkei: Thank you, good point.
The dataclass contains `MultipleECUStatusFlags`, and `asdict` doesn't support serialization of Event objects. So we need to iterate manually and apply special logic for Event objects.

Bodong-Yang: I mean, we should not directly accessing `self.__dict__`, but use `for key, value in asdict(self)` instead.

Directly accessing `__dict__` should only be used when we don't have other choice.
`asdict` is not for serialization purpose, just for converting dataclass object into dict object.

airkei: Understood. Even though it is only for extracting key and value, `asdict(self)` itself raises exception when `self` contains Event objects :cry:  One of the option is to ignore such exceptions. But I think `__dict__` is more safer in this case. What do you think?
```
src/otaclient/grpc/api_v2/ecu_status.py:109: in to_pickle
    for key, value in asdict(self):
/usr/lib/python3.10/dataclasses.py:1238: in asdict
    return _asdict_inner(obj, dict_factory)
/usr/lib/python3.10/dataclasses.py:1245: in _asdict_inner
    value = _asdict_inner(getattr(obj, f.name), dict_factory)
/usr/lib/python3.10/dataclasses.py:1245: in _asdict_inner
    value = _asdict_inner(getattr(obj, f.name), dict_factory)
/usr/lib/python3.10/dataclasses.py:1279: in _asdict_inner
    return copy.deepcopy(obj)
/usr/lib/python3.10/copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
/usr/lib/python3.10/copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
/usr/lib/python3.10/copy.py:146: in deepcopy
    y = copier(x, memo)
/usr/lib/python3.10/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
/usr/lib/python3.10/copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
/usr/lib/python3.10/copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
/usr/lib/python3.10/copy.py:146: in deepcopy
    y = copier(x, memo)
/usr/lib/python3.10/copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

x = <unlocked _thread.lock object at 0x78930f21dfc0>, memo = {132573009318032: <[AttributeError("'Condition' object has no attribute '_lock'") raised in repr()] Condition object at 0x78930f20a440>, 132573009325856: <threading.Event object at 0x78930f20abc0>, 132573011349056: {}, 132573011357312: {}}, _nil = []
```

Bodong-Yang: Oh I see, `asdict` tries to use deepcopy over Event object, which will not work. :notebook:

Bodong-Yang (approved):

PR #518, opened 2025-04-03
fix(v3.8.x): rpi_boot: fix backward compatibility issue when OTA downgrade from otaclient v3.8.x to v3.7.1

Bodong-Yang:
## Introduction

On raspberry pi based ECU, when OTA downgrading from otaclient v3.8.x to v3.7.1, ECU will not be able to finish up slot switching, and will boot back to previous slot.

This only happen when doing OTA downgrading.
1. OTA from otaclient v3.7.1 to v3.8.5 will work.
2. OTA from otaclient v3.8.5 to v3.8.5 will work.

This is caused by raspberry pi firmware update mechanism incompatible with old and new otaclient.
In otaclient v3.7.1, a two-stage reboot mechanism will be used to update firmware:
1. first-reboot: otaclient will do `flash-kernel` and **finalizing slot switching(critical)**, touch a flag file to indicate first-reboot finished.
2. second-reboot: basically do nothing, only clear the flag file and set OTA_STATUS to SUCCESS.

In otaclient v3.8.x, the two-stage reboot mechanism is replaced with doing update firmware with chroot in post-OTA phase, no need to do second reboot anymore. 
otaclient v3.8.x will also create the first-reboot done flag(as the firmware update has already done), but this will let otaclient v3.7.1 **skips slot switching finalizing**, resulting in boot back to old slot. 

This PR fixes the above issue, no longer create first-reboot done flag anymore.

## Tests

On the bench environment, I confirm with this PR, we can do the following update paths:
1. otaclient v3.7.1 -> otaclient with this PR.
2. otaclient with this PR -> otaclient v3.7.1.

### DIFF
--- a/src/otaclient/boot_control/_rpi_boot.py
+++ b/src/otaclient/boot_control/_rpi_boot.py
@@ -359,11 +359,6 @@ class _RPIBootControl:
                     initrd_img,
                     get_sysboot_files_fpath(INITRD_IMG, target_slot),
                 )
-
-            # NOTE(20240603): for backward compatibility(downgrade), still create the flag file.
-            #   The present of flag files means the firmware is updated.
-            flag_file = Path(cfg.SYSTEM_BOOT_MOUNT_POINT) / cfg.SWITCH_BOOT_FLAG_FILE
-            flag_file.write_text("")
             os.sync()
         except Exception as e:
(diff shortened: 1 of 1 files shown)


Zhenfeng-Sun (approved):

airkei (approved):
LGTM, thank you :rocket:

PR #522, opened 2025-04-15
feat: support client update API

airkei:
## Description
Implement `ClientUpdate` command.
The design and implementation detail is described in the following page.
https://tier4.atlassian.net/wiki/spaces/OTA/pages/3507421377/Client+Update+Design+Document

### Overview
This PR implements the client update API by adding a new OTA client update operator and extending related modules (boot control, configuration, and main entry) to support dynamic client updates. Key changes include adding the ClientUpdateControlFlags type and logic in the OTAClientUpdater and OTAClient API methods, as well as updating process shutdown and environment-related behaviors.

### NOTE
- Even though `ClientUpdate` is failed, `Update` will be triggered by Edge regardless of status
- When `ClientUpdate` is stopped for 5 minutes, try to proceed next step in any conditions.
- basically, the downloaded dynamic otaclient uses it's self-contained python framework, GLIBC and binaries. But use the original binaries for the following(host-dependency parts).
    - findmnt
    - reboot
    - grub-mkconfig
    - nvbootctrl

## Check list
- [x] local test is passed.

- Artifacts Size
    - manifest: 708B
    - squashfs
        - x86_64: 81MB
        - arm: 79MB

- E2E Tests
    - Single ECU(VM) 
        - Environment
          - Setup 
            - Host[dummy server] - VM(x86_64/grub)
          - Software
            - Host: [simulation scripts](https://tier4.atlassian.net/wiki/spaces/OTA/pages/3567092373/Simulate+an+OTA+update+locally) and [artifacts](https://github.com/tier4/ota-client/actions/runs/15410107542)(manifest and squashfs)
            - VM: [python package in artifact](https://github.com/tier4/ota-client/actions/runs/15412857510)
            
        - Test Procedure and Results
            - Log
            Verified that [the expected log](https://github.com/user-attachments/files/20564009/vm.txt) was output
            - Process
              - Verified that the status back to the original immediately by restarting when `ClientUpdate` is failed
              - Verified that the expected process is working after `ClientUpdate`(main process id is not changed)
                - before
                 ```
                 (venv) root@autoware:/opt/ota/client# ps aux | grep ota
                 ...
                root      662537  3.6  0.8 793384 67540 ?        Ssl  16:39   0:00 /opt/ota/client/venv/bin/python3 -m otaclient
                root      662573  0.2  0.1  16328 10868 ?        S    16:39   0:00 /opt/ota/client/venv/bin/python3 -c from multiprocessing.resource_tracker import main;main(6)
                root      662574  4.3  0.8 867096 67420 ?        Sl   16:39   0:00 /opt/ota/client/venv/bin/python3 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=7, pipe_handle=13) --multiprocessing-fork
                root      662575  3.0  0.6 806044 52428 ?        Sl   16:39   0:00 /opt/ota/client/venv/bin/python3 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=7, pipe_handle=15) --multiprocessing-fork
                 ```
                - after
                 ```
                 (venv) root@autoware:/opt/ota/client# ps aux | grep ota
                 ...
                root      662537  8.6  0.9 796216 73360 ?        Ssl  16:39   0:04 /otaclient/venv/bin/python3 -m otaclient
                root      667744  1.2  0.1  20156 15004 ?        S    16:40   0:00 /otaclient/venv/bin/python3 -c from multiprocessing.resource_tracker import main;main(6)
                root      667751 14.3  0.9 870344 76616 ?        Sl   16:40   0:01 /otaclient/venv/bin/python3 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=7, pipe_handle=13) --multiprocessing-fork
                root      667752 11.0  0.7 808660 57772 ?        Sl   16:40   0:01 /otaclient/venv/bin/python3 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=7, pipe_handle=15) --multiprocessing-fork
                 ```
              - Verified that these new process restarts by systemd after killing dynamic proces
              - Verified that unnecessary old process doesn't exist(no zombie) 
              ```
              (venv) root@autoware-ecu:/opt/ota/client# sudo kill -9 667751
              ...
              (venv) root@autoware-ecu:/opt/ota/client# ps aux | grep ota
              root      681328  3.2  0.8 793360 67592 ?        Ssl  16:42   0:00 /opt/ota/client/venv/bin/python3 -m otaclient
              root      681370  0.1  0.1  16460 10904 ?        S    16:42   0:00 /opt/ota/client/venv/bin/python3 -c from multiprocessing.resource_tracker import main;main(6)
              root      681380  3.5  0.8 867220 67260 ?        Sl   16:42   0:00 /opt/ota/client/venv/bin/python3 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=7, pipe_handle=14) --multiprocessing-fork
              root      681381  2.6  0.6 806044 53012 ?        Sl   16:42   0:00 /opt/ota/client/venv/bin/python3 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=7, pipe_handle=16) --multiprocessing-fork
              ```
          - Mount Point
              - Verified that mount points is not changed in the original process by `unshare`
              - Verified that the necessary points are mounted in process
              ```
              (venv) root@autoware-ecu:/opt/ota/client# cat /proc/7727/mountinfo
              647 626 7:0 / / rw,relatime unbindable - squashfs /dev/loop0 ro,errors=continue
              648 647 8:2 / /boot rw,relatime unbindable - ext4 /dev/sda2 rw
              649 647 0:5 / /dev rw,nosuid,relatime unbindable - devtmpfs udev rw,size=4035232k,nr_inodes=1008808,mode=755,inode64
              650 649 0:25 / /dev/shm rw,nosuid,nodev unbindable - tmpfs tmpfs rw,inode64
              651 647 8:3 /etc /etc rw,relatime unbindable - ext4 /dev/sda3 rw
              652 647 8:3 /home /home rw,relatime unbindable - ext4 /dev/sda3 rw
              653 647 8:3 /opt /opt rw,relatime unbindable - ext4 /dev/sda3 rw
              ...
              688 657 8:3 / /run/otaclient/mnt/active_slot ro,relatime unbindable - ext4 /dev/sda3 rw
              689 657 8:4 / /run/otaclient/mnt/standby_slot rw,relatime unbindable - ext4 /dev/sda4 rw
              ```
          - Behavior
              - Verified that Storage Status was kept during `ClientUpdate`
                  - `FAILURE` -> `ClientUpdate` -> `FAILURE`
                  - `SUCCESS` -> `ClientUpdate` -> `SUCCESS`
              - Verified that correct architecture(`x86_64`) squashfs was downloaded from otaclient. 
              - Verified that both `ClientUpdate` and `Update` was succeeded regardless of the Storage Status
              - Verified that `ClientUpdate` is finished immediately and back to the original status when there is no target squashfs
              - Verified that multiple `ClientUpdate` calls raises `RecursionError`
              - Verified that otaclient restart immediately when `ClientUpdate` is called and there is no `manifest.json` on the server
          - Bootloader Specific Behavior
              - Verified that both `grub-reboot` and `grub-mkconfig`  was succeeded

    - Single ECU(Perception1) 
        - Environment
          - Setup
          Verified in the both cases
            - Host[dummy server] - RQX-58G(aarch64/cboot/bsp 32.6.1)
            -  Host[dummy server] - RQX-58G(aarch64/uefi/bsp r35.4.1-rel.1v2.3)
          - Software
            - Host: [simulation scripts](https://tier4.atlassian.net/wiki/spaces/OTA/pages/3567092373/Simulate+an+OTA+update+locally) and [artifacts](https://github.com/tier4/ota-client/actions/runs/15410107542)(manifest and squashfs)
            - Perception1: [python package in artifact](https://github.com/tier4/ota-client/actions/runs/15161899475)
            
        - Test Procedure and Results
            - Log
            Verified that [the expected log(cboot)](https://github.com/user-attachments/files/20378354/perception1_log.txt) and [uefi.txt](https://github.com/user-attachments/files/20585226/perception1.txt) was output
          - Behavior
            Verified that correct architecture(`arm64`) squashfs was downloaded from otaclient. 
          - Bootloader Specific Behavior
            Verified that `nvbootctrl` was succeeded
          - Others
              Verified same tests as VM.

    - Multiple ECU
        - Environment
          - Setup 
            - Host[dummy server] - VM - RQX-58G(aarch64/cboot/bsp 32.6.1)
          - Software
            - Host: [simulation scripts](https://tier4.atlassian.net/wiki/spaces/OTA/pages/3567092373/Simulate+an+OTA+update+locally) and [artifacts](https://github.com/tier4/ota-client/actions/runs/15410107542)(manifest and squashfs)
            - VM and Perception1: [python package in artifact](https://github.com/tier4/ota-client/actions/runs/15412857510)

        - Test Procedure and Results
          - Behavior
            Verified that main ECU wait for the perception ECU finish when `ClientUpdate` is specified for multiple ECUs.

## Documents

https://tier4.atlassian.net/wiki/spaces/OTA/pages/3507421377/Client+Update+Design+Document

## Behavior changes

- [x] Yes, external behaivor (will impact user experience).
- [ ] No.

### Previous behavior

`ClientUpdate` command returns `NotImplementedError`

### Behavior with this PR

`ClientUpdate` command works based on the design.

## Breaking change

- [ ] Yes.
- [x] No.

## Related links & tickets
- https://tier4.atlassian.net/browse/RT4-13645
- https://github.com/tier4/autoware_ecu_system_setup/pull/1280

### DIFF
--- a/tests/test_otaclient/test_main.py
+++ b/tests/test_otaclient/test_main.py
@@ -303,6 +303,7 @@ class TestMain:
 
         # Mock execve to prevent actual process replacement
         mock_execve = mocker.patch(f"{MAIN_MODULE}.os.execve")
+        mocker.patch(f"{MAIN_MODULE}.sys.exit")
 
         mock_environ = {"PATH": "/usr/bin"}
         if is_preparing:


src/otaclient/configs/_cfg_configurable.py
```diff
@@ -58,6 +58,8 @@ class _OTAClientSettings(BaseModel):
     MAX_CONCURRENT_DOWNLOAD_TASKS: int = 128
     DOWNLOAD_INACTIVE_TIMEOUT: int = 5 * 60  # seconds
     MAX_RETRY_ON_ENTRY_COUNT: int = 300  # counts. almost 5 minutes
+    CLIENT_UPDATE_TIMEOUT: int = 5 * 60  # seconds
```
airkei, line 61: Will be used to wait for sub ecus.

src/otaclient/configs/_cfg_configurable.py
```diff
@@ -58,6 +58,8 @@ class _OTAClientSettings(BaseModel):
...
+    CLIENT_WAKEUP_RETRY_MAX: int = 3  # counts
```
airkei, line 62: Will be used to retry to run the downloaded client status.

src/otaclient/grpc/api_v2/main.py
```diff
@@ -79,9 +92,18 @@ async def _grpc_server_launcher():
...
             await server.stop(1)
             thread_pool.shutdown(wait=True)
+            logger.info("grpc API server stopped")
 
     asyncio.run(_grpc_server_launcher())
+
+    # Keep the process alive even after the gRPC server stops by stop_server_event
+    logger.info("gRPC server has stopped, but keeping the process alive")
+    while client_update_control_flags.stop_server_event.is_set():
+        time.sleep(1)
```
airkei, line 109: As described in the comments, keep sleeping the gRPC process without killing once the server is stopped. Because main process will do shutdown when the process is terminated.

src/otaclient/main.py
```diff
@@ -77,11 +81,59 @@ def _signal_handler(signal_value, _) -> None:  # pragma: no cover
     _on_shutdown(sys_exit=True)
 
 
+def _thread_dynamic_client(
```
airkei, line 84: try to run new application three times. If all of cases are failed, restart the existing otaclient.

src/otaclient/main.py
```diff
@@ -181,3 +245,27 @@ def main() -> None:  # pragma: no cover
...
+
+            # Create a closure with the current thread instance to avoid the loop variable binding issue
+            def create_thread_exit_handler(thread):
+                def _thread_exit():
+                    thread.join(6)
+
+                return _thread_exit
+
+            # Register the atexit handler with the current thread instance
+            atexit.register(create_thread_exit_handler(dynamic_client_thread))
```
airkei, line 271: In the existing design, each process is handled in main process.
Thus suggest to pass Event from core to main process, then handle each process based on the Event in main process.

src/otaclient/ota_core.py
```diff
@@ -786,6 +790,129 @@ def execute(self) -> None:
...
+            self._handle_upper_proxy()
+            self._process_metadata()
+            self._download_client_package_resources()
+            self._wait_sub_ecus()
+            if self._is_same_client_package_version():
+                self._handle_same_version()
+            else:
+                self._perform_update()
+        except Exception as e:
+            self._handle_update_failure(e)
```
airkei, line 828: This is core flowchart of client update. Restart will be triggered for any exceptions/errors.

src/otaclient/ota_core.py
```diff
@@ -786,6 +790,129 @@ def execute(self) -> None:
...
+        result = wait_and_log(
+            check_flag=self.ecu_status_flags.any_child_ecu_in_update.is_set,
+            check_for=False,
+            message="permit reboot flag",
+            log_func=logger.info,
+            timeout=cfg.CLIENT_UPDATE_TIMEOUT,
+        )
+        if result is False:
+            logger.warning("sub-ECU client update was aborted, shutdown client")
+            self.client_update_control_flags.request_shutdown_event.set()
```
airkei, line 875: Set timeout to wait for sub ecus(5mins).

src/otaclient/ota_core.py
```diff
@@ -962,6 +1091,56 @@ def update(self, request: UpdateRequestV2) -> None:
...
 
+    def client_update(self, request: ClientUpdateRequestV2) -> None:
+        """
+        NOTE that client update API will not raise any exceptions. The failure information
+            is available via status API.
+        """
+        if os.getenv(cfg.DOWNLOADED_DYNAMIC_OTA_CLIENT):
+            raise NotImplementedError(
+                "multiple dynamic clients downloading is not supported"
+            )
```
airkei, line 1102: For downloaded client, allow to wake up multiple clients.
The original otaclient is registered in systemd for restarting. Thus, think it's better to:
- wakeup new client from old client, the status is monitored by old client
- keep alive old client without doing anything(restart is supported by systemd).

src/otaclient/client_package.py
```diff
@@ -0,0 +1,297 @@
...
+                        self._rootfs_url, _package_file.lstrip("/")
+                    ),
+                    dst=_downloaded_package_path,
+                )
+            ]
+            condition.wait()  # wait for download finished
+        self.package = _available_package_metadata
+        self.downloaded_package_path = _downloaded_package_path
+
+    def _get_available_package_metadata(self) -> ReleasePackage:
```
airkei, line 141: This is the part to decide the target package/patch from available list in manifest.

src/otaclient/_utils.py
```diff
@@ -41,26 +41,53 @@ def wait_and_log(
     check_interval: int = 2,
     log_interval: int = 30,
     log_func: Callable[[str], None] = logger.info,
-) -> None:
-    """Wait for <flag> until it is set while print a log every <log_interval>."""
+    timeout: int | None = None,
```
airkei, line 44: add timeout for `wait_and_log`.
The motivation is to set timeout for sub ecu waiting in client update.

src/otaclient/_utils.py
```diff
@@ -41,26 +41,53 @@ def wait_and_log(
...
     """Check if there is another otaclient instance running, and then
     create a pid lock file for this otaclient instance.
 
     NOTE that otaclient should not run inside a PID namespace.
     """
+    # Skip check if the condition is met
+    if is_skip:
+        logger.info("Skipping otaclient multiple instances check")
+        return
+
```
airkei, line 90: if a specific environment variable is set, skip to check multiple otaclient running.

src/otaclient/client_package.py
airkei, line 1: This file handles the logic specific to client updating.

src/otaclient/configs/_cfg_consts.py
```diff
@@ -48,13 +48,17 @@ def ACTIVE_ROOT(self) -> str:  # NOSONAR
     MOUNT_SPACE = "/run/otaclient/mnt"
     ACTIVE_SLOT_MNT = "/run/otaclient/mnt/active_slot"
     STANDBY_SLOT_MNT = "/run/otaclient/mnt/standby_slot"
+    # mount point for downloaded otaclient
+    DYNAMIC_CLIENT_MNT = "/run/otaclient/mnt/dynamic_otaclient"
```
airkei, line 52: downloaded otaclient will be mount at this point.

src/otaclient/configs/_cfg_consts.py
```diff
@@ -48,13 +48,17 @@ def ACTIVE_ROOT(self) -> str:  # NOSONAR
...
 
     OTA_TMP_STORE = "/.ota-tmp"
     """tmp store for local copy, located at standby slot."""
     OTA_TMP_META_STORE = "/.ota-meta"
+    OTA_TMP_STATUS_STORE = "/.ota-ecu-status"
```
airkei, line 57: pickle file will be stored in this file.

src/otaclient/_types.py
```diff
@@ -133,6 +133,15 @@ class MultipleECUStatusFlags:
...
 
+@dataclass
+class ClientUpdateControlFlags:
+    """Flags for controlling the client update process."""
+
+    stop_server_event: mp_sync.Event  # for stopping gRPC server
+    request_shutdown_event: mp_sync.Event  # for requesting to shutdown
+    start_dynamic_client_event: mp_sync.Event  # for starting downloaded dynamic client
+
+
```
airkei, line 144: https://tier4.atlassian.net/wiki/spaces/OTA/pages/3507421377/Client+Update+Design+Document#Event-Flag-in-old-application

src/otaclient/configs/_cfg_consts.py
```diff
@@ -81,6 +85,11 @@ def ACTIVE_ROOT(self) -> str:  # NOSONAR
     OTA_API_SERVER_PORT = 50051
     OTAPROXY_LISTEN_PORT = 8082
 
+    #
+    # ------ env vars ------ #
+    #
+    DOWNLOADED_DYNAMIC_OTA_CLIENT = "DOWNLOADED_DYNAMIC_OTA_CLIENT"
```
airkei, line 91: https://tier4.atlassian.net/wiki/spaces/OTA/pages/3507421377/Client+Update+Design+Document#Environment-Variable-Communication-between-old%2Fnew-application

src/otaclient/configs/_cfg_consts.py
```diff
@@ -81,6 +85,11 @@ def ACTIVE_ROOT(self) -> str:  # NOSONAR
     OTA_API_SERVER_PORT = 50051
     OTAPROXY_LISTEN_PORT = 8082
 
+    #
+    # ------ env vars ------ #
+    #
+    RUNNING_DOWNLOADED_DYNAMIC_OTA_CLIENT = "RUNNING_DOWNLOADED_DYNAMIC_OTA_CLIENT"
+
```
airkei, line 92: https://tier4.atlassian.net/wiki/spaces/OTA/pages/3507421377/Client+Update+Design+Document#Environment-Variable-Communication

src/otaclient/ota_core.py
```diff
@@ -962,6 +1091,56 @@ def update(self, request: UpdateRequestV2) -> None:
...
 
+    def client_update(self, request: ClientUpdateRequestV2) -> None:
+        """
+        NOTE that client update API will not raise any exceptions. The failure information
+            is available via status API.
+        """
+        if os.getenv(cfg.RUNNING_DOWNLOADED_DYNAMIC_OTA_CLIENT):
+            raise NotImplementedError(
+                "multiple dynamic clients downloading is not supported"
+            )
```
airkei, line 1102: For downloaded client, allow to wake up multiple clients.
The original otaclient is registered in systemd for restarting. Thus, think it's better to:
- wakeup new client from old client, the status is monitored by old client
- keep alive old client without doing anything(restart is supported by systemd).

src/otaclient/grpc/api_v2/main.py
```diff
@@ -54,7 +62,15 @@ async def _grpc_server_launcher():
...
 
+        async def monitor_stop_event() -> None:
+            while not client_update_control_flags.stop_server_event.is_set():
+                await asyncio.sleep(1)
+            logger.info("grpc API server stop event detected")
+
         ecu_status_storage = ECUStatusStorage(ecu_status_flags=ecu_status_flags)
+        # Load the ECU status when the downloaded client is running
+        if load_state:
+            ecu_status_storage.load_state()
```
airkei, line 73: load ecu state only when current client is downloaded dynamic client.

src/otaclient/grpc/api_v2/main.py
```diff
@@ -79,9 +95,20 @@ async def _grpc_server_launcher():
         logger.info(f"launch grpc API server at {_address_info}")
         await server.start()
         try:
-            await server.wait_for_termination()
+            # Wait for the stop event without busy polling
+            await monitor_stop_event()
+            # Save the ECU status to a file
+            ecu_status_storage.save_state()
```
airkei, line 101: save ecu state after stopping gRPC.

src/otaclient/ota_core.py
```diff
@@ -962,6 +1091,56 @@ def update(self, request: UpdateRequestV2) -> None:
...
+
+        self._live_ota_status = OTAStatus.CLIENT_UPDATING
```
airkei, line 1104: set the status to `OTAStatus.CLIENT_UPDATING` for reporting at here.

src/otaclient/main.py
```diff
@@ -38,7 +38,7 @@
 
 logger = logging.getLogger(__name__)
 
-HEALTH_CHECK_INTERAVL = 6  # seconds
+HEALTH_CHECK_INTERVAL = 6  # seconds
```
airkei, line 41: Fix typo.

src/otaclient/main.py
```diff
@@ -101,7 +116,12 @@ def _thread_dynamic_client(
         # Run the OTA client
         # retry to start the OTA client multiple times if it fails
         for _ in range(cfg.CLIENT_WAKEUP_RETRY_MAX):
-            process = subprocess.Popen(
+            global _shutdown_processing, _dynamic_client_p
+            if _shutdown_processing:
+                logger.info("Shutdown requested, exiting dynamic client thread...")
+                return
```
airkei, line 122: this condition prevents to restart dynamic client in shutting down.

src/otaclient/_utils.py
```diff
@@ -41,26 +41,52 @@ def wait_and_log(
...
+) -> None:  # pragma: no cover
     """Check if there is another otaclient instance running, and then
     create a pid lock file for this otaclient instance.
 
     NOTE that otaclient should not run inside a PID namespace.
     """
+    # Skip check if the condition is met
+    if is_skip:
+        logger.info("skipping otaclient multiple instances check")
+        return
```
airkei, line 88: When dynamic client wakes up, skip the multiple instances check.

src/otaclient/ota_core.py
```diff
@@ -826,6 +854,124 @@ def execute(self) -> None:
...
+            self._handle_upper_proxy()
+            self._process_metadata()
+            self._download_client_package_resources()
+            self._wait_sub_ecus()
+            if not self._is_same_client_package_version():
+                self._perform_update()
+        except Exception as e:
+            logger.warning(f"failed to run squashfs: {e!r}")
+        finally:
+            self._request_shutdown()
```
airkei, line 892: [IMPORTANT] this is core sequence of `ClientUpdate`.

src/otaclient_common/cmdhelper.py
```diff
@@ -458,6 +520,54 @@ def mount_ro(
...
+    ]
+    # fmt: on
+    try:
+        subprocess_call(cmd, raise_exception=True)
+    except Exception:
+        subprocess_call(["losetup", "-d", loop_dev], raise_exception=False)
+        if raise_exception:
+            raise
+
+
```
airkei, line 570: When `mount` is failed, `losetup -d` is necessary to remove loop devices in addition to `umount`.
Thus, first setup loop device, then mount it. If error is occurred, remove the loope device.

tests/data/client_package/v1.squashfs
airkei, line 1: small squashfs file for test.

src/otaclient/client_package.py
```diff
@@ -0,0 +1,618 @@
...
+            self._create_mount_namespaces()
+            self._mount_squashfs_file(_squashfs_file, _mount_base)
+            self._bind_mount_host_dirs(_mount_base)
+            self._rbind_mount_current_root(_mount_base)
+            self._bind_mount_active_slot(_mount_base)
+
+            logger.info("mounted squashfs successfully")
+        except subprocess.CalledProcessError as e:
+            logger.exception(f"failed to mount squashfs: {e!r}")
+            raise
```
airkei, line 525: This is the core part of mount/bind.

src/otaclient/_otaproxy_ctx.py
```diff
@@ -119,6 +121,15 @@ def otaproxy_control_thread(
...
         _all_success = ecu_status_flags.all_success.is_set()
+        _stop_server_event = client_update_control_flags.stop_server_event.is_set()
+
+        if _stop_server_event:
+            logger.info("shutting down otaproxy as client update requested ...")
+            # kill the otaproxy process immediately
+            _shutdown_otaproxy()
+            while True:
+                # keep alive the thread without otaproxy running
+                time.sleep(1)
```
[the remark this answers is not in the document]
timothylin: I feel like we need a better way to handle these without `Thread.sleep()` everywhere :grimacing:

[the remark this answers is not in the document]
airkei: Good catch. Another option is to finish(kill) proxy thread/process without using sleep.
But this thread is registered as daemon, because it is expected to wake up in normal case for robustness.
So this process wake up again after finishing the thread/process.
https://github.com/tier4/ota-client/blob/5f534536e36b502e3b7d4f4c84b6839a7fd53290/src/otaclient/main.py#L159-L164

Based on the above background, my suggestion is to use sleep like current code without finishing thread/process.
Another another option is to create `Event` and wait it for forever without setting the flag, but I think current approach is better rather than creating new resource only for this purpose.

src/otaclient/client_package.py
```diff
@@ -0,0 +1,618 @@
...
+        with condition:
+            yield [
+                DownloadInfo(
+                    url=urljoin_ensure_base(
+                        self._rootfs_url, self.ENTRY_POINT.lstrip("/")
+                    ),
+                    dst=_client_manifest_fpath,
+                )
+            ]
+            condition.wait()  # wait for download finished
```
timothylin, line 122: what if download fails?

src/otaclient/configs/_cfg_configurable.py
airkei: Thank you. I don't have strong opinion to use either `5 * 60` or `300` :)
The existing code uses (min * 60) style, so let me keep the code consistency at here.
https://github.com/tier4/ota-client/blob/5f534536e36b502e3b7d4f4c84b6839a7fd53290/src/otaclient/configs/_cfg_configurable.py#L59

https://github.com/tier4/ota-client/blob/5f534536e36b502e3b7d4f4c84b6839a7fd53290/src/otaclient/configs/_cfg_configurable.py#L97

timothylin (dismissed):
LGTM!

Bodong-Yang (commented):
Just finish up half of the files' reviewing. I still need to take more time to read through the changes and review the implementation :pray:

src/otaclient_common/common.py
```diff
@@ -65,11 +66,14 @@ def subprocess_check_output(
...
 
     Returns:
         str: UTF-8 decoded stripped stdout.
     """
     try:
+        if chroot:
+            cmd = ["chroot", str(chroot)] + (cmd if isinstance(cmd, list) else [cmd])
         res = subprocess_run_wrapper(
             cmd, check=True, check_output=True, timeout=timeout
         )
```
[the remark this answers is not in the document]
airkei: sorry! I overlooked, fixed

src/otaclient_common/common.py
```diff
@@ -91,15 +95,19 @@ def subprocess_call(
...
     Args:
         cmd (str | list[str]): command to be executed.
         raise_exception (bool, optional): raise the underlying CalledProcessError. Defaults to False.
         timeout (Optional[float], optional): timeout for execution. Defaults to None.
+        chroot (Optional[str], optional): chroot path. Defaults to None.
     """
     try:
+        if chroot:
+            cmd = ["chroot", str(chroot)] + (cmd if isinstance(cmd, list) else [cmd])
         subprocess_run_wrapper(cmd, check=True, check_output=False, timeout=timeout)
```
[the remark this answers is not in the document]
airkei: fixed.

src/otaclient_common/cmdhelper.py
```diff
@@ -307,6 +315,8 @@ def reboot(args: list[str] | None = None) -> NoReturn:  # pragma: no cover
     if args:
         logger.info(f"will reboot with argument: {args=}")
         cmd.extend(args)
+    if chroot:
+        cmd = ["chroot", chroot, *cmd]
 
     logger.warning("system will reboot now!")
     subprocess_call(cmd, raise_exception=True)
```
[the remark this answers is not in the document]
airkei: fixed.

src/otaclient/client_package.py
```diff
@@ -0,0 +1,618 @@
...
+            # For older versions, use system call directly
+            import ctypes
+
+            CLONE_NEWNS = 0x00020000  # From linux/sched.h
+            libc = ctypes.CDLL("libc.so.6", use_errno=True)
+            if libc.unshare(CLONE_NEWNS) != 0:
+                errno = ctypes.get_errno()
+                raise OSError(
+                    errno, f"Failed to unshare mount namespace: {os.strerror(errno)}"
+                )
```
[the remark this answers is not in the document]
airkei: Thank you, added the description and log.

src/otaclient/client_package.py
```diff
@@ -0,0 +1,618 @@
...
+            f"dynamic client thread started with PID: {_otaclient_dynamic_client_t.ident}"
+        )
+        # wait for the thread to finish
+        _otaclient_dynamic_client_t.join()
+        _otaclient_dynamic_client_t = None
+        logger.info("dynamic client thread exited")
+
+
+def _dynamic_client_thread() -> None:
+    """Thread to run the dynamic client process."""
```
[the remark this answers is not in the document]
airkei: Renamed.

src/otaclient_common/cmdhelper.py
```diff
@@ -458,6 +520,54 @@ def mount_ro(
...
+    # fmt: off
+    cmd = [
+        "mount",
+        "--make-private", "--make-unbindable",
+        "-o", "ro",
+        "-t", "squashfs",
+        loop_dev,
+        str(mount_point),
+    ]
+    # fmt: on
```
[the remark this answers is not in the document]
airkei: Thank you! As you say, confirmed that `mount` and `umount` are enough to handle loop devices.

src/otaclient/ota_core.py
```diff
@@ -775,6 +780,27 @@ def _process_persistents(self, ota_metadata: OTAMetadata):
...
+        """Copy the client squashfs file to the standby slot."""
+        if not _env.is_dynamic_client_running():
+            logger.info(
+                "dynamic client is not running, no need to copy client squashfs file"
+            )
+            return
+
+        _src = Path(cfg.ACTIVE_SLOT_MNT) / Path(
+            cfg.DYNAMIC_CLIENT_SQUASHFS_FILE
+        ).relative_to("/")
```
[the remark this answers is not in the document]
Bodong-Yang: I defined a `CANONICAL_ROOT`(which value is `/`) in config package, but not all the places are using this const.

src/otaclient/client_package.py
```diff
@@ -0,0 +1,618 @@
...
+        os.makedirs(_mount_base, exist_ok=True)
+        try:
+            logger.info(f"mounting {_squashfs_file} to {_mount_base}")
+            self._cleanup_mount_point(_mount_base)
+            self._copy_client_package()
+            self._create_mount_namespaces()
+            self._mount_squashfs_file(_squashfs_file, _mount_base)
+            self._bind_mount_host_dirs(_mount_base)
+            self._rbind_mount_current_root(_mount_base)
+            self._bind_mount_active_slot(_mount_base)
```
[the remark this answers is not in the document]
Bodong-Yang: OK I know the purpose of this method. Resolved.

src/otaclient/client_package.py
```diff
@@ -0,0 +1,590 @@
...
+            _mount_point,
+            ignore_error=True,
+        )
+
+        cmdhelper.ensure_mount(
+            target=cfg.ACTIVE_ROOT,
+            mnt_point=_mount_point,
+            mount_func=cmdhelper.bind_mount_ro,
+            raise_exception=True,
+        )
```
Bodong-Yang, line 445: I know the purpose of this `_bind_mount_active_slot` function now, but it doesn't work as expected I think.

Here the target is `cfg.ACTIVE_ROOT`, which has value `/`. But for dynamic otaclient, `/` is the squashfs image, not the actual host root. You might need to change the `target` to `mount_base`.

Also, when mounting host rootfs, let's remove the `--make-unbindable` flag in `cmdhelper.rbind_mount_ro`, we don't need to worry about unexpected bind mount anymore as we are using `unshare`.

Bodong-Yang (commented):
Another batch of reviews. Let me continue the reviewing in next week's Monday, I still need to take more time on the review.

airkei:
Hi @Bodong-Yang That's really good point.
In the initial design, I consider to let main process to wake/kill dynamic client process like the following diagram.
https://tier4.atlassian.net/wiki/pages/viewpage.action?pageId=3507421377&pageVersion=48
image

The reason why I changed(let ota-core process to handle dynamic client process) is `unshare`.
- we need to keep the following order
    1. unshare
    2. mount/bind
    3. run
- `unshare` mount information will be propagated only into subprocess. So what was happened was:
    1. unshare in ota-core
    2. mount in ota-core
    3. run in main    <- this process is not unshared, the child dynamic client is also not unshared :cry: 

So 1,2,3 should be executed in same process, this is the reason why ota-core handles dynamic client in current code. 
As you say, we can move them to main process. What do you think?

src/otaclient/boot_control/_jetson_common.py
```diff
@@ -184,7 +185,10 @@ def _nvbootctrl(
         cmd.append(_cmd)
         if _slot_id:
             cmd.append(str(_slot_id))
+        if chroot:
+            cmd = ["chroot", str(chroot)] + cmd
 
+        logger.warning(f"nvbootctrl command: {cmd}")
         res = subprocess.run(
             cmd,
             check=True,
```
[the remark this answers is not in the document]
airkei: Fixed in https://github.com/tier4/ota-client/pull/522/commits/cc792cef88576aeac750397fddd2fc27c0c352a7.

src/otaclient/_otaproxy_ctx.py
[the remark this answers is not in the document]
airkei: I misunderstood `daemon` :bow: . And confirmed that proxy process never re-wake up automatically when it is killed.
Removed unnecessary loop in https://github.com/tier4/ota-client/pull/522/commits/b1ca08678b861988ae2cfce488958cb19dd96e1a.

src/otaclient/ota_core.py
```diff
@@ -1115,24 +1326,19 @@ def main(
                 )
 
 
-def _sign_handler(signal_value, frame) -> NoReturn:
-    print(f"ota_core process receives {signal_value=}, exits ...")
-    sys.exit(1)
-
-
```
airkei, line 1122: removed signal handler in ota_core, because it will be handled in main process.

Bodong-Yang:
@airkei Thank you for the updates! 
By the way, I notice that at my local demo, zombie process appears after `os.execve`, it turns out the python multiprocessing resource tracker process is not properly shutdown.
I take sometime searching, and found that this is a python bug(https://github.com/python/cpython/issues/88887), and it is fixed since python3.12 (https://github.com/python/cpython/pull/131530). 

Although for our workflow, we will at most only have one zombie process, as second `os.execve` doesn't involve multiprocessing, but it is better to handle it. 

A minimum cleanup can be achieved as follow:

```python
import multiprocessing as mp
import multiprocessing.resource_tracker as mp_resource_tracker

def main():

    ....

        # right before doing os.execve, after all sub processes are terminated and joined.
        if sys.version < (3, 12):
            _resource_tracker = getattr(mp_resource_tracker, "_resource_tracker", None)
            if _resource_tracker and hasattr(_resource_tracker, "_stop"):
                try:
                    _resource_tracker._stop()
                except Exception as e:
                    logger.error(f"failed to stop the resource tracker: {e!r}")
...
```

airkei:
@Bodong-Yang Good catch! Added before the 1st `os.execve` in https://github.com/tier4/ota-client/pull/522/commits/6b74007e94aa85feea3c8f4587cffa2d140da472, and tested this logic doesn't prevent `ClientUpdate`.

src/otaclient/boot_control/_jetson_common.py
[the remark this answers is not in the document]
airkei: Found the regression in the test and reverted.
 Let me keep this part as is for now, I left `TODO` comments.

Bodong-Yang (commented):
Another batch of reviews, I think I will call it a day, not all changes are reviewed though.
Mostly the implementation should work.

For any further issues found later, unless they break otaclient's feature, let's just add them into TODO list.

src/otaclient_common/_env.py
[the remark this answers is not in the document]
airkei: Fixed, using either `functools.cache` or `functools.lru_cache(maxsize=None)`.
```
try:
    cache = functools.cache  # type: ignore[attr-defined]
except AttributeError:
    cache = functools.lru_cache(maxsize=None)
```

src/otaclient/main.py
```diff
@@ -165,19 +214,62 @@ def main() -> None:  # pragma: no cover
...
                 f"ota API server is dead, whole otaclient will exit in {SHUTDOWN_AFTER_API_SERVER_EXIT}seconds ..."
             )
             time.sleep(SHUTDOWN_AFTER_API_SERVER_EXIT)
-            return _on_shutdown()
+            return _on_shutdown(sys_exit=True)
+
+        # launch the dynamic client preparation process
+        if client_update_control_flags.notify_data_ready_event.is_set():
+            try:
+                # kill ota proxy thread if it is running
```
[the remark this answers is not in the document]
airkei: Rephrased to `exit`.

src/otaclient/grpc/api_v2/ecu_status.py
```diff
@@ -473,3 +474,24 @@ async def export(self) -> api_types.StatusResponse:
...
+        _path = Path(cfg.OTACLIENT_STATUS_FILE)
+        if not _path.exists():
+            logger.warning(f"file {_path} does not exist, skipping loading state.")
+            return False
+
+        with open(_path, "rb") as f:
+            _pickle = f.read()
+        self._state.from_pickle(_pickle)
+        logger.info(f"Loaded ECU status storage from {_path}")
+        return True
```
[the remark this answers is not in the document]
airkei: Left TODO comments. But I think it will be not small effort to store the encryption key.
- hard code: not good
- receive from Secret Manager via FMS: will need another interface
- TPM

Maybe it will be future work when we actually need to use key.

src/otaclient/client_package.py
```diff
@@ -0,0 +1,462 @@
...
+
+    def copy_client_package(self) -> None:
+        """Copy the client package."""
+        _squashfs_file = cfg.DYNAMIC_CLIENT_SQUASHFS_FILE
+        # copy the squashfs file
+        os.makedirs(os.path.dirname(_squashfs_file), exist_ok=True)
+        shutil.copy(self._get_target_squashfs_path(), _squashfs_file)
+
+
+class OTAClientPackagePrepareter:
```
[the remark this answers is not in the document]
airkei: Sorry :bow:  Fixed.

src/otaclient/grpc/api_v2/ecu_status.py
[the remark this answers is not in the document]
Bodong-Yang: You miss the points, here what I mean is, what we need actually is not encryption, there is no secret in OTA status, what we need is keyed authentication, to ensure that actually the pickled object comes from old otaclient, not from attackers.
Or just a sha256 hash of the pickled object is also enough, although I would prefer keyed authentication.

The reason is unpicking arbitrary pickled object is a vulnerability and will lead to execution of arbitrary code during unpickling.

Bodong-Yang (approved):
Thank you! After we finish the test on bench, we are ready to go :+1:

PR #527, opened 2025-05-21
fix: publish metrics per each session

airkei:
### Why
Currently, metrics is published per each otaclient initialization.
The original intention is to publish it per each session.

### What
Control the session ids which have already published.

## Check list

- [x] local test is passed.

### DIFF
--- a/src/otaclient/metrics.py
+++ b/src/otaclient/metrics.py
@@ -19,6 +19,7 @@ import logging
 from dataclasses import asdict, dataclass
 
 from _otaclient_version import __version__
+
 from otaclient._logging import LogType
 from otaclient.configs.cfg import (
     ecu_info,
--- a/src/otaclient/metrics.py
+++ b/src/otaclient/metrics.py
@@ -19,7 +19,6 @@ import logging
 from dataclasses import asdict, dataclass
 
 from _otaclient_version import __version__
-
 from otaclient._logging import LogType
 from otaclient.configs.cfg import (
     ecu_info,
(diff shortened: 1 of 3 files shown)


Bodong-Yang:
If we want to use a new metric object for every session, what about just pass new metric object for every new session?
For example, we can utilize dataclasses' `replace` function(https://docs.python.org/3/library/dataclasses.html#dataclasses.replace) to make a copy of `OTAClient`'s metric object and pass it to the updater when creating new session, something like:

https://github.com/tier4/ota-client/blob/b26fdf83294e53940e4f6d6d3838b39175796e67/src/otaclient/ota_core.py#L991-L1005

```python
from dataclasses import replace
    ....
     _OTAUpdater( 
         ...
         metrics=replace(self._metrics), 
     ).execute() 
    ....
```

src/otaclient/metrics.py
```diff
@@ -71,15 +71,15 @@ class OTAMetricsData:
...
         Publishes the metrics data to the metrics server.
         """
-        if self._already_published:
+        if self.session_id in self._already_published_session_ids:
             # metrics data has already been published.
             return
 
         logger.info(json.dumps(asdict(self)), extra={"log_type": LogType.METRICS})
-        self._already_published = True
+        self._already_published_session_ids.add(self.session_id)
```
timothylin, line 85: how is the session ID generated?

airkei:
> If we want to use a new metric object for every session, what about just pass new metric object for every new session?
For example, we can utilize dataclasses' replace function(https://docs.python.org/3/library/dataclasses.html#dataclasses.replace) to make a copy of OTAClient's metric object and pass it to the updater when creating new session, something like:

Sounds like more smart way, thank you! Fixed.

src/otaclient/metrics.py
airkei: Good question. When ota-client receives gRPC API(`Update`, `ClientUpdate`, ...) requests, ota-client generates `<update_version>-<unix_timestamp_in_sec_str>-<4bytes_hex>` session id for each requests.
https://github.com/tier4/ota-client/blob/fix/publish_metrics_per_each-session/src/otaclient/_utils.py#L100-#L111

Bodong-Yang:
https://github.com/tier4/ota-client/blob/e9e900e70398132dff61070809cc6e1daf49dfee/src/otaclient/ota_core.py#L966-L974

Here the session_id is assigned before the copy is generated, what about let the `_OTAUpdater` assign the session_id to the metric copy.

airkei:
> Here the session_id is assigned before the copy is generated, what about let the _OTAUpdater assign the session_id to the metric copy.

Thank you! Fixed.

Bodong-Yang (approved):
LGTM!

PR #533, opened 2025-06-02
chore: bump up API proto version to 0.8.0

airkei:
### Why
To release API protocol buffer

### What
Bun up it's version to `0.8.0`.

### DIFF
--- a/proto/pyproject.toml
+++ b/proto/pyproject.toml
@@ -8,7 +8,7 @@ requires = [
 
 [project]
 name = "otaclient-pb2"
-version = "0.7.2"
+version = "0.8.0"
 readme = "README.md"
 requires-python = ">=3.8"
 classifiers = [


Bodong-Yang:
@airkei Ah please also update this version here:
```
[tool.hatch.build.hooks.custom]
proto_builds = [
  { proto_file = "otaclient_v2.proto", output_package = "otaclient_pb2/v2", api_version = "2.1.0" },
]
```

The version `0.8.0` is the python package version, while the above version is the API(`2.1.0`) is the API version.
Since we introduce a new API, and it is backward-compatible, we can bump the API version to `2.2.0`.

UPDATED: ah the API version in `pyproject.toml` is for reference only, please update the API version in https://github.com/tier4/ota-client/blob/main/proto/src/otaclient_pb2/v2/__init__.py .

airkei:
Thank you! In addition to bump up package version to `0.8.0`, update API version to `2.2.0` in the latest commit.

Bodong-Yang (approved):
LGTM!

PR #540, opened 2025-06-05
[3.9][RT4-17099][feat-dev-part2] feat(otaclient.create_standby): re-implement the create_standby functionality, introduce in-place update mode and refine rebuild mode

Bodong-Yang:
## Introduction

> [!NOTE]
> This PR is for review purpose only, it is 2nd part of the **RT4-17099 - inplace update mode** feature. It should be merged into `feat/inplace_update_mode` branch after 1st part is merged.
>
> Note that the new feature is not yet integrated into the ota_core, this is done in #541 .

This PR introduces the new implementation of create_standby feature, providing combinations of two delta calculation modes with two operation modes:
1. rebuild mode, with active slot base file_table assist.
2. rebuild mode, with full scanning active slot.
3. inplace mode, with standby slot base file_table assist.
4. inplace mode, with full scanning standby slot.

and a new update standby slot implementation, that utilizing the prepared resources by delta generation and downloading to update the standby slot..

For the design of the new implementation, please refer to https://tier4.atlassian.net/wiki/x/LISz2g.

## Ticket

https://tier4.atlassian.net/browse/RT4-17099
https://tier4.atlassian.net/browse/RT4-17100

PR #541, opened 2025-06-05
[3.9][RT4-17099][feat-dev-part3] feat: integrate the new create_standby implementation into otaclient

Bodong-Yang:
## Introduction

> [!NOTE]
> This PR is for review purpose only, it is 3rd part of the **RT4-17099 - inplace update mode** feature. It should be merged into `feat/inplace_update_mode` branch after 2nd part is merged.
>
> **Starting from this PR, the new feature is integrated and otaclient is functional.**

This PR integrates the changes from #540 and #539 into ota_core. 

Major changes include:
1. now otaclient will automatically choose which standby slot update mode by checks implemented at `can_use_in_place_mode`.
1. with introduction of inplace mode, the changes to standby slot before delta_calculation will be discarded, so `ota_status_control.pre_update_standby` becomes `ota_status_control.post_update_standby`.
1. for inplace update mode, it needs to know the device of standby slot, so update boot_control protocol and each implements to expose an API to provide `standby_slot_dev` information.

## Ticket

https://tier4.atlassian.net/browse/RT4-17099
https://tier4.atlassian.net/browse/RT4-17100


src/otaclient/boot_control/protocol.py
```diff
@@ -37,6 +37,11 @@ def get_booted_ota_status(self) -> OTAStatus:
     def get_standby_slot_path(self) -> Path:
         """Get the Path points to the standby slot mount point."""
 
+    @property
+    @abstractmethod
+    def standby_slot_dev(self) -> Path:
+        """The device of the standby slot."""
+
```
Bodong-Yang, line 44: When checking whether we can use `inplace` or not, we need to direct access to the standby slot's device. Let boot controller implementation exposes this information.

src/otaclient/configs/_cfg_configurable.py
```diff
@@ -62,8 +60,7 @@ class _OTAClientSettings(BaseModel):
     #
     # ------ create standby settings ------ #
     #
-    CREATE_STANDBY_METHOD: CREATE_STANDBY_METHOD_LTIERAL = "REBUILD"
```
Bodong-Yang, line 65: Don't expose the option of choosing which create_standby method to use now.
Detect automatically by `create_standby.utils.can_use_in_place_mode` helper, and prefer `inplace` mode if possible.

src/otaclient/ota_core.py
```diff
@@ -536,40 +542,70 @@ def _execute_update(self):
...
         )
-
         # prepare the tmp storage on standby slot after boot_controller.pre_update finished
-        self._resource_dir_on_standby.mkdir(exist_ok=True)
+        self._resource_dir_on_standby.mkdir(exist_ok=True, parents=True)
+        self._ota_tmp_meta_on_standby.mkdir(exist_ok=True, parents=True)
 
+        # NOTE(20250529): first save it to /.ota-meta, and then save it to the actual
+        #                 destination folder.
         logger.info("save the OTA image file_table to standby slot ...")
```
Bodong-Yang, line 573: Due to `inplace` mode will cleanup the standby slot during delta calculation, we cannot directly save the file_table to `/opt/ota/image-meta`, so we first save to a protected folder `/.ota-meta`, and later in post OTA update phase, we then move it into the `/opt/ota/image-meta`.

src/otaclient/ota_core.py
```diff
@@ -536,40 +545,45 @@ def _execute_update(self):
...
+
+        with TemporaryDirectory() as _tmp_dir:
+            use_inplace_mode = can_use_in_place_mode(
+                dev=self._boot_controller.standby_slot_dev,
+                mnt_point=_tmp_dir,
+                threshold_in_bytes=int(
+                    self._ota_metadata.total_regulars_size
+                    * STANDBY_SLOT_USED_SIZE_THRESHOLD
+                ),
+            )
```
Bodong-Yang, line 557: Prepare a temporary directory for `can_use_in_place_mode` to mount the standby slot for checking over it.

src/otaclient/ota_core.py
```diff
@@ -536,40 +545,45 @@ def _execute_update(self):
...
         self._boot_controller.pre_update(
             self.update_version,
-            standby_as_ref=False,  # NOTE: this option is deprecated and not used by bootcontroller
-            erase_standby=True,  # NOTE: as of now, we only have rebuild mode, so always erase_standby
+            # NOTE: this option is deprecated and not used by bootcontroller
+            # TODO:(20250613) when standby_as_ref is set, skip mounting active slot.
+            #       we cannot do this for now, as some boot controller impl still refer to
+            #       active_slot mounts.
+            standby_as_ref=use_inplace_mode,
+            erase_standby=not use_inplace_mode,
```
Bodong-Yang, line 569: Ideally, when we use inplace mode, we can set `standby_as_ref=True` to skip mounting active slot, but grub boot controller and jetson boot controller will look for boot config files in active slot.
So the `standby_as_ref` option is still not used.

src/otaclient/ota_core.py
```diff
@@ -536,40 +545,45 @@ def _execute_update(self):
...
         )
-
         # prepare the tmp storage on standby slot after boot_controller.pre_update finished
-        self._resource_dir_on_standby.mkdir(exist_ok=True)
+        self._resource_dir_on_standby.mkdir(exist_ok=True, parents=True)
+        self._ota_tmp_meta_on_standby.mkdir(exist_ok=True, parents=True)
```
Bodong-Yang, line 573: NOTE: for inplace update mode, these two OTA working folders will be ignored by inplace update mode delta generator.
Also see create_standby.delta_gen for more details.

src/otaclient/boot_control/_grub.py
```diff
@@ -741,6 +741,7 @@ def grub_reboot_to_standby(self):
 class GrubController(BootControllerProtocol):
     def __init__(self) -> None:
         try:
+            self._update_version = "unknown"
```
airkei, line 744: How about to define "unknown" in `BootControllerProtocol` class and reuse it in each sub-class?

src/otaclient/boot_control/_grub.py
```diff
@@ -871,6 +876,7 @@ def on_operation_failure(self):
     def pre_update(self, version: str, *, standby_as_ref: bool, erase_standby=False):
         try:
             logger.info("grub_boot: pre-update setup...")
+            self._update_version = version
```
airkei, line 879: It might be better to pass it as an argument like the following, rather than storing it in a class variable:
```
def pre_update(self, *, standby_as_ref: bool, erase_standby=False):
def post_update(self, version: str):
```

src/otaclient/ota_core.py
```diff
@@ -98,12 +99,14 @@
 DEFAULT_STATUS_QUERY_INTERVAL = 1
 WAIT_BEFORE_REBOOT = 6
 DOWNLOAD_STATS_REPORT_BATCH = 300
-DOWNLOAD_REPORT_INTERVAL = 1  # second
+DOWNLOAD_REPORT_INTERVAL = 3  # second
```
airkei, line 102: FYI. currently, Edge check status every second, then report it to backaend every 10 seconds.

src/otaclient/ota_core.py
airkei: How about rebuild mode? Just will be ignored like the following?
1. create directory
2. calculated delta based on active
3. copy active to standby, and remove this directory

src/otaclient/ota_core.py
```diff
@@ -580,37 +594,63 @@ def _execute_update(self):
...
+        base_meta_dir_on_standby_slot = None
         try:
-            if _verified_db := check_base_filetable(_base_ft_db):
-                logger.info(
-                    f"file_table for active_slot({_verified_db}) found and valid, use file_table to assist delta calculation!"
-                )
-                delta_calculator = DeltaGenWithFileTable(
+            if use_inplace_mode:
+                # try to use base file_table from standby slot itself
+                base_meta_dir_on_standby_slot = self._ota_tmp_meta_on_standby / "base"
```
airkei, line 601: Let's define "base" as variable somewhere.

src/otaclient/ota_core.py
airkei, line 1: Not requests in this PR.
Now `_execute_update` function is long and not good in terms of testability.
Feel better to divide into sub functions.

airkei (commented):
Thank you, left some comments :)

src/otaclient/boot_control/_grub.py
Bodong-Yang: ! Good catch ! I should really not do it like this, let me fix it up :innocent:

src/otaclient/ota_core.py
Bodong-Yang: Yes, totally agree! We need to do some cleanup and refactoring over `ota_core`, let's do it once 3.10 is released :+1:

src/otaclient/boot_control/_grub.py
Bodong-Yang: Fixed in ca4863e26947287943a55d211b1535368118de85.

airkei (approved):
LGTM! Thank you!!

PR #543, opened 2025-06-06
refactor: remove unnecessary status modules

airkei:
### Why
This is follow-up PR of https://github.com/tier4/ota-client/pull/522
In the original design, we plan to take over the whole status between process. But it is no more necessary.
Current Design:
1. During `ClientUpdate`, the original status in storage will not be updated
    - `SUCCESS` -> `ClientUpdate` -> `SUCCESS`
    - `FAILURE` -> `ClientUpdate` -> `FAILURE`
2. Main ECU wait for the sub ECU's complete, then reboot and set the status after rebooting. This sequence is same as `Update`.
3. Edge side ignores any returns during `ClientUpdate` and do `Update` regardless of `ClientUpdate` results.

### What
Remove pickle and related modules.

## Check list

- [x] local test is passed.

## Behavior changes

- [x] No.

## Breaking change

- [ ] Yes.
- [x] No.


Bodong-Yang (approved):
Thanks for the update!

PR #544, opened 2025-06-06
ci: test-CI: use pre-built test base container images

Bodong-Yang:
## Introduction

This PR refactors the Dockerfiles for container based test, and separate OTA image for test and OTA test base as follow:
1. `docker/mini_ota_img/Dockerfile.sys_img`: a Docker file for creating a minimum amd64 based system image.
1. `docker/mini_ota_img/Dockerfile.ota_img`: a Docker file for creating OTA image from minimum amd64 based system image.
2. `docker/test_base`: refactor to directly refer pre-built OTA image as OTA image source. 

Other changes:
1. `test_grub`: fix `ACTIVE_ROOT` const is not properly mocked with prepared dummy slot.

## Images push

The test base container images are available here: https://github.com/tier4/ota-client/pkgs/container/ota-client%2Ftest_base.
Since there is no any secret in the image, the image registry is set to public available.


tests/keys/gen_certs.sh
```diff
@@ -9,7 +9,7 @@ CA_CHAIN_PREFIX=${1:-test}
 # Root CA:
 openssl ecparam -out root.key -name prime256v1 -genkey
 openssl req -new -x509 \
-    -days $((365 * 10 + 5)) \
+    -days $((365 * 100)) \
```
timothylin, line 12: is there a reason why this number was increased?

docker/mini_ota_img/Dockerfile.ota_img
```diff
@@ -0,0 +1,101 @@
...
+    cp "${CERTS_DIR}"/sign.pem sign.pem; \
+    # git clone the ota-metadata repository
+    git clone ${OTA_METADATA_REPO}; \
+    pushd ota-metadata; git checkout 6c8ad47; popd; \
+    # prepare build environment
+    python3 -m venv ota-metadata/.venv; \
+    source ota-metadata/.venv/bin/activate; \
+    python3 -m pip install --no-cache-dir -U pip; \
+    python3 -m pip install --no-cache-dir -q \
+        -r ota-metadata/metadata/ota_metadata/requirements.txt; \
```
airkei: How about to use the internal operational rule that "We should update this base image when we update dependency". What do you think?

docker/mini_ota_img/Dockerfile.sys_img
```diff
@@ -0,0 +1,31 @@
+ARG UBUNTU_BASE=ubuntu:22.04
+
```
timothylin, line 2: is there a reason we changed the file type from regular `Dockerfile` to `.sys_img`? this breaks the code highlighting :sob:

docker/mini_ota_img/Dockerfile.ota_img
```diff
@@ -0,0 +1,101 @@
+ARG SYS_IMG=ghcr.io/tier4/ota-client/sys_img_for_test:ubuntu_22.04
+ARG UBUNTU_BASE=ubuntu:22.04
+ARG BUSYBOX_VER=busybox:1.37.0
+
```
timothylin, line 4: same for this file type. would it be possible to just keep it as `Dockerfile`?

.github/workflows/test.yaml
```diff
@@ -44,9 +44,6 @@ jobs:
         # sonarcloud needs full git histories
         with:
           fetch-depth: 0
-      - name: Build ota-test_base docker image
-        run: |
-          docker compose -f docker/test_base/docker-compose_tests.yml build tester-${{ matrix.test_base }}
```
airkei, line 49: (nits) We don't have to include the change into this PR.
It will be useful to add "base image creation" pipeline.

docker/mini_ota_img/Dockerfile.sys_img
Bodong-Yang: Ah `sys_img` here is just a suffix.
OK let me change it to prefix, and let the extension to be `Dockerfile`.

docker/mini_ota_img/Dockerfile.ota_img
Bodong-Yang: Updated in [b03f692](https://github.com/tier4/ota-client/pull/544/commits/b03f6929b47f843b0f53a2cc8695ed2971a09c3f).

docker/test_base/entry_point.sh
```diff
@@ -1,12 +1,20 @@
...
+mkdir -p ${TEST_ROOT}
+# source code needed to be rw, so copy it to the test_root
+cp -R ${OTACLIENT_SRC}/src ${TEST_ROOT}
+cp ${OTACLIENT_SRC}/uv.lock ${TEST_ROOT}
+# symlink all the other needed folders/files into test root
+ln -s ${OTACLIENT_SRC}/tests ${TEST_ROOT}
+ln -s ${OTACLIENT_SRC}/.git ${TEST_ROOT}
+ln -s ${OTACLIENT_SRC}/pyproject.toml ${TEST_ROOT}
+ln -s ${OTACLIENT_SRC}/README.md ${TEST_ROOT}
+ln -s ${OTACLIENT_SRC}/LICENSE.md ${TEST_ROOT}
```
airkei, line 17: can we move these copies into docker-compose?

airkei (commented):
Overall, LGTM. Thank you! This PR will improves our development efficiency drastically.
Left some comments, please take a look.

.github/workflows/test.yaml
Bodong-Yang: Currently there is no plan to implement such pipeline, as we expect that the test base container images will be stable, unless we change the specification.

Even we have a pipeline to create the test base docker image, it will not be in the `test.yaml`, and will not be executed by test CI. It should be a standalone CI.

docker/mini_ota_img/Dockerfile.ota_img
Bodong-Yang: Oh, another case when we need to rebuild the test base is, when we want to update the system image we used to build the OTA image.

docker/test_base/entry_point.sh
Bodong-Yang: IMO, we don't need to expose these details into docker-compose file though.

docker/mini_ota_img/Dockerfile.ota_img
airkei: The reason why I suggest is:
For example, in [this PR](https://github.com/tier4/ota-client/pull/537), dependabot suggest to bump up grpcio version. And grpcio deprecates its version at v1.71.0.
- Before this change: we can notice it in the test results in the PR even though we overtook the CHANGELOG.
- After this change: the tests will pass.

So especially when we update dependency, I think it's better to update test base image too. What do you think?

docker/test_base/entry_point.sh
airkei: OK, maybe I understand.
1. host code is bind to `otaclient_src` as ro in `docker-compose.yml`
```
    - ../..:/otaclient_src:ro
    - ../../test_result:/test_result:rw
```

2. copies the files as rw in `entry_point.sh`
```
cp -R ${OTACLIENT_SRC}/src ${TEST_ROOT}
...
```

This will be to prevent to change host code.
Hmmm, OK, I prefer to aggregate such file copy/bind in a single file. But understand the intention.

docker/mini_ota_img/Dockerfile.ota_img
airkei: Great, It's very clear and stable. Thank you for explanation :)

PR #547, opened 2025-06-09
refactor: refactor the client package modules

airkei:
### Why
To organize client package code.

### What
stop to refer `cfg` in `client_package.py`, and pass the necessary configuration variables from caller to `__init__`. Then reuse them in each function as instance variables.
This PR doesn't change the logic at all.

## Check list

- [x] local test is passed.

## Behavior changes

- [x] No.

## Breaking change

- [ ] Yes.
- [x] No.

### DIFF
--- a/src/otaclient/client_package.py
+++ b/src/otaclient/client_package.py
@@ -82,8 +82,8 @@ class OTAClientPackageDownloader:
         self._session_dir = Path(session_dir)
         self._download_dir = self._session_dir / f".download_{os.urandom(4).hex()}"
         self._download_dir.mkdir(exist_ok=True, parents=True)
-        self._package_install_dir = package_install_dir
-        self._squashfs_file = squashfs_file
+        self._package_install_dir = Path(package_install_dir)
+        self._squashfs_file = Path(squashfs_file)
 
         self._manifest = None
         self.package = None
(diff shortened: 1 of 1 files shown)


src/otaclient/client_package.py
```diff
@@ -129,7 +135,7 @@ def _prepare_client_package(
 
         # ------ step 2: download the target package ------ #
         _package_filename = _available_package_metadata.filename
-        _package_file = cfg.OTACLIENT_INSTALLATION_RELEASE + "/" + _package_filename
+        _package_file = str(self._package_install_dir) + "/" + _package_filename
```
Bodong-Yang, line 138: Not recommended to use `+` operator for string operations, also we should not use string operation against Path joining. Please use `str(self._package_install_dir / _package_filename)` instead.

Bodong-Yang (commented):
:+1: Injecting configurations by params is preferred over referring to a global variable, this will also make `client_package.py` more self-contained, more independent, and decoupling from otaclient itself.

I Just left some minor comments, please take a look.

src/otaclient/client_package.py
airkei: Thank you! In the code, `str` and `Path` are mixture for `StrOrPath`.
Fix to use `Path` in `__init__` in https://github.com/tier4/ota-client/pull/547/commits/fffa901e7e28d5680f9adc6e515a96bb8c61bdf4.

PR #551, opened 2025-06-11
fix: fix ota proxy kill procedure before dynamic client launch

airkei:
### Why
When `ClientUpdate` is called, 

## Check list

- [ ] local test is passed.


src/otaclient/_otaproxy_ctx.py
```diff
@@ -137,14 +135,16 @@ def otaproxy_control_thread(
...
             #   will still init the cache even init_cache is False.
-            _otaproxy_p = _mp_ctx.Process(
-                target=partial(otaproxy_process, init_cache=False),
-                name="otaproxy",
-            )
-            _otaproxy_p.start()
-            next_ota_cache_dir_checkpoint = otaproxy_min_alive_until = (
-                _now + OTAPROXY_MIN_STARTUP_TIME
-            )
+            if not _global_shutdown.is_set():
```
airkei, line 138: this is the most important part to prevent to spawn new process

timothylin (approved):
:+1:

src/otaclient/_otaproxy_ctx.py
```diff
@@ -39,26 +39,23 @@
...
-                _otaproxy_p.terminate()
-                _otaproxy_p.join()
-                _otaproxy_p = None
-        finally:
-            _global_shutdown_lock.release()
+    with _global_process_lock:
+        if _otaproxy_p:
+            _otaproxy_p.terminate()
+            _otaproxy_p.join()
+            _otaproxy_p = None
```
Bodong-Yang, line 58: Please use blocking=False here, this is on purpose.
Originally the shutdown handler is for signal handling, when multiple signals arrived, we only need one signal actually being handled at a time, while not blocking other signals handling(directly return).

For example, if we issue multiple `kill` at the same time, if we using blocking lock in the handler, all the `kill` except one will be blocked, that is not needed.

Bodong-Yang (commented):
Thank you for the fix! Just one question here, please have a look.

src/otaclient/_otaproxy_ctx.py
Bodong-Yang: We can add an option `blocking` for the `_shutdown_otaproxy` function.

Bodong-Yang (dismissed):
LGTM, thank you!

src/otaclient/_otaproxy_ctx.py
Bodong-Yang: Ah sorry I misunderstood that this handler is the signal handler, but it is not. Then your original changes are OK. :+1:

PR #323, opened 2024-06-21
refactor: implement downloader as a library module, provide Downloader and DownloaderPool

Bodong-Yang:
## Description

This PR refactors the downloader implementation, improves the cohesion of the module, and provides the new `Downloader` class(for single thread use) and `DownloaderPool`(for multi-threading environment use). 
The new implementation drops the direct thread pool support, instead providing the `DownloaderPool` that manages the thread-specific downloader instances, and can be used together with `ThreadPoolExecutorWithRetry` from #314 to achieve controllable downloading with retries.
Also, the test for `downloader` module is fully re-implemented with much higher coverage rate to over 90%, and fully tests the implementation of OTA-File-Cache-Control protocol support.

Other major changes includes:
1. migrate to urllib3 version2(2.2.2).
2. improve download error handling, add new OTA error for invalid OTA image, wrong cookies and wrong compression_alg.
3. refactor stats collection related to downloading.

## Check list

- [x] test file(s) that covers the change(s) is implemented.
- [x] test with VM is passed.
- [x] local test is passed.

PR #554, opened 2025-06-12
fix: fix otaclient stuck during downloading when OTA cache is broken

Bodong-Yang:
## Introduction

This PR fixes a regression that makes the `retry_caching` feature of otaclient/ota_proxy not working properly.

Originally, the `retry_caching` header directory is injected to request when the downloaded resource digest mismatched. However, since #323 , a pre-check of size is implemented before checking the digest, since in most cases, a broken cache will also have different size, causing the `retry_caching` header directory injection being skipped.

This will cause problem when the file is partially cached, resulting in infinity retrying(as current downloading breakout condition is completely no traffic for 5mins).

Other changes:
1. test_ota_proxy_e2e: refine the implementation, now launch the otaproxy in a new process, improve the test execution time.

## Related Bug Ticket
https://tier4.atlassian.net/browse/RT4-17523

### DIFF
--- a/tests/test_otaclient_common/test_downloader.py
+++ b/tests/test_otaclient_common/test_downloader.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-
 from __future__ import annotations
 
 import itertools
@@ -35,10 +34,12 @@ from requests.structures import CaseInsensitiveDict as CIDict
 
 from otaclient_common.common import urljoin_ensure_base
 from otaclient_common.downloader import (
+    BrokenDecompressionError,
     Downloader,
     DownloaderPool,
     DownloadPoolWatchdogFuncContext,
     HashVerificationError,
+    PartialDownload,
     check_cache_policy_in_resp,
 )
 from otaclient_common.retry_task_map import ThreadPoolExecutorWithRetry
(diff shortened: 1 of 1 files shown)


Bodong-Yang:
Will backport this fix to v3.9.x.

Bodong-Yang:
See https://github.com/tier4/ota-client/blob/ac6a7313c1278f69dcd41c59a65eae5d8acaeb8e/src/otaclient_common/downloader.py#L437-L446,

The size check happens before digest check, and in most cases, a broken OTA cache will have different size, effectively downloader now will mostly raise `PartialDownload` error on OTA cache broken.

src/otaclient_common/downloader.py
```diff
@@ -238,7 +237,7 @@ def retry_on_digest_mismatch(func: Callable[P, T]) -> Callable[P, T]:
     def _wrapper(*args: P.args, **kwargs: P.kwargs) -> T:
         try:
             return func(*args, **kwargs)
-        except (HashVerificationError, BrokenDecompressionError) as e:
+        except (PartialDownload, HashVerificationError, BrokenDecompressionError) as e:
```
airkei, line 240: (nits) How about to catch all of the inherited exceptions from `DownloadError`?
We don't have to care about similar issues in future.
```suggestion
        except DownloadError as e:
```

Bodong-Yang: IMO, being explicit is better here, so I prefer to list the concrete exception types here.

airkei:
LGTM too :)

Zhenfeng-Sun (commented):
Looks good to me. 
Just one line change. 
But the test change is big. 
Anyway, nice change.

Bodong-Yang:
Back ported at 196ec227d16c417326fe89bbbdb8154c1c692681.

PR #556, opened 2025-06-13
fix: cmdhelper.reboot: ensure to call sys.exit(0) after reboot subprocess call

Bodong-Yang:
## Motivation

During bench test, I found that on relatively slow device(comparing to main ECU), the `reboot` command might failed with `SIGTERM` even it actually does the `reboot`, causing the `cmdhelper.reboot` raises Exception unexpectedly. 
Turns out when device is not fast enough, python subprocess call might not be able to return after `reboot` command executed, and being killed by the `reboot` execution. 
We actually don't need to check `reboot` command's result as it will certainly do its job(signalling systemd to do `reboot`).

## Introdution

This PR fixes a potential issue of otaclient reboot procedure being interrupted unexpectedly due to subprocess call to `reboot` killed by `reboot` cmd. Now `cmdhelper.reboot` will not check the result of subprocess `reboot` call, and immediately do sys.exit(0), ensuring a clean exit.

### DIFF
--- a/src/otaclient/boot_control/_grub.py
+++ b/src/otaclient/boot_control/_grub.py
@@ -30,7 +30,6 @@ NOTE(20231027) A workaround fix is applied to handle the edge case of rootfs not
     still expecting new mechanism to fundamentally resolve this issue.
 """
 
-
 from __future__ import annotations
 
 import logging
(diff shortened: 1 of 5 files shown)


airkei (approved):
Thank you, LGTM :rocket:

Bodong-Yang:
Back ported at 893b344cc9c2ebb9b945f572084225b38633bc2b.

PR #557, opened 2025-06-13
fix: ensure status is pushed for every push interval, ensure firmware_version is pushed on otaclient starts up

Bodong-Yang:
## Motivation

Previously, the status_monitor status collector worker will only do pushing to grpc servicer when receiving report AND report is `OTAStatusChangeReport` or over push interval. 
When otaclient starts, the status monitor thread will receive `OTAStatusChangeReport` and `SetOTAClientMetaReport` within a very short time. While the former is reported, it will reset the push interval, causing the `SetOTAClientMetaReport` not being able to be pushed to grpc until next `OTAStatusChangeReport` being received.

## Introduction

When otaclient starts, the `firmware_version` is not reported via grpc status API. The `firmware_version` info is pushed to status_monitor internal storage, but not pushed to grpc in time.

This PR fixes this unexpected behavior by fixing otaclient status pushing implementation in status_monitor module, ensuring that status pushing to grpc servicer will certainly happen for every push interval. Also  the status report order in ota_core is also fixed.

## Ticket
https://tier4.atlassian.net/browse/RT4-17100


src/otaclient/_status_monitor.py
```diff
@@ -309,32 +307,36 @@ def _status_collector_thread(self) -> None:
         """Main entry of status monitor working thread."""
         _next_shm_push = 0
         while True:
-            _now = time.time()
+            _now = time.perf_counter()
```
timothylin, line 310: would you mind explaining this change?

timothylin (approved):
:+1:

src/otaclient/_status_monitor.py
```diff
@@ -309,32 +307,36 @@ def _status_collector_thread(self) -> None:
...
-                            f"failed to push status to shm: {e!r}"
-                        )
+                    logger.info("status collector thread exits on receiving sentinal")
+                    return
+
+                # report that is invalid
+                if not self.load_report(report):
+                    report = None
             except queue.Empty:
                 time.sleep(self.min_collect_interval)
```
airkei, line 324: When queue is empty, should we continue? Maybe same status might be written?

airkei (commented):
Overall, LGTM. Left one comments.

src/otaclient/_status_monitor.py
Bodong-Yang: Ah, here we don't actually need the real time, we just need a reference point to check if we need to do a push.
`time.perf_counter` is slightly faster than `time.time`, so I use it instead.

src/otaclient/_status_monitor.py
Bodong-Yang: ( While for `OTAStatusChangeReport`, we still always push no matter the push interval.

airkei (approved):
LGTM, thank you :rocket:

PR #560, opened 2025-06-16
fix: erase and prepare standby before choort

airkei:
## Check list

- [ ] local test is passed.


airkei:
Close to organize commit.

PR #563, opened 2025-06-16
fix: fix mount option in bind with R/W for dynamic client

airkei:
### Why
This issue was found in the dynamic client downloading test in Raspberry Pi.
In Raspberry Pi, for `flash-kernel`(to install dtb), some directories are binded in `Update` in chroot(dynamic client environment).
So these directories should be bind without `--make-unbindable` in initial bind(host).

### What
remove `--make-unbindable` option from `bind_mount_rw`.
Currently, this `bind_mount_rw` function is used only from the dynamic client package preparation.

## Check list

- [x] local test is passed.

Verified that raspberry Pi update was succeeded in E2E test.

### Current behavior
Fail to update Raspberry Pi in dynamic client.

### Behaivor after fix
Pass to update Raspberry Pi in dynamic client.

### DIFF
--- a/src/otaclient/client_package.py
+++ b/src/otaclient/client_package.py
@@ -401,6 +401,7 @@ class OTAClientPackagePreparer:
                     mnt_point=_mount_point,
                     mount_func=mount_func,
                     raise_exception=True,
+                    set_unbindable=False,
                 )
 
         # bind necessary directories
@@ -474,6 +475,7 @@ class OTAClientPackagePreparer:
             mnt_point=_mount_point,
             mount_func=cmdhelper.rbind_mount_ro,
             raise_exception=True,
+            set_unbindable=False,
         )
 
     # APIs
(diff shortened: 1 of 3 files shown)


Bodong-Yang (commented):
When otaclient is running under normal condition, `--make-unbindable` is required to prevent unexpected mount of otaclient used mount point by other program.

But it is OK to skip this flag when we run otaclient dynamically, as we have a dedicated mount namespace.

airkei:
> When otaclient is running under normal condition, --make-unbindable is required to prevent unexpected mount of otaclient used mount point by other program.
> But it is OK to skip this flag when we run otaclient dynamically, as we have a dedicated mount namespace.

Thank you, added `set_unbindable` option(default=True), and set False only when
- do bind mount to refer host directories in dynamic client preparation
- do rbind mount to execute host binaries in dynamic client preparation

Bodong-Yang (approved):
:+1:

PR #564, opened 2025-06-16
fix: ensure standby device umounted before choort for dynamic client

airkei:
### Why
This issue was found in the dynamic client downloading test in Raspberry Pi.
In `Update`, standby slot is requested to do 1.unmount, 2.erase then 3.mount dynamically.
In Raspberry Pi, unmount and erase(`mkfs.ext4`) is failed in chroot(dynamic client) environment because the standby slot has already been mounted in host process.

### What
unmount the standby device before unshare in host system.

## Check list

- [x] local test is passed.

The existing tests covers new function tests.
In addition, confirmed that E2E tests are passed in VM, Roscube and Raspberry Pi.

### Current behavior
Fail to Update in Raspberry Pi in dynamic client

### Behaivor after fix
Pass to Update in Raspberry Pi in dynamic client

### DIFF
--- a/src/otaclient/client_package.py
+++ b/src/otaclient/client_package.py
@@ -30,6 +30,7 @@ from typing import Generator, Optional
 
 from ota_metadata.legacy2.metadata import OTAMetadata
 from otaclient import __version__
+from otaclient.boot_control import BootloaderType, get_boot_controller
 from otaclient_common import cmdhelper
 from otaclient_common._typing import StrOrPath
 from otaclient_common.common import (
@@ -310,6 +311,7 @@ class OTAClientPackagePreparer:
         active_root: StrOrPath,
         active_slot_mnt_point: StrOrPath,
         host_root_mnt_point: StrOrPath,
+        bootloader: BootloaderType,
     ) -> None:
         """Initialize the OTA client package preparer."""
         self._squashfs_file = squashfs_file
@@ -317,6 +319,7 @@ class OTAClientPackagePreparer:
         self._active_root = active_root
         self._active_slot_mnt_point = active_slot_mnt_point
         self._host_root_mnt_point = host_root_mnt_point
+        self._bootloader = bootloader
 
     def _cleanup_mount_point(self) -> None:
         """Cleanup the mount point."""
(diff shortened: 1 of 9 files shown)


src/otaclient/client_package.py
```diff
@@ -486,7 +475,7 @@ def mount_client_package(self) -> None:
             self._create_mount_namespaces()
             self._mount_squashfs_file()
             self._bind_mount_host_dirs()
-            self._bind_mount_active_slot()
+            self._prepare_active_and_standby_slots()
```
airkei, line 478: In the client mount sequence, erase standby then mount both standby and active slots.

Bodong-Yang (commented):
This will conflicts with in-place mode, standby slot should not be erased when we use in-place update mode, please revise the design.

airkei:
@Bodong-Yang Thank you!
In in-place mode, how do you plan to handle standby in the existing code?
I think we can follow same thing in dynamic client.
https://github.com/tier4/ota-client/blob/646370e058d1ef73cef1820621dc51e8f6485f5b/src/otaclient/ota_core.py#L571-L575

Bodong-Yang:
@airkei 
Please let the dynamic otaclient do the standby slot mounting, just like in normal otaclient mode, do not pre-mount the standby slot before starting up the dynamic otaclient.
The boot controller knows which device is standby slot, just let `slot_mount_helper` does its job to prepare and mount standby slot.

src/otaclient/client_package.py
```diff
@@ -335,6 +338,15 @@ def _cleanup_mount_point(self) -> None:
...
             logger.warning(f"error while removing dynamic client mount point: {e}")
 
+    def _umount_standby_device(self) -> None:
+        """Prepare the active and standby slots for the OTA update."""
+        _boot_controller_type = get_boot_controller(self._bootloader)
+        _boot_controller = _boot_controller_type()
+        _standby_slot_dev = _boot_controller.get_standby_slot_dev()
+
+        logger.info(f"unmounting standby slot device: {_standby_slot_dev}")
+        cmdhelper.ensure_umount(_standby_slot_dev, ignore_error=False)
```
airkei, line 348: specify the device like `/dev/sda2`.

PR #568, opened 2025-06-17
fix: always report SUCCESS after ClientUpdate

airkei:
### Why
Issue: In OTA update, FMS Console might shows "0% -> 3% -> **FAILURE** -> 5% -> ... -> 100% -> SUCCESS" as progress.

Currently, only the last `Update` status is stored in status file, then load and report it when ota-client wakes up.
When `ClientUpdate` is finished at 3% progress, restart by `execve` will be triggered then report the last `Update` status to upper.
If the last `Update` is FAILURE, ota-client will report its FAILURE after `ClientUpdate`.
In FMS side, this is not problem in terms of sequence because they ignore SUCCESS/FAILURE status after `ClientUpdate`.
But the FAILURE might be shown on FMS Console for a while, this will make users anxious.

### What
As temporary solution, always report SUCCESS after `ClientUpdate`.
This status is used only for reporting, this status will not be stored into the existing status file.

For root cause measurement
  - ota-client prepares the dedicated status file for `ClientUpdate`
  - ota-client API adds new ClientUpdate status field in gRPC `Status` API
  - Edge migrate `Status` handling from v1 to v2
  - Edge handle its ClientUpdate status correctly.

## Check list

- [x] local test is passed.

### Current behavior
FAILURE might be shown after `ClientUpdate`(3%) on FMS Console.

### Behaivor after fix
FAILURE will not be shown after `ClientUpdate`.


Bodong-Yang:
Thank you for your explanation! I understand the situation now, as for short-term solution, I agree to hack to re-use the `status` file to store the `CLIENT_UPDATING` status. 
Then please remember to write the `CLIENT_UPDATING` status to the `status` file, so that the newly started dynamic otaclient can read the `CLIENT_UPDATING` status.

Bodong-Yang:
But for long-term, we still need a separated status file and a dedicated field in API proto for otaclient updating.

airkei:
> Thank you for your explanation! I understand the situation now, as for short-term solution, I agree to hack to re-use the status file to store the CLIENT_UPDATING status.
Then please remember to write the CLIENT_UPDATING status to the status file, so that the newly started dynamic otaclient can read the CLIENT_UPDATING status.

Thank you, and sorry I completely misunderstood.
Currently, as you say, `CLIENT_UPDATING` is not stored into the status file. So the original change is meaningless, maybe I misunderstood due to the frontend cache :cry: 

One of the option is store to `CLIENT_UPDATING` into status file as you say, but it means we will lose the last `Update` status even though this is temporary workaround.
Another option is to use environment variable. This `_load_status_file` is called only in ota-core initialization so far. My suggestion is to return SUCCESS  when the dynamic client environment variable is set(= the ota-core initialization after client updating).

What do you think?
I know this is temporary and not clean, and think it's better to narrow down the affected modules without changing status file policy.

Bodong-Yang (approved):
👍 Let's use your workaround for now, and refine it in the future.

PR #570, opened 2025-06-19
fix: exit dynamic client when Update error occurrs

airkei:
### Why
This issue was found in Bench test.

When `Update` in dynamic client is failed, otaclient keep staying in dynamic client environment.
And when FMS sends next `ClientUpdate` based on the user trigger on FMS Console, it will raises duplicated client update call error.
In short, user can't do anything when `Update` is failed in dynamic client environment.

### What
- When `Update` is failed, exit dynamic client environment. Next `ClientUpdate` might request another image.
- When duplicated `ClientUpdate` is called, just return without updating "OTA" status. This should not happen in current sequence.
- Stop to set `Success` or `FAILURE` in `ClientUpdate`, this is future task to set the dedicated client updating status

## Check list

- [x] local test is passed.

Verified the actual behavior in E2E tests.

### Current behavior
When `Update` in dynamic client is failed, otaclient keep staying in dynamic client environment.

### Behaivor after fix
When `Update` in dynamic client is failed, otaclient restart the process and back to the original(installed) otaclient.

### DIFF
--- a/src/otaclient_common/downloader.py
+++ b/src/otaclient_common/downloader.py
@@ -20,6 +20,7 @@ from __future__ import annotations
 
 import hashlib
 import logging
+import os
 import threading
 import time
 from abc import abstractmethod
@@ -401,6 +402,8 @@ class Downloader:
         with self._session.get(
             prepared_url, stream=True, headers=prepared_headers, timeout=timeout
         ) as resp, open(dst, "wb") as dst_fp:
+            dst_fd = dst_fp.fileno()
+            os.posix_fadvise(dst_fd, 0, 0, os.POSIX_FADV_NOREUSE)
             resp.raise_for_status()
 
             digest, compression_alg = check_cache_policy_in_resp(
@@ -433,6 +436,10 @@ class Downloader:
                     _read_size = len(_chunk)
                     self._downloaded_bytes += _read_size
                     traffic_on_wire += _read_size
+
+            dst_fp.flush()
+            os.fsync(dst_fd)
+            os.posix_fadvise(dst_fd, 0, 0, os.POSIX_FADV_DONTNEED)
 
         if size and size != downloaded_file_size:
             _err_msg = (


src/otaclient/ota_core.py
```diff
@@ -892,7 +892,7 @@ def _execute_client_update(self):
                 self._notify_data_ready()
         except Exception as e:
             logger.warning(f"failed to run squashfs: {e!r}")
-            self._request_shutdown()
+            raise
```
airkei, line 895: moved(handle exceptions in caller side).

src/otaclient/ota_core.py
```diff
@@ -1152,21 +1162,16 @@ def update(self, request: UpdateRequestV2) -> None:
...
-            self._live_ota_status = OTAStatus.FAILURE
-            e = ota_errors.DuplicatedClientUpdateRequest(module=__name__)
-            self._on_failure(
-                e,
-                ota_status=OTAStatus.FAILURE,
-                failure_reason=e.get_failure_reason(),
-                failure_type=e.failure_type,
-            )
+            # Duplicates client update should not be allowed.
+            # TODO(airkei) [2025-06-19]: should return the dedicated error code for "client update"
```
airkei, line 1174: Not set status in client update.
When duplicated `ClientUpdate` is called, just return without do anything.

src/otaclient/ota_core.py
```diff
@@ -1201,14 +1206,11 @@ def client_update(self, request: ClientUpdateRequestV2) -> None:
...
-                e,
-                ota_status=OTAStatus.FAILURE,
-                failure_reason=e.get_failure_reason(),
-                failure_type=e.failure_type,
-            )
+        except ota_errors.OTAError:
+            # TODO(airkei) [2025-06-19]: should return the dedicated error code for "client update"
+            self._exit_from_dynamic_client()
+        except Exception:
+            self._exit_from_dynamic_client()
```
airkei, line 1213: Not set status in client update at the moment, just restart and back to the original otaclient.

airkei:
@Bodong-Yang Thank you for comments. Completely agree, I will be in charge of v2 migration(hopefully, will finish by 9/E).

> ( @airkei Just want to confirm, since we don't have version information in v1 status API format, does it mean otaclient client update will occur before every OTA?

Yes. And currently we need to download `manifest.json` to get the target ota-client version.
So in my understanding, even though we migrate to status v2, Edge needs to call `ClientUpdate` every time, then ota-client judge downloads manifest and judge if the client update is necessary or not.

PR #580, opened 2025-06-26
ci: introduce update_lockfile workflow by extending gen_requirements_txt workflow

Bodong-Yang:
## Introduction

As the efforts of introducing lock files for ota-client, this PR introduces the workflow to update lock files on pyproject.toml changed. The following lock files will be updated by:
1. uv.lock: updated with `uv sync`
2. requirements.txt: updated with `uv export --frozen --no-dev --no-editable --no-hashes --no-emit-project > requirements.txt`

Starting from this PR, requirements.txt will also become lock file. Unlink uv.lock that is only recognized by uv.lock, requirements.txt lock file can be recognized by many other compatible tools.


airkei (dismissed):
Thank you!! LGTM.

(nits) there is still comment using `gen_requirements_txt` in `test.yml`. Might be better to update.
https://github.com/tier4/ota-client/blob/6c9829ebe31673c1b7145ba87a0da3b6d9f2884d/.github/workflows/test.yaml#L12-L19

airkei (approved):
LGTM! Thank you.

PR #586, opened 2025-07-01
[3.9] ci: introduce deps locked wheel package build for otaclient with hatch build hook and uv; cherry-pick #580

Bodong-Yang:
## Introduction

As the critical step to achieve locking down dependencies for otaclient distribution, this PR implements a special hatch metadata build hook which enables dependencies overwrite during wheel build, and enables this feature for wheel build.

This is achieved by involving `hatch_build_lock_deps.py` build hook(with `ENABLE_DEPS_LOCKED_BUILD=yes` configured) during wheel build. This build hook will call `uv export --frozen --no-dev --no-color --no-editable --no-hashes --no-emit-project` and update the `dependencies` information with `uv export` result. 

otaclient squashfs app image build is using the just built otaclient wheel in CI, so naturally otaclient squashfs app image will also have dependencies locked down.

Other changes:
1. cherry-pick #580.
2. docker/test_base: fix up entrypoint.sh.

## Ticket
https://tier4.atlassian.net/browse/RT4-18097


Bodong-Yang:
Forward ported to main branch at https://github.com/tier4/ota-client/pull/587.

.github/workflows/release.yml
```diff
@@ -29,6 +29,9 @@ jobs:
           python -m pip install -U hatch
 
       - name: build otaclient package
+        env:
+          # NOTE: see hatch_build_lock_deps.py for more details
+          ENABLE_DEPS_LOCKED_BUILD: "yes"
```
timothylin, line 34: should we do boolean instead of `yes`?

Bodong-Yang: Ah using string literal is on purpose, as "true" has special meaning for yaml. To prevent unexpected conversion, I use "yes" here.
Also, in normal condition(like doing local development), this hook should not be enabled, so in the hook implementation, I use strict full string literal comparison to check if ENABLE_DEPS_LOCKED_BUILD is set.

.github/workflows/update_lockfile.yaml
```diff
@@ -0,0 +1,65 @@
...
+
+permissions:
+  contents: read
+  packages: read
+  actions: read
+
+on:
+  pull_request:
+    branches:
+      - main
```
airkei, line 11: How about to add `- 'v*'` branches?

hatch_build_lock_deps.py
airkei, line 1: 1. Let's add our copyright header.
2. I feel it is natural to put this file under `tools` rather than top directory.

airkei (commented):
Thank you :) left some comments, PTAL.

hatch_build_lock_deps.py
airkei: Thank you, Hmmm, I understand your intention, but still feel strange to put python file on top except `setup.py` or `config.py`...
If this file is expected to be run only in CI, how about to put `.github/tools/` like `gen_requirements_txt.py`?

Bodong-Yang: Update in [b6becd0](https://github.com/tier4/ota-client/pull/586/commits/b6becd05dbf929aee21bd3041ab321c654f667a1) and [cddfe59](https://github.com/tier4/ota-client/pull/586/commits/cddfe59275d0c8eeea6f90b2a4f3cbe914accbf2).

PR #592, opened 2025-07-03
[3.9] fix: replace_root and TopDownCommonShortestPath: DO NOT resolve input paths

Bodong-Yang:
## Introduction

This PR fixes a bug caused by `replace_root` and `TopDownCommonShortestPath` try to resolve the input paths. 

This will introduce side effect when input path is somehow not an actual dir, but a symlink, on the active slot.
The path will be resolved by looking at current rootfs filesystem, which is unexpected if the input path should be relative to standby slot during delta calculation.

Also, paths resolve is not needed in the first place, so the change here doesn't break the current behavior.

### DIFF
--- a/src/otaclient/create_standby/utils.py
+++ b/src/otaclient/create_standby/utils.py
@@ -39,7 +39,6 @@ class TopDownCommonShortestPath:
         self._store: set[Path] = set()
 
     def add_path(self, _path: Path):
-        _path = Path(_path).resolve()
         for _parent in _path.parents:
             # this path is covered by a shorter common prefix
             if _parent in self._store:
--- a/src/otaclient_common/__init__.py
+++ b/src/otaclient_common/__init__.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 """Common shared libs for otaclient."""
 
-
 from __future__ import annotations
 
 import os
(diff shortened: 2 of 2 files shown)


airkei (commented):
Thank you. Maybe I don’t know what the issue is or when it happened(when input is standby slot path, the issue will be happened?)
Could you give me example cases?

Bodong-Yang:
> Thank you. Maybe I don’t know what the issue is or when it happened(when input is standby slot path, the issue will be happened?)
Could you give me example cases?

The case I hit at bench No.3 is as follow:

1. The active slot is running Ubuntu 22.04, the standby slot is still Ubuntu 20.04, the new OTA image is Ubuntu 22.04.
2. There is a path `/usr/include/CharLS`, on Ubuntu 20.04, it is a real folder, while in Ubuntu 22.04, it becomes a symlink points to `/usr/include/charls`.
3. With inplace update mode, otaclient will remember the directories that should be removed. For `/usr/include/CharLS`, on standby slot, it is a real folder, so otaclient needs to remove it. However, `TopDownCommonShortestPath.add_path` do a `resolve` over the input `/usr/include/CharLS`, and it resolves it over `/`(i.e., over active slot), and on active slot, `/usr/include/CharLS` is a symlink, this result in otaclient not actually remove the `/usr/include/CharLS` folder on standby slot.
4. when during update standby slot phase, otaclient tries to create symlink `/usr/include/CharLS`, and failed as the `/usr/include/CharLS` folder still exists on standby slot.

PR #602, opened 2025-07-14
refinement: skip metafiles parsing for client update

airkei:
### Why
Currently, metadata parsing is performed during client updates.
However, this step is unnecessary, as only certificate and signature verification are required.
Removing metadata parsing could reduce the update time by approximately 20 seconds.

### What
Skip metadata parsing for client update, just verify cert/signature verification.

## Check list

- [x] local test is passed.

### Behaivor after fix
parse metadata in `ClientUpdate`.

## Related links & ticket
skip to parse metadata in `ClientUpdate`.

### DIFF
--- a/src/ota_metadata/legacy2/metadata.py
+++ b/src/ota_metadata/legacy2/metadata.py
@@ -342,6 +342,8 @@ class OTAMetadata:
     def download_metafiles(
         self,
         condition: threading.Condition,
+        *,
+        only_metadata_verification: bool = False,
     ) -> Generator[list[DownloadInfo]]:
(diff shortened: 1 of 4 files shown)


Bodong-Yang (approved):
LGTM!
By the way, FYI, the `prepare_ota_image_metadata` converts the TXT based file_table into sqlite3 database with new OTA image spec's file_table db schema.
When we migrate to new OTA image, we don't need `prepare_ota_image_metadata` and `prepare_persist_meta` anymore,  OTA image is ready for OTA right after we finish up verifying the `index.json` with `index.jwt`.
Also, thanks to we do the conversion, when we migrate to new spec, we can re-use the create standby package almost as it.

PR #605, opened 2025-07-16
refactor: refactor exception handling in client update

airkei:
## Check list

- [ ] test file(s) that covers the change(s) is implemented.
- [ ] local test is passed.

## Behavior changes

- [ ] No.

## Breaking change

- [ ] Yes.
- [ ] No.

### DIFF
--- a/src/otaclient/client_package.py
+++ b/src/otaclient/client_package.py
@@ -286,6 +286,7 @@ class OTAClientPackageDownloader:
             logger.exception(
                 f"failure during downloading and verifying OTA client package: {e!r}"
             )
+            raise
 
     def is_same_client_package_version(self) -> bool:
         """Check if the current OTA client package version is the same as the target version in manifest"""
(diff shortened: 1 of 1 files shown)


src/otaclient/client_package.py
```diff
@@ -286,6 +286,7 @@ def download_client_package(
             logger.exception(
                 f"failure during downloading and verifying OTA client package: {e!r}"
             )
+            raise
```
airkei, line 289: When the manifest file is missing from S3, the exception is caught by the upper layer.
Therefore, this change does not affect the actual behavior, but it ensures that exceptions are properly raised at this point.

src/otaclient/client_package.py
```diff
@@ -325,20 +326,14 @@ def __init__(
...
+            self._mount_base, ignore_error=True, max_retry=0, retry_interval=0
+        )
 
-        try:
-            # remove the dynamic client mount point
-            shutil.rmtree(self._mount_base, ignore_errors=False)
-        except Exception as e:
-            logger.warning(f"error while removing dynamic client mount point: {e}")
+        # remove the dynamic client mount point
+        shutil.rmtree(self._mount_base, ignore_errors=True)
```
airkei, line 336: Exceptions are ignored in the original code, so using ignore_error(s) is sufficient.

src/otaclient/client_package.py
Bodong-Yang: Just curious, what will happen if we don't raise here 🤔

airkei: Later step(client copy stage) try to handle the manifest/squashfs, then raises exceptions because it doesn't exist.

PR #612, opened 2025-07-31
fix: update_slot doesn't raise exception on process_regular_files failed

Bodong-Yang:
## Introduction

This PR fixes a bug that `update_slot` doesn't raise exception if `process_regular_files` phase failed and interrupted, but just directly finish up and return.

### DIFF
--- a/src/otaclient/_logging.py
+++ b/src/otaclient/_logging.py
@@ -235,7 +235,16 @@ def configure_logging() -> None:
     # ------ configure each sub loggers and attach ota logging handler ------ #
     log_upload_handler = None
     if logging_upload_endpoint := proxy_info.logging_server:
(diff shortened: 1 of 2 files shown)


src/otaclient/create_standby/update_slot.py
```diff
@@ -247,5 +247,8 @@ def update_slot(self) -> None:
         self._process_non_regular_files()
         self._process_regular_file_entries()
 
+        if self._interrupted.is_set():
+            raise UpdateStandbySlotFailed("failure during regular files processing!")
```
airkei, line 251: The code looks good. How about adding an E2E test case to ensure this exception is raised correctly?

airkei (approved):
:+1:

PR #614, opened 2025-08-01
refactor!: otaproxy: for multiple ECUs environment, disable OTA cache for child ECUs

Bodong-Yang:
## Introduction

This PR disables the otaproxy cache on child ECU(child ECU is ECU that has `upper_ota_proxy` configured) in multiple ECUs environment, no matter how `enable_local_ota_proxy_cache` is configured.

The motivation is, starting from otaclient v3.9.1, we have OTA resume feature(requires inplace mode), so for child ECU, we don't need rely on OTA cache to speed up the OTA retry anymore.

However, due to proxy_info.yaml is forced preserved across each OTA(unfortunately this is ensured by multiple methods from multiple area), updating proxy_info.yaml via OTA is currently impossible, so this PR comes up.

### DIFF
--- a/src/otaclient/_otaproxy_ctx.py
+++ b/src/otaclient/_otaproxy_ctx.py
@@ -104,7 +104,7 @@ def otaproxy_process(
         upper_proxy=upper_proxy,
         # NOTE(20250801): Starting from otaclient v3.9.1, we have inplace update mode with OTA resume,
         #                   on child ECU, we don't need rely on OTA cache to speed up the OTA retry anymore.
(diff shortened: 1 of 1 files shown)


Bodong-Yang:
@airkei Ah inplace mode is not directly related though, it is the OTA resume feature that supersede the role that otaproxy cache previously played on the sub ECUs.

OTA resume requires inplace mode to function(of course, as rebuild mode will simply erase everything from the standby slot).

airkei (approved):
Thank you, LGTM!

PR #611, opened 2025-07-30
feat: add reboot start timestamp/boot controller type into metrics

airkei:
### Why
Current metrics doesn't support sub-ECU waiting time / boot controller type.

### What
Add new `reboot_start_timestamp` and `bootloader_type` metrics.
## Check list

- [x] test file(s) that covers the change(s) is implemented.
- [x] local test is passed.
- [ ] design docs/implementation docs are prepared.

Verified the actual behavior
```
Jul 31 13:14:59 autoware-ecu python3[142166]: [2025-07-31 13:14:59,321][INFO]-otaclient.metrics:publish:125,{"initializing_start_timestamp": 1753931487, "processing_metadata_start_timestamp": 1753931489, "delta_calculation_start_timestamp": 1753931517, "download_start_timestamp": 1753931645, "apply_update_start_timestamp": 1753934469, "post_update_start_timestamp": 1753935293, "finalizing_update_start_timestamp": 1753935299, "reboot_start_timestamp": 1753935299, "ecu_id": "autoware", "session_id": "1753931487-20250717080732-test_ota_v3_9_1-compatibility-b7066c6f", "current_firmware_version": "20250425-v3.0.4-release", "target_firmware_version": "20250717080732-test/ota/v3.9.1-compatibility", "otaclient_version": "3.10.1.dev7+g81d824f3", "failure_type": "", "failure_reason": "", "failed_at_phase": "", "use_inplace_mode": true, "ota_image_total_files_size": 37555383699, "ota_image_total_regulars_num": 396159, "ota_image_total_directories_num": 0, "ota_image_total_symlinks_num": 0, "delta_download_files_num": 163938, "delta_download_files_size": 31324138677, "downloaded_bytes": 481961, "downloaded_errors": 0, "bootloader_type": "grub", "enable_local_ota_proxy_cache": false, "cache_total_requests": 163944, "cache_external_hits": 0, "cache_local_hits": 0}

```
- 6 seconds before rebooting is enough to publish metrics
- `reboot_start_timestamp` and `bootloader_type` are listed in metrics outputs

## Behavior changes

Does this PR introduce behavior change(s)?

- [ ] Yes, internal behaivor (will not impact user experience).
- [ ] Yes, external behaivor (will impact user experience).
- [x] No.

## Breaking change

Does this PR introduce breaking change?

- [ ] Yes.
- [x] No.

PR #617, opened 2025-08-01
feat: add reboot start timestamp/boot controller type into metrics (#…

airkei:
backport 9b8f385be138519879e87e830d43a5f3a9b1d0a1 from https://github.com/tier4/ota-client/pull/611

### DIFF
--- a/src/otaclient/ota_core.py
+++ b/src/otaclient/ota_core.py
@@ -938,25 +938,17 @@ class _OTAUpdater(_OTAUpdateOperator):
     def _finalize_update(self) -> None:
         """Finalize the OTA update."""
         logger.info("local update finished, wait on all subecs...")
-        _current_time = int(time.time())
+        _current_finalizing_time = int(time.time())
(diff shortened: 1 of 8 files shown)


src/otaclient/ota_core.py
```diff
@@ -965,6 +957,18 @@ def _finalize_update(self) -> None:
                 log_func=logger.info,
             )
 
+        _current_reboot_time = int(time.time())
+        self._metrics.reboot_start_timestamp = _current_reboot_time
```
timothylin, line 961: should the `int()` casting be done here instead?

timothylin (approved):
👍

src/otaclient/ota_core.py
airkei: Good point. It would indeed be better to cast at line 961 as you suggested, since keeping the original time value could be useful for future use cases.
However, since this is a backport PR, I’d like to keep it consistent with the original PR. So let me keep as is :bow:

PR #627, opened 2025-08-12
chore: add system uptime debug message

airkei:
### Why
Sometimes it takes time to wake up otaclient after rebooting

### What
Add system uptime for logging message.
NOTE: so far, we don't add this to metrics because it is not a part of Update metrics.

### Test
example message:
```
system uptime: 0h 1m 59.0s (119.0s total)
```

### DIFF
--- a/src/otaclient/main.py
+++ b/src/otaclient/main.py
@@ -130,8 +130,7 @@ def main() -> None:  # pragma: no cover
     )
     # Log system uptime (time since OS boot)
     try:
-        with open("/proc/uptime", "r") as f:
-            uptime_seconds = float(f.read().split()[0])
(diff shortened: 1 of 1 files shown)


Bodong-Yang:
Good idea to add a debug message revealing the uptime of the current ECU :+1: 

Actually python's time lib directly support accessing the system boot uptime via `CLOCK_BOOTTIME`(see https://docs.python.org/3/library/time.html#time.CLOCK_BOOTTIME).

We can simply access the uptime without open `/proc/uptime` as follow:
```
uptime_s = time.clock_gettime(time.CLOCK_BOOTTIME)
```

Bodong-Yang (approved):
👍

PR #639, opened 2025-08-18
fix!: remove rollback API support

Bodong-Yang:
## Introduction

The rollback API is not used, while exposes a potential DoS vulnerability that can be utilized by attackers that has local network access to the otaclient grpc server listen address due to the request is not signed and verified.

This PR removes the actual rollback grpc API handler, replaced with a dummy handler that responds with failure `RECOVERABLE` and message `rollback support is removed`, indicating rollback API is invalid. 

TODO: redesign the rollback API to support request signing, or completely remove this API from the API definition.

### DIFF
--- a/src/otaclient/grpc/api_v2/servicer.py
+++ b/src/otaclient/grpc/api_v2/servicer.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 """OTA Service API v2 implementation."""
 
-
 from __future__ import annotations
 
 import asyncio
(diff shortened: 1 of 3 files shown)


airkei (approved):
LGTM, thank you.

> TODO: redesign the rollback API to support request signing, or completely remove this API from the API definition.

I prefer to remove this API from the API definition/code completely :) Because we don't have plan to support rollback and at least, WebAuto don't use this API at all.

PR #640, opened 2025-08-18
feat: make next OTA delta calculation faster by preserving OTA resources dir after OTA

Bodong-Yang:
## Introduction

> [!NOTE]
> This PR is based on #637 .

This PR introduces to preserve the OTA resources dir on the updated standby slot for making the next OTA(with inplace update mode) delta calculation faster, without the need to scan through the whole file system, but only need to scan through the resource dir. 

Delta calculation will be faster by saving the time of travel through the file system hierarchy. In my local OTA test, this reduce around 10% of the delta calculation time.

## Ticket
https://tier4.atlassian.net/browse/RT4-19104


Bodong-Yang:
Note that, during update_slot, the resources are placed to locations by either hardlink(the first entry in the group of files with the same digest), or by copy(entries other than the first one in the digest group), so preserving the OTA resource dir will not cause extra disk usage.

src/otaclient/create_standby/delta_gen.py
```diff
@@ -572,6 +572,10 @@ def _calculate_delta(self, base_fst: str):
         for _input in self._fst_db_helper.iter_common_regular_entries_by_digest(
             base_fst
         ):
+            _digest, _ = _input
+            if _digest not in self._all_resource_digests:
+                continue  # skip resources that already prepared(via resume OTA)
+
```
airkei, line 578: Could you add some comments regarding `_digest` and `_all_resource_digests`?
From the variable names, I was under the impression that _all_resource_digests might represent the set of digests that have been prepared(processed), like in the following logic:
```
            if _digest in self._all_resource_digests:
                continue 
```

src/otaclient/ota_core.py
```diff
@@ -984,7 +983,8 @@ def execute(self) -> None:
         """
         try:
             self._execute_update()
-            shutil.rmtree(self._resource_dir_on_standby, ignore_errors=True)
+            # NOTE(20250818): preserve the OTA resource dir to speed up next OTA
+            # shutil.rmtree(self._resource_dir_on_standby, ignore_errors=True)
```
airkei, line 987: I think we can delete this `rmtree` line :)

airkei (commented):
Thank you, left some minor comments. Please take a look :+1:

timothylin:
quick question @Bodong-Yang -san. are there any concerns with regards to disk space in keeping these resources?

Bodong-Yang:
> quick question @Bodong-Yang -san. are there any concerns with regards to disk space in keeping these resources?
Yes! Please take a look at this comment:
https://github.com/tier4/ota-client/pull/640#issuecomment-3204122138

src/otaclient/ota_core.py
Bodong-Yang: Updated in [324b952](https://github.com/tier4/ota-client/pull/640/commits/324b9528374851e8fac09d3cd4cdc4d15e28a3ef)

airkei (approved):
LGTM, thank you :)

PR #642, opened 2025-08-19
feat: do fstrim during OTA(inplace mode) and on otaclient starts up

Bodong-Yang:
## Introduction

> [!NOTE]
> This PR is based on #637 .

According to my OTA bench test, `fstrim` can resolve the issue of degraded small files random access IO on specific ECUs emmc. This PR introduces to do `fstrim` when:
1. otaclient starts up, run asynchronously in the background, on current active slot, with 6mins timeout.
2. run synchronously before actually doing OTA, on standby slot, with 30seconds timeout.

`fstrim` itself is very light-weight without putting pressure on disk's IO, and safe to be interrupted.

## Tests

- [x] local pytest passed.
- [x] local OTA test with VM passed.
- [x] local OTA test: confirm fstrim occurs on otaclient starts up and doesn't impact otaclient main process.
- [x] local OTA test: confirm fstrim on inplace update mode occurs, and the timeout is working as expected.

## Ticket
https://tier4.atlassian.net/browse/RT4-19104


src/otaclient/ota_core.py
```diff
@@ -1244,6 +1258,14 @@ def __init__(
         self.started = True
         logger.info("otaclient started")
 
+        # NOTE: not doing fstrim at startup when running as dynamic otaclient
+        if not _env.is_dynamic_client_running() and cfg.FSTRIM_AT_OTACLIENT_STARTUP:
+            fstrim_at_thread(
+                Path(CANONICAL_ROOT),
+                waited=False,
+                timeout=cfg.FSTRIM_AT_OTACLIENT_STARTUP_TIMEOUT,
+            )
```
airkei, line 1267: Just curious, why we can't use fstrim at startup when running as dynamic otaclient?

src/otaclient_common/linux.py
```diff
@@ -252,3 +258,33 @@ def copyfile_nocache(src: StrOrPath, dst: StrOrPath) -> None:
...
+
+    fstrim is safe to be interrupted at any time.
+
+    Returns:
+        True if fstrim finished without reaching timeout.
+    """
+
+    def _thread_entry():
+        _cmd = ["fstrim", "-v", str(target_mountpoint)]
+        subprocess.run(_cmd, timeout=timeout)
```
airkei, line 281: How about to use `subprocess_run_wrapper` instead?

airkei (commented):
Just left a few minor comments. Do we have any data or test results showing how this improves performance? If so, it would be great to include them in the description. 👍

src/otaclient/ota_core.py
Bodong-Yang: Ah, the intention is to do fstrim on every ECU boots up, not at every OTA(this is handled in another place), as currently dynamic otaclient launching only occurs at OTA.

src/otaclient_common/linux.py
Bodong-Yang: Here we don't need the output of the command execution, neither need to check the execution result. We only need to dispatch the execution at the background thread without any inspection, so plain `subprocess.run` is totally enough.

Bodong-Yang: ( Also, the target_mountpoint is visible within the dynamic otaclient mount namespace, and the `fstrim` presents within the otaclient squashfs image(installed in the `ubuntu` docker image by default), so no need to do chroot for the command execution.

src/otaclient_common/linux.py
```diff
@@ -252,3 +258,33 @@ def copyfile_nocache(src: StrOrPath, dst: StrOrPath) -> None:
...
+
+    _t = threading.Thread(target=_thread_entry, name="otaclient_fstrim", daemon=True)
+    try:
+        _t.run()
```
airkei, line 285: According to the comment, the intention is to execute as another thread?
```suggestion
        _t.start()
```

Bodong-Yang: !! Oh no, that is a mistake!
`run` actually directly running the target method in current thread instead of spawning a new thread :face_with_head_bandage: , let me fix that ...

src/otaclient_common/linux.py
```diff
@@ -252,3 +257,48 @@ def copyfile_nocache(src: StrOrPath, dst: StrOrPath) -> None:
...
+    # the spawned subprocess will do a double-fork to fully detach
+    #   the fstrim command execution from otaclient process.
+    _cmd = ["nohup", "timeout", str(timeout), *_cmd]
+    _detached_cmd = f"{shlex.join(_cmd)} &"
+    subprocess.run(
+        _detached_cmd,
+        stdin=subprocess.DEVNULL,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+        shell=True,
```
airkei, line 303: If the purpose is to detach from otaclient process, how about to use `start_new_session` instead? In terms of preventing command injection, it will be more secure.

```suggestion
        start_new_session=True,
```

PR #650, opened 2025-08-26
chore: reflect auto generated protobuf type stub file for request id

airkei:
## Description

This is follow up PR of https://github.com/tier4/ota-client/pull/644.
Automated generated code was not updated, need to fix it.
This file is not used in the actual execution, just only for type hint. So this bug/change will not impact the actual behavior.

## Check list

- [x] local test is passed.


timothylin (approved):
👍

PR #655, opened 2025-08-28
ci: use native arm runner for squashfs release

airkei:
### Why
Currently, using cross platform environment(qemu) to build the squashfs for ARM. It's slow.

### What
Migrate to use native ARM runnner.

### Test
- confirmed the squashfs build was succeeded
https://github.com/tier4/ota-client/actions/runs/17290141443
    - before: 6m14s
    - after: 1m24s
- confirmed the ARM based squashfs is created
```
$ unsquashfs otaclient-arm64-v0.1.dev1.squashfs
$ cd squashfs-root/
$ find . -type f -exec file {} \; | grep ELF | awk -F',' '{print $2}' | sort | uniq -c
    812  ARM aarch64
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -67,7 +67,7 @@ jobs:
             runs-on: ubuntu-latest
           - platform: arm64
             base_image: ubuntu:22.04
-            runs-on: ubuntu-22.04-arm
+            runs-on: ubuntu-24.04-arm
     runs-on: ${{ matrix.runs-on }}
     steps:
       - name: checkout
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -56,7 +56,6 @@ jobs:
           path: dist/*.whl
 
   build_squashfs_images:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -65,8 +64,11 @@ jobs:
         include:
           - platform: x86_64
             base_image: ubuntu:22.04
+            runs-on: ubuntu-latest
           - platform: arm64
-            base_image: arm64v8/ubuntu:22.04
+            base_image: ubuntu:22.04
+            runs-on: ubuntu-24.04-arm
```
airkei, line 70: only `ubuntu-22.04-arm` or `ubuntu-24.04-arm` is available.
https://github.blog/changelog/2025-01-16-linux-arm64-hosted-runners-now-available-for-free-in-public-repositories-public-preview/#how-to-use-the-runners

Bodong-Yang (approved):
Thank you, LGTM!

PR #656, opened 2025-08-28
fix: fix dynamic client exit after metrics publish 

airkei:
### Why
This issue was found in code check.
In a dynamic client environment, if an update fails, the following issues may occur depending on task scheduling. This is because `_exit_from_dynamic_client` requests the main process to exit (restarting the otaclient service):
  - session_wd may not be deleted
  - metrics may not be published

### What
execute `_exit_from_dynamic_client` after file deleting, metrics publishing and wait for 6 seconds

### Test
confirmed the behavior in E2E test.

## Check list

- [x] local test is passed.


timothylin (approved):
👍

PR #658, opened 2025-08-28
refine: create_standby: IO tuning, let kernel and python runtime handle data flush and fs sync

Bodong-Yang:
## Introduction

> [!NOTE]
> This PR is based on #657 

Previously, on create_standby.update_slot, each file copy will issue a fsync syscall, this is actually not needed, and might cause IO performance degraded and stall IO on relatively slow device. 

Actually we don't need to manually call fsync and sync, just let kernel handle the data flushing and commits. Finally when the OTA finished and the standby slot is being umounted, the kernel will do a full sync to ensure all changes are truly saved to the disk, again, no need to manually call sync. 

For the risk analysis of forcely poweroff during OTA, see #657 's PR description.

### DIFF
--- a/src/otaclient_common/cmdhelper.py
+++ b/src/otaclient_common/cmdhelper.py
@@ -425,7 +425,7 @@ def mount_rw(
 
     When <enable_optimization> is set to True, the following extra mount options will be added:
         noatime: do not update file access time
-        commit: enlarge batch journal commits interval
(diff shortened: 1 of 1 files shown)


airkei (dismissed):
LGTM, reasonable approach :)

airkei (approved):
LGTM :rocket:

PR #595, opened 2025-07-04
feat: forward port from v3.9.x: new create_standby implementation with in-place mode and OTA resume support

Bodong-Yang:

PR #669, opened 2025-09-09
refine: use inplace update mode when active_slot's OTA resource dir presented

Bodong-Yang:
## Introduction

Previously, to prevent too large delta, when standby slot's disk usage is smaller than 80% of the OTA image's system image size, rebuild mode will be used(see #595). 
But since #640 , (OTA resources dir )delta calculation result is preserved for speeding up OTA, when OTA resources dir presented at active slot, we can use it with inplace mode instead of using rebuild mode.

This PR introduces the refinement, the condition of using inplace mode is updated, if active slot's OTA resources dir is available, the disk usage check will be bypassed.

### DIFF
--- a/src/otaclient/ota_core/_updater.py
+++ b/src/otaclient/ota_core/_updater.py
@@ -211,6 +211,10 @@ class OTAUpdater(OTAUpdateOperator):
     def _pre_update(self):
         """Prepare the standby slot and optimize the file_table."""
         logger.info("enter local OTA update...")
+
+        # NOTE(20250905): if ota_resources dir on active slot presented,
+        #                 no need to use rebuild mode.
+        _ota_resources_dir_presented = self._resource_dir_on_active.is_dir()
         with TemporaryDirectory() as _tmp_dir:
             self._can_use_in_place_mode = use_inplace_mode = can_use_in_place_mode(
                 dev=self._boot_controller.standby_slot_dev,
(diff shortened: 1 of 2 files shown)


src/otaclient/ota_core/_updater.py
```diff
@@ -211,13 +211,21 @@ def _execute_update(self):
     def _pre_update(self):
         """Prepare the standby slot and optimize the file_table."""
         logger.info("enter local OTA update...")
+
+        # NOTE(20250905): if ota_resources dir on active slot presented,
+        #                 no need to use rebuild mode.
+        _ota_resources_dir_presented = self._resource_dir_on_active.is_dir()
```
airkei, line 217: Could you elaborate why we don't have to use rebuild mode when there is ota_resources dir on active slot? In my understanding,
if there is `_resource_dir_on_active`, we can copy active resource_dir from active to standby in in-place mode. So we don't have to do rebuild mode, is my understanding correct?

https://github.com/tier4/ota-client/blob/c00c0699e4c9ed1d72bace5049885b07b4f6d3dd/src/otaclient/ota_core/_updater.py#L418-L427

airkei (commented):
LGTM, just left one question, please take a look it :bow:

airkei (approved):
LGTM, thank you :rocket:

PR #682, opened 2025-10-01
refine: ThreadPoolExecutorWithRetry: now TaskEnsureFailed records the last exception raised from the pool on interruption

Bodong-Yang:
## Introduction

This PR refines the ThreadPoolExecutorWithRetry by improving exception handling to capture and propagate the last exception raised from worker threads that triggers the interruption. The TasksEnsureFailed exception is enhanced to hold the root cause exception, providing better debugging information when tasks fail.
Caller of the ThreadPoolExecutorWithRetry now can retrieve the exception object from the TaskEnsureFailed, and use the original exception for further processing(like re-raising the original exception rather than TaskEnsureFailed).

### DIFF
--- a/src/otaclient_common/retry_task_map.py
+++ b/src/otaclient_common/retry_task_map.py
@@ -228,7 +228,7 @@ class _ThreadPoolExecutorWithRetry(ThreadPoolExecutor):
                 try:  # raise exc to upper caller
                     raise _wrapped_exc
                 finally:
-                    del _wrapped_exc # break circular ref
+                    del _wrapped_exc  # break circular ref
 
             try:
                 done_fut = self._fut_queue.get_nowait()
--- a/src/otaclient_common/retry_task_map.py
+++ b/src/otaclient_common/retry_task_map.py
@@ -214,9 +214,9 @@ class _ThreadPoolExecutorWithRetry(ThreadPoolExecutor):
                         "execution interrupted due to thread pool shutdown"
                     )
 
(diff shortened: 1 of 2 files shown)


timothylin (dismissed):
👍

src/otaclient_common/retry_task_map.py
```diff
@@ -192,7 +215,20 @@ def _fut_gen(self, interval: float) -> Generator[Future[Any]]:
...
+                    _wrapped_exc = TasksEnsureFailed(
+                        "execution interrupted due to thread pool shutdown"
+                    )
+
+                # NOTE: still raise TaskEnsureFailed to upper,
+                #       let choose to upper dig into the TaskEnsureFailed or not.
+                try:  # raise exc to upper caller
+                    raise _wrapped_exc
+                finally:
+                    del _wrapped_exc
```
airkei, line 231: do we need to del this local variable clearly? I think `raise _wrapped_exc` will be enough.

airkei (commented):
Thank you! overall LGTM. Just left one minor comment.

src/otaclient_common/retry_task_map.py
Bodong-Yang: To be noticed that, the `try ... except: raise` statement has special treatment to the above condition, python will break the circular reference for us.

airkei: Thank you!! very deep and useful knowledge :+1:

PR #703, opened 2025-10-22
fix: limit the target fstab entries to preserve from active to standby

airkei:
## Description
`/etc/fstab` might be unintentionally preserved from active to standby even when the project removes certain entries.

## Check list

- [x] E2E tests are passing
  - manually added test entries to `/etc/fstab`, then executed the OTA on VM. After rebooting, confirmed that the added entries were removed.

### Current behavior
Always merge fstab from active to standby in grub system only when the entry doesn't exist in standby-slot.

### Behavior after fix
merge `boot` and `boot/efi` entries in fstab from active to standby in grub system. This `boot` and `boot/efi` are from [USB Installer Specification](https://github.com/tier4/ota-internal/blob/main/usb_installer/make_autoinstall_fstab_additionals.py#L56).
- `/` : always should use standby UUID(has already been implemented)
- `none`: for swap file. we are using a dedicated systemd service to configure the swap file now, don't have to handle at here.

## Related links & ticket

https://tier4.atlassian.net/browse/T4DEV-39187

### DIFF
--- a/src/otaclient/boot_control/_grub.py
+++ b/src/otaclient/boot_control/_grub.py
@@ -737,9 +737,12 @@ class GrubController(BootControllerProtocol):
             raise ota_errors.BootControlStartupFailed(_err_msg, module=__name__) from e
 
     def _update_fstab(self, *, active_slot_fstab: Path, standby_slot_fstab: Path):
-        """Update standby fstab based on active slot's fstab and just installed new stanby fstab.
+        """Rebuild standby fstab using valid mount entries only.
 
-        Preserve standby fstab entries, only merge /boot and /boot/efi from active fstab.
+        - Keep only valid mount entries (skip comments, invalid or broken lines)
+        - Always include '/', '/boot', and '/boot/efi' from active slot
+        - Replace root ('/') UUID with standby slot UUID
+        - Drop all other comments or extra metadata lines
         """
         try:
             standby_uuid = cmdhelper.get_attrs_by_dev(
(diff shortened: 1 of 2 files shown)


timothylin (dismissed):
🚀

Bodong-Yang:
@airkei Emmmm about the `/boot` and `/boot/efi` entries, these entries are generated by USB installer dynamically detecting and preparing the actual ECU's disk according to the USB installer's autoinstall user-data during installing the system, it means **the setup of `/boot` and `/boot/efi` belongs to part of initial setup, thus must be preserved across each OTA**(just like other initial setup configs).

On the other hand, on evaluator builder virtual environment, there is no way to reliably generate the /boot and /boot/efi mount points, so actually we could not and should not provide `/boot` or `/boot/efi` overrides in evaluator building. If by accident we accept incorrect overrides from OTA image, the functionality of the ECU might under the risk.

airkei:
@Bodong-Yang Thank you! sounds reasonable. Fixed in https://github.com/tier4/ota-client/pull/703/commits/ede92acbeb467f228136d17e7fa49b381ac169fb (always preserve `/boot` and `/boot/efi` from active if exists)

src/otaclient/boot_control/_grub.py
```diff
@@ -763,36 +763,53 @@ def _update_fstab(self, *, active_slot_fstab: Path, standby_slot_fstab: Path):
...
+                    # Reference: https://tier4.atlassian.net/browse/T4DEV-39187
+                    entry = active_dict.get(mount_point, standby_entry_match)
+                    merged.append("\t".join(entry.groups()))
                 else:
-                    merged.append("\t".join(ma.groups()))
-            elif line.strip().startswith("#"):
-                merged.append(line)  # re-add comments to merged
+                    merged.append("\t".join(standby_entry_match.groups()))
+            else:
+                merged.append(line)  # Keep comments and blank lines
```
Bodong-Yang: After some considerations, I think let's just only add valid mount point entry into the fstab. 
`update_fstab` method should be able to correct the invalid fstab in some extend(like filtering away invalid mount point entries).

Also, comments are not so necessary, besides `/`, `/boot` and `/boot/efi`, all other entries must either come from pilot-auto integration, or from post OTA services, we can track down the origin without relying on comments.

Bodong-Yang: The new implement is much cleaner and more comprehensive than the aged old implementation, good job!

PR #716, opened 2025-11-10
fix: new OTA image spec support: fix how the CA stores are handled

Bodong-Yang:
## Introduction

Background: the OTA image signing cert chain in different environment are using the same subjects(`ota-sign.tier4.jp`, `ota-intermediate.tier4.jp` and `ota-root.tier4.jp`).

This PR updates the CA certificate store handling to support multiple environments (dev, stg, prd) with identical certificate subjects. Instead of loading all CA certificates into a single store, the code now creates separate CA stores for each root certificate and attempts verification against each store sequentially until finding out the matching chain, or reject the OTA.

## Tests

- [x] Tests covered the case is implemented. 
- [x] OTA with this PR is working against `ota_image` deployment.

### DIFF
--- a/src/ota_metadata/utils/cert_store.py
+++ b/src/ota_metadata/utils/cert_store.py
@@ -223,6 +223,7 @@ def load_ca_store(cert_dir: StrOrPath) -> CAStoreMap:
     ca_stores = CAStoreMap()
     for _cert in cert_dir.glob("*"):
         if _cert.suffix not in [".pem", ".crt"]:
+            logger.info(f"found unrelated files under certs dir: {_cert}")
             continue  # not a cert
 
         try:


src/ota_metadata/utils/cert_store.py
```diff
@@ -189,25 +189,63 @@ def load_ca_cert_chains(cert_dir: StrOrPath) -> CAChainStore:
...
     Raises:
         CACertStoreInvalid on failed import.
     """
     cert_dir = Path(cert_dir)
 
-    ca_store = CACertStore()
+    ca_stores = CAStoreMap()
     for _cert in cert_dir.glob("*"):
+        if _cert.suffix not in [".pem", ".crt"]:
+            continue  # not a cert
```
timothylin: should we log something here in case this was an honest mistake (i.e. typo or trying to import the wrong type of file?)

airkei (approved):
Thank you! LGTM.

PR #735, opened 2025-12-11
chore: change error levels for normal use cases

airkei:
### Why
In the near future, we will extract ERROR logs to trigger alerts.
However, some current ERROR logs are actually expected behaviors, such as:
- `ClientUpdate` fails because the otaclient version is the same
- `oci-layout` file download fails for legacy image formats

### What
Change these log levels to `WARNING`, and reserve `ERROR` for actual error conditions.


Bodong-Yang (approved):
LGTM!

Bodong-Yang (approved):

PR #719, opened 2025-11-19
feat: T4DEV-40870: modify OTA Client to support offline OTA with NFS mount

timothylin:
## Description

https://tier4.atlassian.net/browse/T4DEV-40870

Add support for using pre-mounted NFS drive as external cache for offline OTA.

```json
{
    "timestamp": "2025-12-09 12:16:28,830",
    "level": "INFO",
    "logger": "ota_proxy.ota_cache",
    "function": "__init__",
    "line": "177",
    "message": "external NFS cache source is mounted at: /run/otaclient/mnt/external_nfs_cache_mount_point"
}
```

```json
{
    "initializing_start_timestamp": 1765250207,
    "processing_metadata_start_timestamp": 1765250207,
    "delta_calculation_start_timestamp": 1765250247,
    "download_start_timestamp": 1765250457,
    "apply_update_start_timestamp": 1765250526,
    "post_update_start_timestamp": 1765250629,
    "finalizing_update_start_timestamp": 1765250634,
    "reboot_start_timestamp": 1765250634,
    "ecu_id": "autoware",
    "request_id": "1765250192-f7518c69",
    "session_id": "1765250192-20251116200104-beta_v4_3-1e9c6cc5",
    "current_firmware_version": "",
    "standby_firmware_version": "",
    "target_firmware_version": "20251116200104-beta/v4.3",
    "otaclient_version": "3.12.0rc3.dev37+gffb1c1ea5.d20251208",
    "failure_type": "",
    "failure_reason": "",
    "failed_status": "",
    "use_inplace_mode": true,
    "ota_image_format": "legacy",
    "ota_image_total_files_size": 30267855712,
    "ota_image_total_regulars_num": 409009,
    "ota_image_total_directories_num": 114817,
    "ota_image_total_symlinks_num": 51357,
    "delta_download_files_num": 4895,
    "delta_download_files_size": 1904428339,
    "downloaded_bytes": 1792809899,
    "downloaded_errors": 1,
    "bootloader_type": "grub",
    "enable_local_ota_proxy_cache": true,
    "cache_total_requests": 4915,
    "cache_cdn_hits": 8,
    "cache_external_hits": 0,
    "cache_external_nfs_hits": 3812,
    "cache_local_hits": 0
}
```

## Check list

- [x] test file(s) that cover the change(s) are implemented.
- [x] local tests are passing.
- [x] design docs/implementation docs are prepared.

## Documents

[...]

PR #745, opened 2025-12-12
fix: use local cache first then NFS cache

timothylin:
## Check list

### Current behavior

OTA Client reaches out to NFS cache first instead of local cache, increasing download time.

### Behavior after fix

OTA Client will check local cache first before reaching out to NFS cache.

## Related links & ticket

#719


airkei (approved):
LGTM! Thank you.

PR #757, opened 2025-12-24
fix: check_if_ota_image_v1/download_bsp_version_file: fix http code checking, add missing 403 check

Bodong-Yang:
## Description

This PR fixes an issue of long detection time when OTA with old OTA image.
The cause of the issue is the `check_if_ota_image_v1` helper method misses checking the 403 error code.

Also fix the same issue for `download_bsp_version_file`.

## Check list

### Current behavior

When OTA with old OTA image on web.auto, the `check_if_ota_image_v1` function takes 12 tries to finish.

### Behavior after fix

The `check_if_ota_image_v1` should return with only one try when OTA with old OTA image.


airkei (commented):
Thank you. There is similar check in `_download_bsp_version_file.py`, could you also apply the changes ? 🙇

timothylin (approved):
👍

PR #759, opened 2025-12-24
ci: fix otaclient version incorrect in app image

Bodong-Yang:
## Description

When releasing otaclient, the otaclient within the built otaclient app image's has incorrect version, different from the tag name. This PR fixes the above issue, and add a quick check of built otaclient app image right after build.

Also see https://github.com/tier4/ota-image-builder/pull/84, which fixes the similar issue.

Detailed changes:
1. `actions/build_squashfs_image/action.yml`: add a step to quickly check the built otaclient app image and its version.
2. `app_image/Dockerfile`: fix build version incorrect issue; add an entrypoint to the built image.
3. `src/otaclient/__main__.py`: add support for `version` command.

## Check list

- [x] otaclient app image can be built without problem.
- [x] the version of otaclient in the built otaclient app image is correct: see https://github.com/tier4/ota-client/actions/runs/20812525478/job/59780033611. 
image

## Related links & ticket

https://tier4.atlassian.net/browse/T4DEV-44546?atlOrigin=eyJpIjoiZGY5YWEwZWIzOGE4NGQ0ZmFlZGFhYWMwMzQ4NjQ5YjciLCJwIjoiaiJ9

### DIFF
--- a/docker/app_img/Dockerfile
+++ b/docker/app_img/Dockerfile
@@ -22,14 +22,15 @@ RUN --mount=type=bind,source=.,target=/build,rw \
     apt-get install -y -qq --no-install-recommends \
         git python3 python3-setuptools; \
     uv run --with=pyinstaller==${PYINSTALLER_VER} --locked --no-managed-python --python=3.13 \
-        pyinstaller -D -s --name otaclient src/otaclient/__main__.py
+        pyinstaller -D -s --name otaclient src/otaclient/__main__.py; \
+    cp -R /build/dist /dist
 
 FROM ${UBUNTU_BASE}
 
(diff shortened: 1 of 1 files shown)


src/otaclient/__main__.py
```diff
@@ -17,6 +17,15 @@
 if __name__ == "__main__":
     freeze_support()
 
-    from otaclient import main
+    import sys
 
-    main.main()
+    # NOTE that we currently don't configure otaclient to take any other args,
+    #   so no need to use argparser but just directly check the first arg.
```
airkei: OK. In the future, when we need to support additional arguments such as `--help`, let’s consider using argparser.

PR #792, opened 2026-01-28
fix: compare image BSP version with current firmware BSP version

airkei:
### Why
The standby slot's firmware BSP version may be uncertain or unavailable in some cases when it is changed by external tools. Using the current slot's BSP version (which is always reliably available from `nvbootctrl`) provides a more robust compatibility check.

### What
- Change BSP version compatibility check to compare OTA image BSP version against **current slot** firmware BSP version instead of standby slot
- Simplify `SlotBSPVersion` to `CurrentBSPVersion` (no longer need to track both slots)
- Remove `firmware_bsp_version` file I/O operations - always trust the value from `nvbootctrl`
- Rename `standby_slot_bsp_ver_check` to `current_slot_bsp_ver_check`
- Change BSP error code from unrecoverable to recoverable, as this error is related to the target image itself.

## Check list

- [x] E2E tests are passing
    - From `r35.4.1` to `36.4.0`: update is failed due to BSP version mismatch
    image

    - From `r35.4.1` to `r35.4.1`: update is succeeded

### Current behavior
BSP version compatibility check compares the OTA image's BSP version against the **standby slot's** firmware BSP version, which is read from a `firmware_bsp_version` file.

### Behavior after fix
BSP version compatibility check now compares the OTA image's BSP version against the **current slot's** firmware BSP version, obtained directly from `nvbootctrl`.

## Related links & ticket
https://star4.slack.com/archives/C06T083QWTT/p1769584429412459

### DIFF
--- a/src/otaclient/boot_control/_jetson_common.py
+++ b/src/otaclient/boot_control/_jetson_common.py
@@ -25,7 +25,7 @@ from functools import partial
 from pathlib import Path
 from typing import Any, NamedTuple, Optional
 
-from pydantic import BaseModel, BeforeValidator, PlainSerializer
+from pydantic import BeforeValidator, PlainSerializer
 from typing_extensions import Annotated, Literal, Self
 
 from otaclient_common import cmdhelper, replace_root
@@ -113,20 +113,6 @@ BSPVersionStr = Annotated[
     PlainSerializer(BSPVersion.dump, return_type=str),
 ]
 """BSPVersion in string representation."""
-
-
-class CurrentBSPVersion(BaseModel):
-    """
-    BSP version string schema: Rxx.yy.z
-    """
-
-    version: Optional[BSPVersionStr] = None
-
-    def set_version(self, ver: BSPVersion | None) -> None:
-        self.version = ver
-
-    def get_version(self) -> BSPVersion | None:
-        return self.version
 
 
 NVBootctrlTarget = Literal["bootloader", "rootfs"]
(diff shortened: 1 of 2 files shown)


timothylin (approved):
👍

src/otaclient/boot_control/_jetson_uefi.py
```diff
@@ -975,26 +967,18 @@ def _firmware_update(self) -> bool:
...
-        standby_fw_bsp_ver = self._firmware_bsp_ver_control.standby_slot_bsp_ver
-        active_fw_bsp_ver = self._firmware_bsp_ver_control.current_slot_bsp_ver
-
-        fw_bsp_ver = standby_fw_bsp_ver
-        if not fw_bsp_ver:
-            logger.warning(
-                "standby slot fw BSP version is not available, use active_slot fw BSP ver instead ..."
-            )
-            fw_bsp_ver = active_fw_bsp_ver
+        fw_bsp_ver = self._firmware_bsp_ver_control.current_slot_bsp_ver
```
Bodong-Yang: Let's add a docstring here to explain why we compare check the active slot's firmware BSP version:
1. it is not possible to access standby slot's firmware BSP version, only current slot's version can be acquired.
2. as otaclient will not do firmware update(we use another tool for BSP update), we assume that both slot's firmware BSP version is the same.

src/otaclient/boot_control/_jetson_common.py
```diff
@@ -290,48 +279,21 @@ def __init__(
...
-        # NOTE: only check the standby slot's firmware BSP version info from file,
-        #   for current slot, always trust the value from nvbootctrl.
-        self._version.set_by_slot(current_slot, current_slot_bsp_ver)
-
-    def write_to_file(self, fw_bsp_fpath: StrOrPath) -> None:
-        """Write firmware_bsp_version from memory to firmware_bsp_version file."""
-        write_str_to_file_atomic(fw_bsp_fpath, self._version.model_dump_json())
+        self._version = CurrentBSPVersion()
+        # for both current and standby slot, always trust the value from nvbootctrl.
+        self._version.set_version(current_slot_bsp_ver)
```
Bodong-Yang: Previously as we maintain the record for both slots, we have a dedicated `SlotBSPVersion` pydantic model for this purpose. Now we only care about active slot's firmware BSP version, then there is no need to use a dedicated model, what about just directly use  `BSPVersion` for the `self._version` field.

src/otaclient/boot_control/_jetson_uefi.py
airkei: Thanks, added docstring in https://github.com/tier4/ota-client/pull/792/changes/0910b9da19342042e9b6da3bfe7c30f87fc1b16a.

PR #810, opened 2026-02-04
refactor: remove fstrim over active slot at otaclient startup

Bodong-Yang:
## Description

This PR removes the fstrim over active_slot at otaclient startup. To be noticed that this is not related to the fstrim over standby_slot during OTA with in-place update mode, which is still required.

The fstrim over active_slot should be realized by active slot's system itself.

## Ticket

https://tier4.atlassian.net/browse/T4DEV-48176


timothylin (approved):
LGTM

airkei (approved):
LGTM! Thank you.

PR #821, opened 2026-02-13
deps: deprecate python3.8 and python3.9

airkei:
deprecate(remove) python3.8 and python3.9 completely.
In addition, deprecate ubuntu 20.04 tests.

### Tests
Verified that OTA was succeeded on python 3.10 environment.

### DIFF
--- a/src/ota_metadata/legacy2/rs_table.py
+++ b/src/ota_metadata/legacy2/rs_table.py
@@ -16,7 +16,7 @@
 from __future__ import annotations
 
 import random
-from typing import Annotated, Any, ClassVar, Generator, Literal
+from typing import Annotated, Any, ClassVar, Generator, Literal, Optional
 
 from pydantic import SkipValidation
 from simple_sqlite3_orm import (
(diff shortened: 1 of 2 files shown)


src/ota_proxy/cache_streaming.py
```diff
@@ -24,11 +24,10 @@
 from concurrent.futures import Future, ThreadPoolExecutor
 from pathlib import Path
 from queue import SimpleQueue
-from typing import Any, AsyncGenerator, Callable, TypeVar
+from typing import Any, AsyncGenerator, Callable, ParamSpec, TypeVar
```
airkei: `ParamSpec` is supported in typing after python 3.10.
https://peps.python.org/pep-0612/

src/otaclient_common/_typing.py
```diff
@@ -18,10 +18,9 @@
 import sys
 from enum import Enum
 from pathlib import Path
-from typing import Any, Callable, TypeVar, Union
+from typing import Annotated, Any, Callable, Concatenate, ParamSpec, TypeVar, Union
```
airkei: Concatenate is supported after python 3.10
https://peps.python.org/pep-0612/

src/otaclient_api/v2/_types.py
```diff
@@ -16,18 +16,10 @@
 from __future__ import annotations
 
 from abc import abstractmethod
+from collections.abc import Iterable, Iterator, Mapping
```
airkei: These are supported from python3.10, and can use for both type hint and instance check.

PR #895, opened 2026-04-09
refine(otaproxy): use in-memory light cache index and make sqlite3 DB pure back storage

Bodong-Yang:
## Motivation

With profiling, we further find out two critical hot spots:

1. **per-request sqlite3 cache DB look up**: every requests will go through one look up over the sqlite3 DB directly, although with WAL mode and other code optimization applied, it still contributes to obvious hot spot.
2. **LRU cache rotation when storage reaches soft_limit**: the efficiency of the otaproxy significantly drops on LRU cache rotation. Things become even worse when the cache DB is shallow, causing too early cache rotations.

Considering the above, we propose to:
1. **remove the LRU cache rotation, simplify the otaproxy's implementation**: the costs severely surpass the  benefits brought by the cache rotation.
3. **add a in-memory cache index at the front, and make sqlite3 cache DB a pure storage back**: avoid per-request DB access with fast dict look up,  cache DB now only serves for preload cache metadata at otaproxy restarts.

## Description

This PR introduces a new in-memory cache index for speeding up the requests handling and get rid of sqlite3 DB look up, turn the role of sqlite3 cache DB to a pure storage back, only serves for persisting cache metadata(for otaproxy restarts). Also, simplify the storage management strategy, drop LRU cache rotation in flavor of a much simple strategy.
With this PR, we observe a further 12%\~20% for single client downloading, 13~20% improvement for 4 clients downloading on top of #884.

Also, with new OTA image spec introduced, the files downloading for OTA reduced a lot(thanks to small files in-line and small files bundle), the current cap of 600k entries is much enough(for example, for a typicall main ECU build, we only have ~87k resource blobs).

Detailed of changes:

- **New cache_index.py: In-memory dict-based CacheIndex**, with preloaded from DB at startup. The max entries allowed in cache index are capped at 600k entries(at most take ~112MB memory). CacheDBWriter batches writes/deletes in dedicated background threads(batch 128 commits into one insertion, commit per-second).
- **Simplified disk space management**: 
    - below hard_limit reached soft_limit: cache teeing(multi-clients request the same resource simultaneously) enabled, but skip cache persist.
    - reached hard_limit: only allow direct proxy download.
- **Refine the empty files handling**: now pure empty file request will go to a dedicated fast path. Especially, the test that covers the empty files downloading are added.
- **Further refine URL processing**: aiohttp uses `yarl.URL` to encode the input URL, while or `process_raw_url` also does the encode, let `process_raw_url` directly return an instance of `yarl.URL` to prevent double-encoding.
- **Speed up DNS looks up**: use `aiohttp[speedups]` variant which uses `aiodns`.

Test files refinement:

- e2e/test_otaproxy now tests the two layer proxy chain instead, fully cover the previous single layer e2e test.
- migrate the leftover otaproxy related test files to `test` folder.

## Check list

- [x] Actual OTA test with VM and full downloading from real OTA image server.

## Tests & Improvements

With setting up local test environment, we observe the following further improvement with this PR on top of #884 :

### Single client test throughput

| Scenario   | #884 (MiB/s) | this PR (MiB/s) | Change |
|------------|-----------|-----------|--------|
| Cold cache |      63.4 |      71.3 | +12.5% |
| Warm cache |      83.4 |      101.1 | +21.2% |

At single download client setup, we observe a further 12%~20% improvement on top of #884.

### Multiple clients test (4 clients) throughput

| Scenario   | #884  (MiB/s) | this PR (MiB/s) | Change |
|------------|-----------|-----------|--------|
| Cold cache |     156.0 |     187.6 | +20.03% |
| Warm cache |     206.8 |     234.5 | +13.4% |

At 4 download clients setup, we observe a further 13~20% improvement on top of #884.

## Behavior changes

- [x] No.

## Breaking change

- [x] No.


src/ota_proxy/config.py
```diff
@@ -43,14 +38,20 @@ class Config:
     MAX_PENDING_READ = 512
 
     # ------ storage quota ------ #
-    DISK_USE_LIMIT_SOFT_P = 70  # in p%
+    DISK_USE_LIMIT_SOFT_P = 75  # in p%
```
Bodong-Yang, line 41: Slightly narrow down the space between soft limit and hard limit.

src/ota_proxy/ota_cache.py
```diff
@@ -616,8 +545,11 @@ async def _retrieve_file_by_new_caching(
             # NOTE(20260403): only provider can do the cleanup for retry_cache
             # if set, cleanup any previous cache file before starting new cache
             if cache_policy.retry_caching:
-                await self._lru_helper.remove_entry(cache_identifier)
-                await (self._base_dir / cache_identifier).unlink(missing_ok=True)
+                self._cache_index.remove_entry(cache_identifier)
+                try:
+                    os.unlink(os.path.join(self._base_dir, cache_identifier))
```
Bodong-Yang, line 559: `os.unlink` is quite light comparing to dispatch the call to thread pool.

src/ota_proxy/ota_cache.py
```diff
@@ -628,10 +560,9 @@ async def _retrieve_file_by_new_caching(
                 compression_alg,
                 resp_headers_from_upper=resp_headers,
             )
+            tracker.cache_meta = cache_meta
             wrapped_fd = self._cache_write_pool.stream_writing_cache(
-                fd=remote_fd,
-                tracker=tracker,
-                cache_meta=cache_meta,
+                fd=remote_fd, tracker=tracker
```
Bodong-Yang, line 574: IMPORTANT: now we directly assign the cache_meta to the tracker instead of let the cache writing handler to do so.

src/ota_proxy/cache_streaming.py
```diff
@@ -181,38 +183,42 @@ def _finalize_cache(self) -> None:
...
+            if not self._commit_cache_cb(self.cache_meta):
+                return
+        except Exception as e:
+            burst_suppressed_logger.warning(f"failed to commit cache entry: {e}")
+            return
+
         try:
             os.link(self.fpath, self.save_path)
         except FileExistsError:
             pass
```
Bodong-Yang, line 211: Since now we use in-memory cache index for committing cache entry at front, we don't need to care about the delay between file is finalized and cache entry committed.
Also, for aligning with the new space management strategy, skip file finalizing when cache commit doesn't occur(like in-memory cache entries exceeds limit).

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dynamic = [
 ]
 dependencies = [
   "anyio>=4.5.1,<5",
-  "aiohttp>=3.10.11,<3.14",
+  "aiohttp[speedups]>=3.10.11,<3.14",
```
Bodong-Yang, line 32: This will pull in `aiodns`, which will speed up the DNS look up.

src/ota_proxy/ota_cache.py
```diff
@@ -470,14 +398,14 @@ async def _retrieve_file_by_cache_lookup(
             # fallback to use URL based hash, and clear compression_alg for such case
             cache_identifier = url_based_hash(raw_url)
 
-        meta_db_entry = await self._lru_helper.lookup_entry(cache_identifier)
-        if not meta_db_entry:
+        index_entry = self._cache_index.lookup_entry(cache_identifier)
+        if not index_entry:
```
Bodong-Yang, line 407: Now we look up cache from the in-memory cache index instead of directly accessing the sqlite3 DB.

src/ota_proxy/cache_index.py
```diff
@@ -0,0 +1,339 @@
...
+            for row in orm.orm_select_entries(
+                _stmt=gen_sql_stmt(
+                    "SELECT", "*",
+                    "FROM", self._table_name,
+                    "ORDER BY", "last_access", "DESC"
+                )
+            ):
+            # fmt: on
+                if _count < cfg.MAX_INDEX_ENTRIES:
+                    _key = sys.intern(row.file_sha256)
```
airkei, line 275: (nit) According to the official specification, it speeds up if both keys in a dictionary and  the lookup key is interned.
https://docs.python.org/3/library/sys.html#sys.intern
In `lookup_entry`, `file_sha256` is not interned, so the benefits to use `intern` at here might be limited.

airkei (approved):
Thank you, LGTM! Please check copilot comments too.

src/ota_proxy/cache_index.py
```diff
@@ -0,0 +1,339 @@
...
+        if write_batch:
+            # Fill bucket_idx for backward compat with older otaproxy LRU
+            for entry in write_batch:
+                entry.bucket_idx = (
+                    bisect.bisect_right(self._bucket_size_list, entry.cache_size) - 1
+                )
+
+            try:
+                self._orm.orm_insert_entries(write_batch, or_option="replace")
+            except Exception as e:
```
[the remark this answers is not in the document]
Bodong-Yang: Ah, doc string mismatch, will fix up it now.
Also this is a trade off for simplicity, we intentionally ignore the order of deletion and insertion, but simply batch them together, and always apply deletions after writes. 
This is fine as sqlite3 DB is just for persisting cache metadata for next otaproxy restarts, it doesn't affect the current otaproxy session.

src/ota_proxy/cache_streaming.py
```diff
@@ -283,6 +287,11 @@ def provider_write_once_in_thread(self, cache_meta: CacheMeta, data: bytes) -> N
         finally:
             del self, data, cache_meta
 
+    def provider_handle_empty_file(self) -> None:
+        assert self.cache_meta
+        self._tracker_events.set_writer_finished()
+        self._commit_cache_cb(self.cache_meta)
+
```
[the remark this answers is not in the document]
Bodong-Yang: Because empty file doesn't take space, so no need to care about soft/hard SPACE usage limitation.

src/ota_proxy/cache_streaming.py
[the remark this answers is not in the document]
Bodong-Yang: No, we CANNOT first link the file and then commit the cache entry records. 
In case of the in-memory cache index reaching the maximum, link the file first will result in untracked cache files leaking.

airkei (approved):

src/ota_proxy/cache_index.py
Bodong-Yang: Good catch! This should also contribute to make the cache look up faster, let me fix it.

Bodong-Yang: Updated in [4f37baa](https://github.com/tier4/ota-client/pull/895/commits/4f37baa8474845cb99155c848e680f517618df72), now both cache lookup and new cache path will have the cache_identifier(the key of the in-memory cache index) interned.

airkei (approved):

PR #907, opened 2026-04-16
fix: client_package: fix package downloading not have hash checked

Bodong-Yang:
## Description

Currently, when doing otaclient update otaclient package downloading, the hash verification is not performed over the downloaded package. This might cause issue when the network is unstable and the package is partially downloaded. Further more, after the package is downloaded, it will be persisted to skip same packing being downloaded again if OTA failed and retried, this will fail the client update(due to broken otaclient image cannot be started).

This PR fixes this issue by providing the hash digest value when requesting downloading otaclient package. 

Other changes:
1. migrate the otaclient client_package downloading test to follow the new test layout.

## Check list

- [x] OTA test(with client update) with local VM passed.


timothylin (approved):
👍

airkei (approved):
Thank you, LGTM!

PR #912, opened 2026-04-20
feat: add version detail handling

airkei:
## Description

Persist and restore an extended `version_detail` metadata file (`release_name`, `release_id`, `image_id`) alongside the existing `version` file on each OTA slot. The upper layer may attach these fields to an `Update` / `ClientUpdate` request; when present, they are stored on the standby slot at post-update and can be read back via the Status API after reboot.

Motivation: `version` alone is not enough to identify which release / image a slot was built from. Exposing the richer triple through the Status API lets the upper layer (FMS) correlate a running slot with the release it originated from.

## Check list

- [x] executed OTA and verified that both `version` and `version_detail` are stored in the expected location.
```
autoware@autoware-ecu:~$ cat /boot/ota-slot_b/version
20260422153720-test/ota/beta/v4.4
autoware@autoware-ecu:~$ cat /boot/ota-slot_b/version_detail 
{"release_name": "20260422153720-test/ota/beta/v4.4", "release_id": "014caa00-b8b7-4660-ab69-6d3349538a0b", "image_id": "7d0368d9-b6b4-46e7-9466-2a8b91c2a070"}
```
- [x] detail version information is stored in Status API response
```
Apr 23 09:08:13 autoware-ecu otaclient[2880]: {"timestamp":"2026-04-23 09:08:13,963","level":"INFO","logger":"otaclient._status_monitor","function":"_ota_status_logging_thread","line":"355","message":
"ongoing OTA: OTAClientStatus(
firmware_version='20260422153720-test/ota/beta/v4.4',
version_detail=VersionDetail(release_name='20260422153720-test/ota/beta/v4.4', release_id='014caa00-b8b7-4660-ab69-6d3349538a0b', image_id='7d0368d9-b6b4-46e7-9466-2a8b91c2a070'),
...
```

## Documents

https://tier4.atlassian.net/wiki/spaces/OTA/pages/5151260680/ECU-level+Firmware+Update+History+Management

## Changes

- Add `VersionDetail` dataclass (`release_name`, `release_id`, `image_id`) in `otaclient/_types.py`.
- Add `version_detail` field to `OTAClientStatus` and propagate it through `_status_monitor`, gRPC API v2 types, and the servicer.
- Add `OTA_VERSION_DETAIL_FNAME` config constant.
- `OTAStatusFilesControl`: add `_store_standby_version_detail`, `load_active_slot_version_detail`, `load_standby_slot_version_detail`; extend `post_update_standby` to accept an optional `VersionDetail` and write it as JSON only when provided.
- `BootControllerProtocol.post_update` / `_base.py`: accept and forward `version_detail`; expose `load_version_detail()` for active slot.
- `ota_core/_updater.py`: construct `VersionDetail` at post-update **only when `release_name`, `release_id`, and `image_id` are all non-empty**; otherwise pass `None` and no file is written.
- Plumb `release_name` / `release_id` / `image_id` from `UpdateRequestV2` / `ClientUpdateRequestV2` through `_main.py` into `OTAUpdateInitializer`.
- Tests: new cases in `test_ota_status_control.py` (store with/without detail, load from active slot, missing file, invalid JSON) and `test_base.py` (load_version_detail, post_update forwarding).

## Behavior changes

- [x] Yes, internal behavior (will not impact user experience).
- [ ] Yes, external behavior (will impact user experience).
- [ ] No.

### Previous behavior

Only the firmware `version` string was stored per slot. `release_name` / `release_id` / `image_id` supplied in an update request were not persisted and were not observable after reboot.

### Behavior with this PR

When all three fields are supplied on an update request, a `version_detail` JSON file is written to the standby slot's OTA status directory at post-update. After reboot, it is loaded and exposed through the Status API as `OTAClientStatus.version_detail`. If any of the three fields is empty, the file is not written and `version_detail` is reported as `None` — preserving backward compatibility with callers that do not set these fields.

## Breaking change

- [ ] Yes.
- [x] No.

The new fields are optional on both the request and the status response. Existing callers and existing on-disk state (slots with no `version_detail` file) continue to work unchanged.

## Related links & tickets
https://tier4.atlassian.net/browse/T4DEV-52538

### DIFF
--- a/docker/app_img/README.md
+++ b/docker/app_img/README.md
@@ -4,6 +4,7 @@
 
 ```bash
 # at the project root
+export OTACLIENT_VERSION=<DESIRED_VERSION>
 sudo docker build \
     -f docker/app_img/Dockerfile \
     --no-cache \
@@ -16,7 +17,11 @@ sudo docker build \
 
 ```bash
 # at the project root
+sudo docker rm -f otaclient_app_export || true
 sudo docker create --name otaclient_app_export otaclient_app:${OTACLIENT_VERSION}
+
+mkdir -p dist
+rm -rf dist/otaclient_*.squashfs
 sudo docker export otaclient_app_export | mksquashfs - dist/otaclient_${OTACLIENT_VERSION}.squashfs \
     -tar -b 1M \
     -mkfs-time 1729810800 \
--- a/docker/app_img/README.md
+++ b/docker/app_img/README.md
@@ -4,7 +4,6 @@
 
 ```bash
 # at the project root
-export OTACLIENT_VERSION=<DESIRED_VERSION>
 sudo docker build \
     -f docker/app_img/Dockerfile \
     --no-cache \
@@ -17,11 +16,7 @@ sudo docker build \
 
 ```bash
 # at the project root
-sudo docker rm -f otaclient_app_export || true
 sudo docker create --name otaclient_app_export otaclient_app:${OTACLIENT_VERSION}
-
-mkdir -p dist
-rm -rf dist/otaclient_*.squashfs
 sudo docker export otaclient_app_export | mksquashfs - dist/otaclient_${OTACLIENT_VERSION}.squashfs \
     -tar -b 1M \
     -mkfs-time 1729810800 \
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
   "multidict<7.0,>=4.5",
   "msgpack>=1,<1.2",
   "otaclient_iot_logging_server_pb2 @ https://github.com/tier4/otaclient-iot-logging-server/releases/download/v1.3.0/otaclient_iot_logging_server_pb2-1.0.0-py3-none-any.whl#sha256:e5a2e6474a8b6f5656679877a5df40891e6a65c29830a31faed3d47d9973be60",
-  "otaclient_pb2 @ https://github.com/tier4/otaclient-api/releases/download/v1.0.0/otaclient_pb2-1.0.0-py3-none-any.whl#sha256:86a8095345c65ef5b91c658960723341eaa5e4588eb6b9c731f7401341ae8604",
+  "otaclient_pb2 @ https://github.com/tier4/otaclient-api/releases/download/v1.1.0/otaclient_pb2-1.1.0-py3-none-any.whl#sha256:f10ffc958ff18ca16ebf62b083e0e4903fb24a26c3fec0386c6ec2a05387fdc5",
   "ota-image-libs@https://github.com/tier4/ota-image-libs/releases/download/v0.3.0/ota_image_libs-0.3.0-py3-none-any.whl#sha256:98a411c53a01f313ddfbfcb8c8d20dcd9ced829c7282a01763c81f25c695d79c",
   "protobuf>=4.25.8,<7.35",
   "pydantic<3,>=2.10",
--- a/src/otaclient/_status_monitor.py
+++ b/src/otaclient/_status_monitor.py
@@ -32,6 +32,7 @@ from otaclient._types import (
     UpdatePhase,
     UpdateProgress,
     UpdateTiming,
+    VersionDetail,
 )
 from otaclient._utils import SharedOTAClientStatusWriter
 from otaclient_common._logging import get_burst_suppressed_logger
@@ -61,6 +62,7 @@ atexit.register(_global_shutdown)
 @dataclass
 class SetOTAClientMetaReport:
     firmware_version: str = ""
+    version_detail: VersionDetail | None = None
 
 
 @dataclass
@@ -268,6 +270,8 @@ class OTAClientStatusCollector:
         # ------ update otaclient meta ------ #
         if isinstance(payload, SetOTAClientMetaReport):
             status_storage.firmware_version = payload.firmware_version
+            if payload.version_detail is not None:
+                status_storage.version_detail = payload.version_detail
             return True
 
         # ------ on session start/end ------ #
--- a/src/otaclient/boot_control/_ota_status_control.py
+++ b/src/otaclient/boot_control/_ota_status_control.py
@@ -16,11 +16,12 @@
 from __future__ import annotations
 
 import contextlib
+import json
 import logging
 from pathlib import Path
 from typing import Callable, Optional, Union
 
-from otaclient._types import OTAStatus
+from otaclient._types import OTAStatus, VersionDetail
 from otaclient.configs.cfg import cfg
 from otaclient_common import _env
 from otaclient_common._io import read_str_from_file, write_str_to_file_atomic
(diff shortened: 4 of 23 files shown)


src/otaclient/boot_control/_ota_status_control.py
```diff
@@ -257,6 +274,8 @@ def post_update_standby(self, *, version: str):
         # store status to standby slot
         self._store_standby_status(OTAStatus.UPDATING)
         self._store_standby_version(version)
+        if version_detail is not None:
+            self._store_standby_version_detail(version_detail)
```
airkei, line 278: When ota-client doesn't receive detailed version information from Agent, doesn't create `version_detail` file.

src/otaclient/ota_core/_main.py
```diff
@@ -97,6 +97,8 @@
 WAIT_FOR_OTAPROXY_ONLINE = 3 * 60  # 3mins
 WAIT_BEFORE_DYNAMIC_CLIENT_EXIT = 6  # seconds
 
+SKIP_CLIENT_UPDATE = False  # for testing purpose, skip client update
```
airkei: internal macro to skip client update

src/otaclient/ota_core/_updater.py
```diff
@@ -311,7 +311,17 @@ def _post_update(self) -> None:
...
 
         self._preserve_client_squashfs_at_post_update()
-        self._boot_controller.post_update(self.update_version)
+        _version_detail: VersionDetail | None = None
+        if self.release_name and self.release_id and self.image_id:
+            _version_detail = VersionDetail(
+                release_name=self.release_name,
+                release_id=self.release_id,
+                image_id=self.image_id,
+            )
```
airkei, line 320: create `version_detail` file only when ota-client receives the valid `release_name`, `release_id` and `image_id`.

src/otaclient/ota_core/_main.py
```diff
@@ -201,11 +203,13 @@ def __init__(
         _boot_ctrl_loaded_ota_status = self.boot_controller.get_booted_ota_status()
         self._live_ota_status = _boot_ctrl_loaded_ota_status
         self.current_version = self.boot_controller.load_version()
+        self.current_version_detail = self.boot_controller.load_version_detail()
```
airkei, line 204: load `version_detail` in boot-up sequence.

src/otaclient/_status_monitor.py
```diff
@@ -268,6 +270,8 @@ def load_report(self, report: StatusReport) -> bool:
         # ------ update otaclient meta ------ #
         if isinstance(payload, SetOTAClientMetaReport):
             status_storage.firmware_version = payload.firmware_version
+            if payload.version_detail is not None:
+                status_storage.version_detail = payload.version_detail
```
airkei, line 274: update `Status` field only when the detailed version information is valid.

pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
   "multidict<7.0,>=4.5",
   "msgpack>=1,<1.2",
   "otaclient_iot_logging_server_pb2 @ https://github.com/tier4/otaclient-iot-logging-server/releases/download/v1.3.0/otaclient_iot_logging_server_pb2-1.0.0-py3-none-any.whl#sha256:e5a2e6474a8b6f5656679877a5df40891e6a65c29830a31faed3d47d9973be60",
-  "otaclient_pb2 @ https://github.com/tier4/otaclient-api/releases/download/v1.0.0/otaclient_pb2-1.0.0-py3-none-any.whl#sha256:86a8095345c65ef5b91c658960723341eaa5e4588eb6b9c731f7401341ae8604",
+  "otaclient_pb2 @ https://github.com/tier4/otaclient-api/releases/download/v1.1.0/otaclient_pb2-1.1.0-py3-none-any.whl#sha256:f10ffc958ff18ca16ebf62b083e0e4903fb24a26c3fec0386c6ec2a05387fdc5",
```
airkei, line 38: upgrade `otaclient-api` package.

src/otaclient/ota_core/_main.py
Bodong-Yang: I am afraid to ship this hack into the code, if we need this kind of hack for testing, please use a test branch and apply the hack on-demand(For me, I always apply patch when need to skip client update.)

src/otaclient/ota_core/_main.py
```diff
@@ -447,6 +454,10 @@ def client_update(self, request: ClientUpdateRequestV2) -> None:
         NOTE that client update API will not raise any exceptions. The failure information
             is available via status API.
         """
+        if SKIP_CLIENT_UPDATE:
+            logger.warning("SKIP_CLIENT_UPDATE is set to True, skip client update")
+            return
+
```
Bodong-Yang: Please remove this branch.

src/otaclient/_status_monitor.py
```diff
@@ -61,6 +62,7 @@ def _global_shutdown():
 @dataclass
 class SetOTAClientMetaReport:
     firmware_version: str = ""
+    version_detail: VersionDetail | None = None
```
Bodong-Yang, line 65: This will cause issue at python3.8 as dataclass will parse the typing here, but python3.8 still doesn't support pipe for optional, need to use `Optional[VersionDetail]` here.

docker/app_img/README.md
Bodong-Yang, line 1: The changes here is not related to the PR, let's remove it.

Bodong-Yang (commented):
I leave some comments, please have a check 🙇

src/otaclient/ota_core/_main.py
airkei: Hmmm, since this branch doesn’t change the security strength, I think it’s still useful.
That said, it’s unrelated to this PR, so I’ve reverted it in 7045f59 :+1:

docker/app_img/README.md
airkei: OK, reverted in https://github.com/tier4/ota-client/pull/912/changes/9fd1d3b867cd65e17f1d1f0d20ddcd52acbc7d09. I will create another PR :+1:

src/otaclient/_status_monitor.py
airkei: Hmm, seems this is safe on Python 3.8. The file has `from __future__ import annotations`, which per PEP 563 causes all annotations to be stored as strings in `__annotations__` rather than being evaluated at runtime. So `VersionDetail | None` is never executed as an expression — it's just stored as the string.

The `@dataclass` decorator only reads the annotation keys (field names) and default values from `__annotations__`. it does not `eval()` the type strings. So seems there is no TypeError on Python 3.8.

- cpython: `eval` is not used.
https://github.com/python/cpython/blob/3.8/Lib/dataclasses.py#L856-#L862

- just in case, I tested the following code and it wokrs in python3.8 env :+1: 
```
"""Test 1: X | None in dataclass WITH from __future__ import annotations"""
from __future__ import annotations
from dataclasses import dataclass


@dataclass
class MyData:
    name: str = ""
    value: int | None = None


obj1 = MyData()
print(f"default: {obj1}")

obj2 = MyData(name="hello", value=42)
print(f"with values: {obj2}")

print(f"__annotations__: {MyData.__annotations__}")

import sys
print(f"Python {sys.version}")
print("SUCCESS: dataclass with pipe annotation works!")
```

Bodong-Yang: Oh got it, seems like dataclass field type hint is only for typing purpose, no parsing over it :+1:

Bodong-Yang (approved):
LGTM, thank you :+1:

PR #913, opened 2026-04-20
refine(otaproxy): cache_index: store pre-computed headers dict; read_file_once: skip fadvise for small files read

Bodong-Yang:
## Description

Following #884 and #895, this PR applies another optimizations to the otaproxy:

1. now the element of the in-memory cache, the CacheIndexEntry stores the pre-computed read-only header(as `CIMultiDictProxy`), directly ready for use, reducing two header builds per request to one.

2. for small files handling path(`read_file_once`),  skip two syscalls of fadvice, as small files grow the memory usage much slower, kernel can handle the cache evicts without much problem. 

The above two optimizations further reduce processing time cost per-request, contributing to faster request handling. 

## Check list

- [x] OTA bench test passed. Test report: https://tier4.atlassian.net/wiki/x/voC-MwE


airkei (approved):
Thank you, LGTM.

PR #918, opened 2026-04-23
[backport v3.13.x] refine(otaproxy): allow otaproxy to run even if sqlite3 DB init failed

Bodong-Yang:
## Description

This PR adds safety guard around sqlite3 DB preload and init paths at CacheIndex starts up. The back storage sqlite3 DB now can be optional if somehow we cannot make the DB work, allows otaproxy to still run normally without a back sqlite3 DB.

This will target the otaclient v3.13.5 release.

## Check list


airkei (approved):
Thank you, LGTM!

PR #930, opened 2026-04-28
fix(ecu_status): add LocalECUStatusNotReady exception and handle in gRPC service

airkei:
## Description
This PR fix the existing issues that FMS Console outputs the following errors when OTA is triggered.
```
Update request to OTA client failed.
autoware are listed as available_ecu_ids, but autoware: not responding
```
This happens when `autoware` is listed in `available_ecu_ids` but Status response(`ecu_v2` field) doesn't include `autoware`.

## Check list

- [x] created [a test image](https://evaluation.dev.tier4.jp/evaluation/reports/d164eab2-10c1-50a1-b076-00bfefc4de17?project_id=tier4_demo) and tested on FMS Dev environment 

### Current behavior

The gRPC API server starts and begins answering `Status` RPCs before the OTA core finishes its bootstrap and writes its first `OTAClientStatus` into shared memory. Meanwhile `available_ecu_ids` is already populated synchronously from `ecu_info.yaml` at storage construction.

The result is a `StatusResponse` where:
- `available_ecu_ids` contains the local ECU (e.g. `autoware`)
- `ecu_v2` does **not** contain an entry for the local ECU

Downstream agents can misread this as "the ECU is alive but reporting no status," which is semantically ambiguous and not a transient signal the caller knows to retry on. The polling task itself never crashes, so the symptom is silent.

### Behavior after fix
Make the Status RPC return `UNAVAILABLE` instead of an asymmetric `StatusResponse` (where the local ECU appears in `available_ecu_ids` but is missing from `ecu_v2`) during the startup window before the OTA core has written its first status into shared memory.

**Behavior matrix:**

| Local ECU configuration | Local status in storage | `export()` result |
|---|---|---|
| OTA-managed (in `available_ecu_ids`) | absent | raises `LocalECUStatusNotReady` → `UNAVAILABLE` |
| OTA-managed (in `available_ecu_ids`) | present | normal `StatusResponse` |
| Proxy/aggregator (not in `available_ecu_ids`) | absent | normal `StatusResponse` (steady state, not a startup race) |
| Proxy/aggregator (not in `available_ecu_ids`) | present | normal `StatusResponse` |

### DIFF
--- a/src/otaclient/grpc/api_v2/ecu_status.py
+++ b/src/otaclient/grpc/api_v2/ecu_status.py
@@ -61,6 +61,23 @@ IDLE_POLLING_INTERVAL = 10  # second
 ACTIVE_POLLING_INTERVAL = 1  # seconds
 
 
+class LocalECUStatusNotReady(Exception):
+    """Raised by export() during the startup race for an OTA-managed local ECU.
+
+    When the local ECU is listed in its own available_ecu_ids but the OTA core
+    has not yet written its first OTAClientStatus into shared memory, the
+    storage has no entry for my_ecu_id. Returning a StatusResponse in that
+    state would advertise my_ecu_id in available_ecu_ids while omitting it
+    from ecu_v2, which downstream agents can misinterpret as a healthy ECU
+    with no status. The gRPC layer maps this exception to UNAVAILABLE so the
+    caller will retry instead.
+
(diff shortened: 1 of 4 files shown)


timothylin (approved):
👍

Bodong-Yang (approved):
LGTM 👍

PR #948, opened 2026-05-18
chore(test): migrate e2e test for ota_metadata

Bodong-Yang:
## Introduction

This PR migrates the test_ota_metadata e2e tests (legacy test_legacy2/test_e2e.py and test_ota_image_v1/test_e2e.py) into the new test/e2e/ layout.

### DIFF
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -11,21 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Top-level test helpers shared across unit / integration / e2e tiers.
-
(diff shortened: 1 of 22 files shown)


test/conftest.py
```diff
@@ -26,8 +33,13 @@
     StatusReport,
 )
 
+logger = logging.getLogger(__name__)
+
 MAX_TRACEBACK_SIZE = 2048
 
+_HTTP_SERVER_SCRIPT = Path(__file__).parent / "_utils" / "_http_server.py"
```
airkei, line 40: This might be a bit fragile.

airkei (approved):
LGTM! left one nit comment, but I think it's acceptable as-is since it's not critical and is only used in tests.

PR #954, opened 2026-05-19
chore(test): migrate integration tests for rpi_boot control

Bodong-Yang:
## Introduction

This PR migrates the integration tests for rpi_boot control following the new test layout.
The original test file is already integration test oriented, so directly move it into integration test category.


airkei (approved):
LGTM. might be better to update description, since this is not simple move.

PR #959, opened 2026-05-21
fix: grub_new: fix found ubuntu 24.04 -> 22.04 backward compatibility issue; add grub.cfg min version strategy 

Bodong-Yang:
## Description

This PR introduces the following fixes and features:
1. explicitly disable the `30_uefi-firmware` grub hook on bootstrap and post OTA paths. 
2. introduce a min version track mechanism for `grub.cfg` and set the minimum required version to `v3.14.1`, otaclient will re-bootstrap the boot control when the `grub.cfg` is managed by otaclient older than `v3.14.1`.
3. at pre-update, also set the `FAILURE` OTA status to legacy boot slot folder's OTA status file.

The introduced changes resolve the following reported issues:
1. (critical) under some scenarios, OTA from ubuntu 24.04 back to ubuntu 22.04 will cause system boot failed, boot to the UEFI boot menu.
4. (minor) with otaclient update, when OTA from old otaclient to new otaclient, if the OTA failed in the middle, OTA status will report the old value from the last finished OTA due to boot control hands over back to the old otaclient.

Other changes:
1. set the `packaging` package as explicit dep for otaclient.
2. uv.lock file maintenance.
3. bump download inactive timeout to 10mins.

## Check list

- [x] tests with VM passed, test report: https://tier4.atlassian.net/wiki/x/ugHxNwE


Bodong-Yang:
@claude

airkei (approved):
Thank you, LGTM!

> (minor) with otaclient update, when OTA from old otaclient to new otaclient, if the OTA failed in the middle, OTA status will report the old value from the last finished OTA due to boot control hands over back to the old otaclient.

Good catch, this fix is also important.

timothylin (approved):
👍

PR #1042, opened 2026-07-28
fix: wait for reboot instead of exiting

airkei:
## Description

Make `cmdhelper.reboot` wait for the init system to terminate otaclient, instead of calling `sys.exit(0)` right after issuing the reboot request.

The `reboot` command only *requests* the shutdown to the init system; the actual shutdown happens asynchronously and can take tens of seconds. During that window the system is still running, so otaclient (and its status API) must stay alive.

## Check list

### Current behavior

`reboot` exits the otaclient process immediately after requesting the reboot:

```python
try:
    subprocess_call(cmd, raise_exception=False, chroot=chroot)
finally:
    sys.exit(0)
```
This causes the following problems:

otaclient disappears while the system is still up, so the status API is gone for the tens of seconds between the reboot request and the actual shutdown.
If **the systemd restarts original (legacy) otaclient** during that window, the status API served there no longer reports the ongoing OTA. The upper layer can then treat the OTA as already finished, before the reboot has actually happened.
Because of the finally, the process exits with 0 even when the reboot request itself fails — a failed reboot is indistinguishable from a normal OTA finish.
### Behavior after fix
After the reboot request succeeds, otaclient keeps running (and keeps serving the status API) and waits to be terminated by the init system during shutdown.
The wait is bounded by WAIT_FOR_REBOOT_TIMEOUT (60s, checked every 1s). If the system still hasn't rebooted by then, otaclient exits with 1.
If the reboot request itself fails, otaclient logs the error and exits with 1 immediately, letting the init system restart otaclient, which then detects the unfinished switching-boot.

### Related links & ticket
https://tier4.atlassian.net/browse/T4DEV-57865

### DIFF
--- a/src/otaclient_common/cmdhelper.py
+++ b/src/otaclient_common/cmdhelper.py
@@ -412,7 +412,7 @@ def reboot(args: list[str] | None = None, *, chroot: str | None = None) -> NoRet
     try:
         subprocess_call(cmd, raise_exception=False, chroot=chroot)
     except Exception as e:
-        logger.error(f"failed to request reboot: {e!r}, exit now")
+        logger.exception(f"failed to request reboot: {e!r}, exit now")
         sys.exit(1)
 
     logger.info(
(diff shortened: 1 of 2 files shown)


timothylin (approved):
👍

